PR #11, opened 2017-07-25
fix port

sbose78:
.. for test env setup

### DIFF
--- a/.make/docker-compose.integration-test.yaml
+++ b/.make/docker-compose.integration-test.yaml
@@ -4,11 +4,11 @@ services:
     image: centos/postgresql-95-centos7
     network_mode: "bridge"
     ports:
-      - "5432"
+      - "5433"
     environment:
       POSTGRESQL_ADMIN_PASSWORD: mysecretpassword
     healthcheck:
-      test: "bash -c 'pg_isready -h localhost -p 5432 -q -U postgres'"
+      test: "bash -c 'pg_isready -h localhost -p 5433 -q -U postgres'"
       interval: 3s
       timeout: 7s
       retries: 5
--- a/.make/docker-compose.integration-test.yaml
+++ b/.make/docker-compose.integration-test.yaml
@@ -4,11 +4,11 @@ services:
     image: centos/postgresql-95-centos7
     network_mode: "bridge"
     ports:
-      - "5433"
+      - "5432"
     environment:
       POSTGRESQL_ADMIN_PASSWORD: mysecretpassword
     healthcheck:
-      test: "bash -c 'pg_isready -h localhost -p 5433 -q -U postgres'"
+      test: "bash -c 'pg_isready -h localhost -p 5432 -q -U postgres'"
       interval: 3s
       timeout: 7s
       retries: 5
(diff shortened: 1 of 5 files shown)


sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-11-1`

sbose78:
@alexeykazakov yeah.. i know. I was trying to see if we can have the same port but use a different database inside it ( similar to the way its done in `init-tenant` where a database named `tenant` is used in the same db server as wit )

sbose78:
[test]

sbose78:
trying this out so that SD can proceed with setup of pipelines - was getting this error   

```
--- FAIL: TestMigrations (0.00s) panic: dial tcp 172.17.0.3:5433: getsockopt: connection refused [recovered] panic: dial tcp 172.17.0.3:5433: getsockopt: connection refused
```

but looks like there's the issue of entitlements' test failing..

hectorj2f:
@sbose78 connection refused means that there is a problem to talk to the db, no ?

sbose78:
@hectorj2f : could be an issue with the db being setup on a different port for tests: https://github.com/fabric8-services/fabric8-wit/blob/master/.make/docker-compose.integration-test.yaml

alexeykazakov:
[test]

alexeykazakov:
The entitlement test failed because there were too many resources for the test user in our test realm. I suspect there are some test outside of wit and auth services that create a lot of spaces/resources for the test user and don't bother to remove the spaces properly. As a result the request to obtain an RPT for all resources fails (too big response which killed by OS).
I now cleaned the resources so test are passing again (in both auth and wit).
@sbose78 we need to remove all tests that obtain all the resources available to the user as part of your work on stopping obtaining all the resources.

alexeykazakov:
@sbose78 well, this PR actually doesn't specify a new DB name. It's still using the same "postgres" name. But even if we change the DB name the "make dev" command will fail if there is already a postgres container running created by make dev for fabric8-wit.

I tried to make the make_postgres_integration_test_1 container available via 5433 but it still fails:  https://github.com/fabric8-services/fabric8-auth/pull/12/files

@hectorj2f if I apply https://github.com/fabric8-services/fabric8-auth/pull/12/files then the DB is available via 0.0.0.0:5433 but not via 172.17.0.3:5433.
But even if I set AUTH_POSTGRES_HOST to 0.0.0.0 from 172.17.0.3 then 0.0.0.0:5433 is still unavailable from the auth container. But I can connect to the DB from my host (0.0.0.0:5433).

Any idea how to fix it?

sbose78:
I was trying to use db auth + port 5432 here in this commit and then removed it because our tests failed to find the `auth` database. 

https://github.com/fabric8-services/fabric8-auth/pull/11/commits/1ab8c2f26082dae8295b1a2243893b9c6db24d2c

sbose78:
> @sbose78 we need to remove all tests that obtain all the resources available to the user as part of your work on stopping obtaining all the resources.

Okay, will do this.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-11-2`

alexeykazakov:
[test]

Issue #14, opened 2017-07-26
It should be possible to run both wit and auth services locally on the same host for development and testing 


alexeykazakov:
This is a follow up on https://github.com/fabric8-services/fabric8-auth/pull/11 which was a temporal workaround to fix our CI.

@sbose78 please work with @hectorj2f to fix our docker-test-integration.

So, we will probably need to revert https://github.com/fabric8-services/fabric8-auth/pull/11 and properly fix the test integration docker container (my attempt https://github.com/fabric8-services/fabric8-auth/pull/12/ didn't help).

alexeykazakov:
@hectorj2f if I apply https://github.com/fabric8-services/fabric8-auth/pull/12/files then the DB is available via 0.0.0.0:5433 but not via 172.17.0.3:5433.
But even if I set AUTH_POSTGRES_HOST to 0.0.0.0 from 172.17.0.3 then 0.0.0.0:5433 is still unavailable from the auth container. But I can connect to the DB from my host (0.0.0.0:5433).

hectorj2f:
@alexeykazakov I'll push a quick PR with the fix, it is an easy one.

alexeykazakov:
Oh, thanks @hectorj2f I was hopping for your docker expertise ;)

alexeykazakov:
Fixed.

PR #15, opened 2017-07-26
Auth db in port 5433

hectorj2f:
related to: https://github.com/fabric8-services/fabric8-auth/issues/14

Questions to @alexeykazakov:
- Do we want to keep the port 5433 only for tests and local environment ? That will mean that the openshift templates should be configured for 5432.

### DIFF
--- a/.make/docker-compose.integration-test.yaml
+++ b/.make/docker-compose.integration-test.yaml
@@ -4,7 +4,7 @@ services:
     image: centos/postgresql-95-centos7
     network_mode: "bridge"
     ports:
-      - "5432"
+      - "5433:5432"
     environment:
       POSTGRESQL_ADMIN_PASSWORD: mysecretpassword
     healthcheck:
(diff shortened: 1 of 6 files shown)


fabric8cd:
@hectorj2f snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-15-1`

alexeykazakov:
The tests are still failing in Jenkins: 172.17.0.3:5433: getsockopt: connection refused
I reproduce it locally to when do `make docker-test-integration`
It's the same issue I mentioned before:
the DB is available via 0.0.0.0:5433 but not via 172.17.0.3:5433
And yes, let's keep 5433 port only for test and dev env. We still can use 5432 in OS templates.

hectorj2f:
@alexeykazakov alright! Let me find a solution for the tests. I need to upgrade my docker-compose to test it locally.

fabric8cd:
@hectorj2f snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-15-2`

hectorj2f:
@alexeykazakov 
- CI tests use 5432
- local: make dev uses 5433
- Openshift deployment set a PORT env var to 5432

PR #22, opened 2017-07-27
Add the http properties to our log errors when possible

alexeykazakov:
See https://github.com/fabric8-services/fabric8-wit/pull/1536

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -14,7 +14,7 @@ import (
 	"github.com/goadesign/goa"
 )
 
-const defaultPackageName = "github.com/fabric8-services/fabric8-wit/"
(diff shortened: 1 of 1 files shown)


hectorj2f:
@alexeykazakov I discussed with Aslak, but I think we should start to extract some common package (logs, error handlers, utils, ...) out of f8-wit to avoid having multiple versions. That way we could update or fix bug in these repos without going all over our projects such as f8-tenant, f8-notification, f8-wit, f8-auth...

alexeykazakov:
@hectorj2f yep. I was thinking about it too. We should definitely start extracting such code to separate repos and use them as libraries. But until that I'm trying to pick up all relevant changes from WIT ;)

PR #24, opened 2017-07-28
WIP - Add support for github+openshiftv3 oauth - token storage

sbose78:
## API endpoints for linking 

- GET `/api/link?provider=github&scope=xxx`
- GET `/api/link?provider=openshift-v3&scope=xxx`
- GET `/api/link?provider=https://console.starter-us-east-2.openshift.com&scope=xxx`

The support for the last pattern above ensures that we are in a good position to support multicluster in future.

## API endpoints for retrieving stored tokens

- GET `/api/provider/github/token?scope=XXXX`  ( secure ) 
- GET `/api/provider/openshift-v3/token?scope=XXXX&cluster_url=https://console.starter-us-east-2.openshift.com` (secure ) 

`cluster_url` is an optional parameter. By default, it will choose `https://console.starter-us-east-2.openshift.com`

 
### Notes:

As part of the effort to support token storage for multiple OpenShift clusters, we are moving out external provider account linking out of keycloak, and maintaining the same in the Auth service.

The first step would be to replicate linking using keycloak, inside our auth service. That would include supporting oauth workflow ( and token storage ) for github & OS ( OSO to be precise )

- [x] Remove linking related endpoints which are not needed `LinkSession` , `LinkCallback`
- [x] Get `github-v3` OAuth flow working from the /link
- [x] Get `openshift-v3` OAuth flow working from the /link ( local setup using minishift https://github.com/bartoszmajsak/keycloak-openshift-auth-example )
- [ ] Store the returned token in the database.
- [x] Empty REST API to return token
- [ ] Make REST API return actual token.
- [x] DAO layer for tokens
- [x] Refactoring for having generic code for all external providers.
- [x] new configuration properties for github client & secret
- [ ] write new Tests
- [ ] fix old tests.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-24-9`

alexeykazakov:
@sbose78 please note that UI is currently using the /linksessioin endpoint. Not the /link endpoint. We had issues with setting Authorization: Bearer header in UI to be able to use /link (which requires an access token). So we will also need to refactor the UI side.

configuration/configuration.go
```diff
@@ -152,7 +156,7 @@ func (c *ConfigurationData) setConfigDefaults() {
 	//-----
 	// HTTP
 	//-----
-	c.v.SetDefault(varHTTPAddress, "0.0.0.0:8089")
+	c.v.SetDefault(varHTTPAddress, "0.0.0.0:8080")
```
alexeykazakov: I switched the default address to 8089 to be able to run both WIT and Auth locally without changing the default configurations.

configuration/configuration.go
```diff
@@ -613,6 +646,9 @@ vwIDAQAB
 	defaultKeycloakClientID = "fabric8-online-platform"
 	defaultKeycloakSecret   = "7a3d5a00-7f80-40cf-8781-b5b6f2dfd1bd"
 
+	defaultGithubClientID = "7498db1eaea3b051fbba"
```
alexeykazakov: What GitHub client is used by default?

alexeykazakov:
And one more item for the TODO list:
- Update the list for valid redirect URLs for both the GitHub and OSO client (auth.openshift.io instead of sso.openshift.io)

configuration/configuration.go
sbose78: this is temporary, will revert this.

configuration/configuration.go
sbose78: I created a test github client in my account. We should eventually use a github client details from `fabric8-services`

sbose78:
As per discussion with @aslakknutsen 

- Parallel to the context of token storage & account linking, there needs to be an API/service that helps discover which OSO/OSD cluster is being used. Where will it be? We don't know yet..
- The token storage backend needs to store info about what is the scope of the token as well.
- Token retrieval needs to happen on the basis of the scope needed by the specific operation.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-24-10`

sbose78:
> Token retrieval needs to happen on the basis of the scope needed by the specific operation.

- For a specific Github App & a specific scope, only 1 token per user is valid at any moment.
- If we need to store multiple tokens based on scope ( example, if an operation needs only `user:email` , we shouldn't be giving it a token for `org` ) , then there needs to be a separate github app for each scope.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-24-11`

alexeykazakov:
> For a specific Github App & a specific scope, only 1 token per user is valid at any moment.

The GitHub doc doesn't specify the actual limit. Why do you think it's just 1 token? Have you tried it? Or found it in another doc?

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-24-12`

PR #48, opened 2017-08-08
Use push.registry.devshift.net

hectorj2f:


fabric8cd:
@hectorj2f snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-48-1`

PR #51, opened 2017-08-12
Use DEVSHIFT_TAG_LEN to define the docker image tag size

hectorj2f:
We need to have some harmony regarding the tag size of the image generated in ci centos and cd.fabric8 platforms. That is why the cico script should rely on the value of this variable.


fabric8cd:
@hectorj2f snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-51-1`

fabric8cd:
@hectorj2f snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-51-2`

PR #60, opened 2017-08-17
Remove tenant services

alexeykazakov:
Our tenant services is outdated comparing to the same services in WIT. Let's delegate tenant service calls to WIT when needed. So, this PR removing the tenant services from Auth.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-60-1`

Issue #13, opened 2017-07-26
Remove any calls to keycloak which query 'all entitlements'


sbose78:
similar to https://github.com/fabric8-services/fabric8-wit/issues/1539

Issue #47, opened 2017-08-04
Create model classes for authorization DB tables


sbryzak:
Add the model classes as per the auth server design doc.

PR #68, opened 2017-08-23
Issue 47. Model classes for authorization DB tables

sbryzak:
This pull request contains changes for [ISSUE-47](https://github.com/fabric8-services/fabric8-auth/issues/47):

* New model classes for authorization services - resources, roles
* Migration DDL script for creation of new tables
* Unit tests for model classes (ongoing)


fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-1`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-2`

authorization/resource/resource.go
```diff
@@ -0,0 +1,223 @@
...
+	obj, err := m.Load(ctx, model.ID)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"resource_id": model.ID,
+			"ctx":         ctx,
+			"err":         err,
+		}, "unable to update the resource")
+		return errs.WithStack(err)
+	}
+	err = m.db.Model(obj).Updates(model).Error
```
alexeykazakov: I think using m.db.Save(&user) is safer than using Updates() in our Save() methods.
Save() updates all fields but Updates() updates only changed and not blank(!). So if resource.Description is an empty string then it wont be updated!
See more details here: http://jinzhu.me/gorm/crud.html#update

Overall our space resource repo https://github.com/fabric8-services/fabric8-auth/blob/master/space/space_resource.go is probably a better example of CRUD methods.

account/identity_relationship.go
```diff
@@ -0,0 +1,189 @@
...
+
+// IdentityRelationshipRepository represents the storage interface.
+type IdentityRelationshipRepository interface {
+	//repository.Exister
+	Load(ctx context.Context, ParentIdentity Identity, ChildIdentity Identity) (*IdentityRelationship, error)
+	Create(ctx context.Context, u *IdentityRelationship) error
+	Save(ctx context.Context, u *IdentityRelationship) error
+	List(ctx context.Context) ([]IdentityRelationship, error)
+	Delete(ctx context.Context, parentIdentityID uuid.UUID, childIdentityID uuid.UUID) error
+	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]IdentityRelationship, error)
```
alexeykazakov: I see there are Query() method in all repositories in this PR. Do you really need them everywhere? Are they all used?

authorization/resource/resource.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	log.Debug(nil, map[string]interface{}{
+		"resource_query": resources,
+	}, "Resource query executed successfully!")
+
+	return resources, nil
+}
+
+// First returns the first Resource element that matches the given criteria
+func (m *GormResourceRepository) First(funcs ...func(*gorm.DB) *gorm.DB) (*Resource, error) {
```
alexeykazakov: The same question as "Query". Is this First() method is going to be used? I would suggest to add only CRUD methods that we will need for sure right now. Create(), Load(), Save(), Delete() and others only if needed. If not sure right now then let's not add it. We can add it later if needed.

authorization/resource/resource.go
```diff
@@ -0,0 +1,223 @@
...
+// ResourceRepository represents the storage interface.
+type ResourceRepository interface {
+	repository.Exister
+	Load(ctx context.Context, id string) (*Resource, error)
+	Create(ctx context.Context, resource *Resource) error
+	Lookup(ctx context.Context, id string) (*Resource, error)
+	Save(ctx context.Context, resource *Resource) error
+	Delete(ctx context.Context, id string) error
+	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]Resource, error)
+	List(ctx context.Context) ([]Resource, error)
```
alexeykazakov: There is no List() implementation which makes sense. So, remove List() from interface too.

design/resource.go
```diff
@@ -0,0 +1,104 @@
...
+		)
+		a.Params(func() {
+			a.Param("resourceId", d.String, "Identifier of the resource to update")
+		})
+		a.Description("Update the details of the specified resource")
+		a.Response(d.Unauthorized, JSONAPIErrors)
+		a.Response(d.TemporaryRedirect)
+		a.Response(d.InternalServerError, JSONAPIErrors)
+		a.Response(d.BadRequest, JSONAPIErrors)
+		a.Response(d.NotFound, JSONAPIErrors)
```
alexeykazakov: We could return Forbidden too I guess?

design/resource.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+
+	a.Action("list", func() {
+		a.Routing(
+			a.GET(""),
+		)
+		a.Description("List all resources")
```
alexeykazakov: We are not going to allow to list all the resources, are we? :)

design/resource.go
```diff
@@ -0,0 +1,104 @@
...
+		a.Response(d.BadRequest, JSONAPIErrors)
+	})
+})
+
+// ResourceMedia represents a protected resource
+var ResourceMedia = a.MediaType("application/vnd.resource+json", func() {
+	a.Description("A Protected Resource")
+	a.Attributes(func() {
+		a.Attribute("resource_scopes", a.ArrayOf(d.String), "The valid scopes for this resource")
+		a.Attribute("icon_uri", d.String, "The URI containing the icon for the resource")
```
alexeykazakov: Do we really need any icons here? I know that UMA specify them. But shouldn't we just ignore them?

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-3`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-4`

design/resource.go
sbryzak: This method is required to be UMA compliant, however I'm thinking that we should just have it return an empty list.  What do you think?  Actually, we should probably support listing resources by owner.

design/resource.go
sbryzak: My opinion is that we allow the parameter in order to be compliant with UMA (while compliance might not be important right now it might be in the future) however we just don't do anything with it if one is passed in a request. So yes, we just ignore it.

account/identity_relationship.go
sbryzak: I'm actually thinking we should have more sensible "query" methods that reflect our likely use cases.  For example, it makes sense to be able to list the ResourceTypeScope records for a particular ResourceType, so we could add a ResourceTypeScope.List(ResourceType) method for that.  Same goes with listing Role objects for a particular ResourceType.  What do you think?

design/resource.go
sbryzak: I've added Forbidden.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-5`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-6`

migration/sql-files/004-authorization.sql
```diff
@@ -0,0 +1,70 @@
...
+    resource_type_id uuid NOT NULL references resource_type(resource_type_id),
+    name text NOT NULL,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    deleted_at timestamp with time zone
+);
+
+CREATE TABLE role_scope (
+    resource_type_scope_resource_type_scope_id uuid references resource_type_scope(resource_type_scope_id),
+    role_role_id uuid references role(role_id),
```
alexeykazakov: It would be "role_id" if we use "id" as primary key in "role" table instead of "role_id", correct?

account/identity_relationship.go
alexeykazakov: Sure. We can add whatever methods we need. My point was mostly about not adding anything that we are not sure if we need now.

design/resource.go
alexeykazakov: Let's not add this right now. We can add an empty list and other lists later if/when we need.

design/resource.go
alexeykazakov: Sure. I'm not saying we should forbid this param. But we don't have to declare this param to allow it.  It will be allowed anyway. But by declaring it we generate additional code (this param will be included in goa model, context, tests, etc).

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-7`

migration/sql-files/004-authorization.sql
sbryzak: Correct, but only because of gorm's naming convention defaulting to table_name_column_name without there being any possible way of overriding it.  This means that gorm would be totally broken for legacy databases in which it is not possible to change column names.

design/resource.go
sbryzak: So a JSON document passed in the request body does not have to conform with the MediaType definition declared for that method?  Won't there be some kind of validation error if the document contains properties that the MediaType does not declare?

design/resource.go
sbryzak: No problem, I've removed the endpoint definition.

migration/sql-files/004-authorization.sql
sbryzak: Oh, and other people are experiencing this same issue:

https://stackoverflow.com/questions/33437453/using-different-join-table-columns-than-defaults-with-gorm-many2many-join

design/resource.go
alexeykazakov: Oh, didn't notice it's a media type. It's not requests parameters. But anyway additional (not declared) attributes are just ignored. So, it's safe not to declare them. Payloads with undeclared attributes are still valid.

migration/sql-files/004-authorization.sql
alexeykazakov: So, another reason to use short column names for primary keys - "id"? :) I understand that's a gorm limitation but columns like resource_type_scope_resource_type_scope_id doesn't look good at all.

sbryzak:
Ok, I think I've addressed everything now.  After a lot of messing around and refactoring to compensate for bugs in gorm, I'm pretty happy with what's in this patch.  I've moved the identity_relationship stuff out of this issue and will implement it in a future issue as it is quite a comprehensive feature request.   I've also been able to fix the crazy column names for the ROLE_SCOPE table to something much more sensible.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-14`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-15`

sbryzak:
The build failed with this error, which is caused by something outside the scope of this patch:

=== RUN   TestShowUserOKUsingExpiredIfModifedSinceHeader
time="2017-08-31T12:43:40Z" level=info msg="Identity created!" identity_id=b4a38d8b-0b4a-4fac-897c-d45ad46d28a6 pkg=account req_id= 
time="2017-08-31T12:43:40Z" level=info msg="Deleting entities from 'identities' table with key b4a38d8b-0b4a-4fac-897c-d45ad46d28a6" hook_name="mighti:record-27fd9b08-627b-437d-a94b-d3122c6a7967" key=b4a38d8b-0b4a-4fac-897c-d45ad46d28a6 pkg="gormsupport/cleaner.DeleteCreatedEntities" table=identities 
time="2017-08-31T12:43:40Z" level=info msg="Deleting entities from 'users' table with key 86c3576f-7f0c-41d5-b7d1-e020bb380fe1" hook_name="mighti:record-27fd9b08-627b-437d-a94b-d3122c6a7967" key=86c3576f-7f0c-41d5-b7d1-e020bb380fe1 pkg="gormsupport/cleaner.DeleteCreatedEntities" table=users 
[info] removing callback `mighti:record-27fd9b08-627b-437d-a94b-d3122c6a7967` from /tmp/go/src/github.com/fabric8-services/fabric8-auth/gormsupport/cleaner/db_clean.go:78
--- FAIL: TestShowUserOKUsingExpiredIfModifedSinceHeader (0.01s)
	assertions.go:241: 
                        
	Error Trace:	users_blackbox_test.go:906
		
			users_blackbox_test.go:595
		
	Error:		Not equal: "tQz6146/hefpro+o3m6fOg==" (expected)
			        != "sd3AKmSL8ahq5AHI2Ca7jA==" (actual)

I'm wondering if it is an intermittent problem, because I think I have seen this happen once before locally.  Is there any way to restart the build again without pushing another commit?

sbose78:
[test]

alexeykazakov:
[test]

alexeykazakov:
That test error is inherited problem from WIT. Something is broken (in the test???) with that "if-modified" cache thing. It's not related to this PR.

authorization/resource/resource_type_scope.go
```diff
@@ -0,0 +1,215 @@
...
+
+	log.Debug(nil, map[string]interface{}{
+		"resource_type_scope_list": objs,
+	}, "Resource type scope query successfully executed!")
+
+	return objs, nil
+}
+
+// ResourceTypeScopeFilterByID is a gorm filter for Resource Type Scope ID.
+func ResourceTypeScopeFilterByID(resourceTypeScopeID uuid.UUID) func(db *gorm.DB) *gorm.DB {
```
alexeykazakov: Do we need it?

authorization/resource/resource_type_scope_blackbox_test.go
```diff
@@ -0,0 +1,154 @@
...
+	s.clean = cleaner.DeleteCreatedEntities(s.DB)
+}
+
+func (s *resourceTypeScopeBlackBoxTest) TearDownTest() {
+	s.clean()
+}
+
+func (s *resourceTypeScopeBlackBoxTest) TestOKToDelete() {
+	t := s.T()
+	res.Require(t, res.Database)
```
alexeykazakov: You don't have to require Database explicitly in gormtestsupport.DBTestSuite test suites. It's already required.

authorization/resource/resource_type_scope_blackbox_test.go
```diff
@@ -0,0 +1,154 @@
...
+	require.Nil(s.T(), err, "Could not load resource type scope")
+	assert.Equal(s.T(), resourceTypeScope.Name, updatedResourceTypeScope.Name)
+	assert.Equal(s.T(), resourceTypeScope.Description, "Collaborators may perform many operations within an area")
+}
+
+func createAndLoadResourceTypeScope(s *resourceTypeScopeBlackBoxTest) *resource.ResourceTypeScope {
+
+	resourceType := &resource.ResourceType{
+		ResourceTypeID: uuid.NewV4(),
+		Name:           "Area" + uuid.NewV4().String(),
```
alexeykazakov: Let's add the test name to all Names of all records created by tests. If the test doesn't clean the created records then it will be easy to identify what test is responsible for that. It's good to use in all DB tests.

authorization/role/identity_role_blackbox_test.go
```diff
@@ -0,0 +1,194 @@
...
+	s.clean = cleaner.DeleteCreatedEntities(s.DB)
+}
+
+func (s *identityRoleBlackBoxTest) TearDownTest() {
+	s.clean()
+}
+
+func (s *identityRoleBlackBoxTest) TestOKToDelete() {
+	t := s.T()
+	res.Require(t, res.Database)
```
alexeykazakov: Not needed.

authorization/role/role.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	log.Debug(ctx, map[string]interface{}{
+		"role_id":  u.RoleID,
+		"scope_id": s.ResourceTypeScopeID,
+	}, "Role scope created!")
+	return nil
+}
+
+// RoleFilterByID is a gorm filter for Role ID.
+func RoleFilterByID(roleID uuid.UUID) func(db *gorm.DB) *gorm.DB {
```
alexeykazakov: It's not used, right?

authorization/role/role_blackbox_test.go
```diff
@@ -0,0 +1,182 @@
...
+	s.clean = cleaner.DeleteCreatedEntities(s.DB)
+}
+
+func (s *roleBlackBoxTest) TearDownTest() {
+	s.clean()
+}
+
+func (s *roleBlackBoxTest) TestOKToDelete() {
+	t := s.T()
+	res.Require(t, res.Database)
```
alexeykazakov: You can remove it.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-68-16`

sbryzak:
[test]

sbryzak:
[test]

sbryzak:
I keep getting test failures in the CI build, the latest one is this:

=== RUN   TestCreateAndDeletePermissionOK
time="2017-09-01T12:00:10Z" level=error msg="unable to obtain token" file="/tmp/go/src/github.com/fabric8-services/fabric8-auth/controller/token.go" func="github.com/fabric8-services/fabric8-auth/controller.GenerateUserToken" line=198 pid=10246 pkg=controller req_id= response_body="{\"error\":\"invalid_grant\",\"error_description\":\"Invalid user credentials\"}" response_status="401 Unauthorized" 
--- FAIL: TestCreateAndDeletePermissionOK (0.40s)
	assertions.go:241: 
                        
	Error Trace:	authz_blackbox_test.go:492
		
			authz_blackbox_test.go:437
		
			authz_blackbox_test.go:137
		
	Error:		Expected nil, but got: errors.InternalError{Err:unable to obtain toke. Response status: 401 Unauthorized. Responce body: }

alexeykazakov:
[test]

alexeykazakov:
[test]

alexeykazakov:
It was a mess in our test realm in KC. I cleaned it up now it should be better. We need to fix https://github.com/fabric8-services/fabric8-auth/issues/13 to prevent such failures.

Issue #83, opened 2017-09-01
Auth service account


alexeykazakov:
- [x] New Private/Public keys to be used to sign service account tokens
- [x] Provide endpoint to expose the public keys https://github.com/fabric8-services/fabric8-auth/issues/30
- [x] Load public key from KC instead storing in configuration https://github.com/fabric8-services/fabric8-auth/issues/29
- [x] Generate an auth service account token to use it for communicating with WIT when relevant https://github.com/fabric8-services/fabric8-auth/pull/96
- [x] Switch WIT to use the public keys exposed by Auth to verify tokens https://github.com/fabric8-services/fabric8-wit/pull/1623

PR #96, opened 2017-09-12
Auth Service Account

alexeykazakov:
New method of `token.Manager`: `ServiceAccountToken(req *goa.RequestData)` returns the service account token which authenticates the Auth service. This token can be used to call other services when we need to make sure that the request is coming from the Auth service. For example when we need to call WIT to update users.

An example of usage. Auth calls WIT to update an user:

1. Auth receives a request from UI to update the user.
2. Auth updates the user in Auth model/DB/KC/etc. And now needs to update WIT:
3. Auth gets the auth service account token from the token manager: tokenManager.ServiceAccountToken(req *goa.RequestData) and calls WIT to update the user passing that service account token.
4. WIT receives the request from Auth and verify the token using the public keys loaded from Auth.
5. WIT should also check if the "service_accountname" claim from the token == "auth" to make sure it authenticates the Auth service.

Part of https://github.com/fabric8-services/fabric8-auth/issues/83


token/token.go
```diff
@@ -3,21 +3,27 @@ package token
...
 	"github.com/pkg/errors"
-	uuid "github.com/satori/go.uuid"
-	jose "gopkg.in/square/go-jose.v2"
+	"github.com/satori/go.uuid"
+	"gopkg.in/square/go-jose.v2"
+	"time"
+)
+
+const (
+	ServiceAccountID = "8f558668-4db7-4280-8e65-408bcb95f9d9"
```
sbose78: Could we maintain this as part of configuration ?

alexeykazakov: Yeah.. good point. Will move it to configuration.

alexeykazakov: But actually, do we really need to keep it in configuration? We don't really use the service account ID. It's mostly just to fill up the sub claim. WIT will check the "service_accountname" claim.
Is there any use case when will will need configure the auth service account ID?
It's similar to the hardcoded SystemSpace ID in WIT.

sbose78: @alexeykazakov , that's a fair point. I just feel a bit uneasy maintaining hard-coded strings this way ( but configuration doesn't feel like the right place either.. )

Issue #98, opened 2017-09-14
Limit users search result


alexeykazakov:
We should limit users search result not to expose the number of users in the system.
Let's set it to 50 by default.

PR #99, opened 2017-09-14
Limit users search result

alexeykazakov:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/98


configuration/configuration.go
```diff
@@ -175,6 +176,9 @@ func (c *ConfigurationData) setConfigDefaults() {
 	c.v.SetDefault(varKeycloakTesUserName, defaultKeycloakTesUserName)
 	c.v.SetDefault(varKeycloakTesUserSecret, defaultKeycloakTesUserSecret)
 
+	// Max number of users returned when searching users
+	c.v.SetDefault(varUsersListLimit, 50)
```
sbose78: Effectively the number 50 should be fine.. since the UI ( collaborators' ) page doesn't require more than this.

PR #108, opened 2017-09-20
Don't fetch profileURL from WIT

alexeykazakov:
When creating a new identity in Auth which already exists in WIT we should not set profileURL for such identity in Auth. ProfileURL is returned as an empty string in User's result payload which causes `failed to create user/identity pq: duplicate key value violates unique constraint \"uix_identity_profileurl\"` error in Auth DB. The profileURL should be null. Not an empty string.
We could check if profileURL is an empty sting and insert a null instead but profileURL is not currently used anyway (it's supposed to be used in remote, like GitHub, identities). So, let's just ignore this field in that temporal code.

### DIFF
--- a/remoteservice/wit.go
+++ b/remoteservice/wit.go
@@ -160,8 +160,7 @@ func (r *RemoteWITServiceCaller) GetWITUser(ctx context.Context, req *goa.Reques
 		}
 
 		identity = &account.Identity{
-			ProfileURL: witServiceUser.Data.Attributes.URL,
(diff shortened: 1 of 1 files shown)


alexeykazakov:
Without this PR it's impossible to login/create more than one user/identity to Auth if these users/identities also exist in WIT.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-108-1`

PR #111, opened 2017-09-21
fix panic when user not found

alexeykazakov:

### DIFF
--- a/controller/users.go
+++ b/controller/users.go
@@ -321,8 +321,8 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
 
 	if err != nil {
 		log.Error(ctx, map[string]interface{}{
-			"user_name": id.String(),
-			"err":       err,
+			"identity_id": id.String(),
+			"err":         err,
 		}, "failed to update user/identity")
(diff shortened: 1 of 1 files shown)


controller/users.go
```diff
@@ -321,7 +321,7 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
 
 	if err != nil {
 		log.Error(ctx, map[string]interface{}{
-			"user_name": identity.ID,
+			"user_name": id.String(),
```
sbose78: let's call it `identity_id`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-111-1`

controller/users.go
alexeykazakov: Makes sense :)
Done.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-111-2`

PR #117, opened 2017-09-25
Use *.openshift.io as valid redirect URL by default

alexeykazakov:
Use *.openshift.io as valid redirect URL by default even if the corresponding env var is not set.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-117-1`

PR #120, opened 2017-09-26
Space onwer should not be allowed to be removed from the list of collaborators

alexeykazakov:
Instead of checking if the user is the only collaborator in the space we should check if the user is the space owner.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-120-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-120-2`

PR #122, opened 2017-09-26
allow localhost for redirects by default

alexeykazakov:
As a temporal solution to unblock local forge development


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-122-1`

PR #124, opened 2017-09-27
Check token claims if user is approved

alexeykazakov:
Instead of calling KC.

### DIFF
--- a/token/token.go
+++ b/token/token.go
@@ -19,11 +19,15 @@ import (
 	"time"
 
 	"github.com/dgrijalva/jwt-go"
+	"github.com/fabric8-services/fabric8-auth/app"
 	"github.com/goadesign/goa"
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
 	"github.com/pkg/errors"
 	"github.com/satori/go.uuid"
+	"golang.org/x/oauth2"
 	"gopkg.in/square/go-jose.v2"
+	"net/url"
+	"strconv"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-124-1`

token/token.go
```diff
@@ -425,3 +448,64 @@ func ReadManagerFromContext(ctx context.Context) (*tokenManager, error) {
...
+	}
+	result, err := strconv.ParseInt(fmt.Sprintf("%v", number), 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return result, nil
+}
+
+// EncodeToken encodes token
+func EncodeToken(ctx context.Context, referrer *url.URL, outhToken *oauth2.Token) error {
```
sbose78: ( since you are refactoring these parts )
Any reason why we aren't returning the "token_json" ie, `string(b)` from this method , and then we could do the URL related stuff outside.

Also, a pass by reference of `referrer *url.URL` makes it a bit confusing to understand when seen from a package docs perspective what exactly this function is returning. Could we add comments on top of this method to address this, Or, maybe refactor to return a `url.URL` explicitly.

alexeykazakov: Good point. Will address it in https://github.com/fabric8-services/fabric8-auth/pull/126 to avoid additional conflicts.

Issue #125, opened 2017-09-28
Issue API tokens for unapproved users


alexeykazakov:
We should support workflow when we issue tokens for unapproved users to be used for API (like analytics)

Issue #34, opened 2017-07-28
Create tests to cover full auth flow


alexeykazakov:
Currently our tests cover only the first part of the flow when we call /api/login/authorize to create a state and redirect to KC. We need to cover the second part when after successful authentication KC redirects back to /api/login/authorize and passing the state from the first call.

Moved from https://github.com/fabric8-services/fabric8-wit/issues/1277

Issue #46, opened 2017-08-03
External tools (JBossTools) integration requirements


jeffmaury:
This is a place holder to follow all the requirements for the external tools (JBossTools...) auth integration.

As of yet the only issue is to get a well defined landing page: openshiftio/openshift.io#466

alexeykazakov:
This seems to be done. Please re-open if you still need something from our end.

PR #126, opened 2017-09-28
New "api_client" param for /api/login

alexeykazakov:
- Added new `/api/login?api_client=<anystring>` param to issue a set of tokens to be used for api calls even if user is not approved yet.
- If `api_client` is present and not an empty string then the token json will be encoded in `api_token` param instead of `token_json`
- If user is not approved then we do not create a new user/identity in the Auth DB.
- Added tests to cover login flow.

It's based on https://github.com/fabric8-services/fabric8-auth/pull/124 (so, contains the same commit) which can be merged separately.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/125
Fixes https://github.com/fabric8-services/fabric8-auth/issues/34
Related to https://github.com/fabric8-services/fabric8-auth/issues/46

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/fabric8-services/fabric8-auth/account"
 	"github.com/fabric8-services/fabric8-auth/app"
-	"github.com/fabric8-services/fabric8-auth/auth"
 	"github.com/fabric8-services/fabric8-auth/errors"
 	"github.com/fabric8-services/fabric8-auth/jsonapi"
 	"github.com/fabric8-services/fabric8-auth/log"
@@ -85,22 +84,22 @@ func (c *TokenController) Refresh(ctx *app.RefreshTokenContext) error {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, errs.New(res.Status+" "+rest.ReadBody(res.Body))))
 	}
 
-	token, err := auth.ReadToken(ctx, res)
+	t, err := token.ReadTokenSet(ctx, res)
 	if err != nil {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 	ctx.ResponseData.Header().Set("Cache-Control", "no-cache")
-	return ctx.OK(convertToken(*token))
+	return ctx.OK(convertToken(*t))
 }
 
-func convertToken(token auth.Token) *app.AuthToken {
+func convertToken(t token.TokenSet) *app.AuthToken {
 	return &app.AuthToken{Token: &app.TokenData{
-		AccessToken:      token.AccessToken,
-		ExpiresIn:        token.ExpiresIn,
-		NotBeforePolicy:  token.NotBeforePolicy,
-		RefreshExpiresIn: token.RefreshExpiresIn,
-		RefreshToken:     token.RefreshToken,
-		TokenType:        token.TokenType,
+		AccessToken:      t.AccessToken,
+		ExpiresIn:        t.ExpiresIn,
+		NotBeforePolicy:  t.NotBeforePolicy,
+		RefreshExpiresIn: t.RefreshExpiresIn,
+		RefreshToken:     t.RefreshToken,
+		TokenType:        t.TokenType,
 	}}
 }
 
@@ -179,7 +178,7 @@ func GenerateUserToken(ctx context.Context, tokenEndpoint string, configuration 
 		}, "unable to obtain token")
 		return nil, errors.NewInternalError(ctx, errs.Errorf("unable to obtain toke. Response status: %s. Responce body: %s", res.Status, rest.ReadBody(res.Body)))
 	}
-	token, err := auth.ReadToken(ctx, res)
+	t, err := token.ReadTokenSet(ctx, res)
 	if err != nil {
 		log.Error(ctx, map[string]interface{}{
 			"token_endpoint": res,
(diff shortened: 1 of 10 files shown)


alexeykazakov:
@jeffmaury @samuzzal-choudhury does it look good to you?

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-126-1`

jeffmaury:
If the value is left open what about passing jbosstools ou devstudio for the eclipse side and vscode for VSCode to allow analytics ?

controller/token.go
```diff
@@ -142,7 +133,7 @@ func (c *TokenController) Generate(ctx *app.GenerateTokenContext) error {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, errs.Wrap(err, "unable to generate test token")))
 	}
 	// Creates the testuser2 user and identity if they don't yet exist
-	_, _, err = c.Auth.CreateOrUpdateIdentity(*testuser.Token.AccessToken, ctx, profileEndpoint)
+	_, _, err = c.Auth.CreateOrUpdateIdentity(ctx, *testuser.Token.AccessToken)
```
sbose78: Good refactoring - `ctx` ideally should be the first param.

login/service.go
```diff
@@ -82,11 +81,16 @@ var allProvidersToLink = []string{"github", "openshift-v3"}
...
 const (
 	initiateLinkingParam = "initlinking"
+	apiTokenParam        = "api_token"
 )
 
-// Perform performs authentication
-func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, config *oauth2.Config, serviceConfig LoginServiceConfiguration) error {
+type OauthConfig interface {
+	Exchange(ctx netcontext.Context, code string) (*oauth2.Token, error)
+	AuthCodeURL(state string, opts ...oauth2.AuthCodeOption) string
```
sbose78: Should we call it  `AuthorizationGrant`  to align it with Oauth terminologies.

invincibleJai:
@alexeykazakov so we would be passing two parameters in query param as below?

`/api/login?api_token=<anystring>&redirect=<redirect_uri>`

for VSCode extension , `api_token=vscode` looks fine to me.

alexeykazakov:
> If the value is left open what about passing jbosstools ou devstudio for the eclipse side and vscode for VSCode to allow analytics ?

@jeffmaury yes, this is why I added this param as a string, not a boolean.

alexeykazakov:
> so we would be passing two parameters in query param as below?

@invincibleJai correct.

login/service.go
alexeykazakov: I called this OuathConfig bc I needed some interface to replace the oauth2.Config implementation by some interface.

design/login.go
```diff
@@ -20,6 +20,7 @@ var _ = a.Resource("login", func() {
 				a.Enum("offline_access")
 				a.Description("If scope=offline_access then an offline token will be issued instead of a regular refresh token")
 			})
+			a.Param("api_token", d.String, "If not an empty string then return a token for accessing API")
```
sbose78: Could we call it `api_client` instead ?

login/service.go
sbose78: No, that part's fine. Could we replace `AuthCodeURL` with `AuthorizationGrant` .

design/login.go
alexeykazakov: Yeah.. maybe api_client is better.. I don't like api_token much but couldn't come up with anything better ;)

login/service.go
alexeykazakov: Oh, no, we can't :) It's how they call this method in oauth2.Config. I had to use our own interface here instead of oauth2.Config because I needed to pass a dummy implementation here in tests.
So, I added two methods to this interface from the original oauth2.Config type. The methods which we use.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-126-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-126-3`

design/login.go
alexeykazakov: Renamed!

alexeykazakov:
@jeffmaury, @invincibleJai, please note that I renamed the param `api_token` to suggested by @sbose78 `api_client` which I think suites better especially if we send the name of the IDE extension as a value of this param.
`/api/login?api_client=vscode&redirect=<redirect_uri>` for example.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-126-4`

alexeykazakov:
I did one more change: 
- If `api_client` is present and not an empty string then the token json will be encoded in `api_token` param instead of `token_json`

invincibleJai:
thanks @alexeykazakov got it

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-126-5`

Issue #134, opened 2017-10-04
Add API to Link accounts.


sbose78:
## Linking Accounts

```
POST /api/token/link
Content-Type: application/x-www-form-urlencoded
Payload: for=<resource>&token=<access_token>&scope=<scope>&redirect=<redirect_url>
```

**`<resource>`** - Resource we need to link accounts for. For example https://github.com/somecoolrepo or https://console.starter-us-east-2.openshift.com/console/project/coolproject
**`<scope>`** - required scope. Multiple scopes can be specified by separating them with a space. Optional. If not defined then the default scope is used. Not supported in the first version!
**`<redirect>`** - after successful linking the client will be redirected to this URL. If not specified then the URL from the “Referer” header will be used. If both “Referer” header and “redirect” param are missing then a Bar Request response will be returned.
**`<access_token>`** - user’s access token

If the token for such user already exists then the account will be re-linked and the token will be updated.
In the first version, we support only **github** and **openshift** represents OSO-us-starter-2. Later an individual **openshift** provider will be associated with the user during signup by the registration app, as part of multicluster support.

**Example:**
```
POST /api/token/link
Content-Type: application/x-www-form-urlencoded
for=https://github.com/somecoolrepo&token=ABSDEF12345678990&redirect=https%3A%2F%2Fopenshift.io&scope=user%20public_repo
```

## Re-linking Accounts

If some services catches 401 when trying to use Git or OS token this service should return 401 to UI with the following header:

`WWW-Authenticate: Link url=<link_url_optional>, description=”<description_optional>”`

UI should initiate re-linking when catches such 401 response.

alexeykazakov:
I've updated the issue description.

PR #136, opened 2017-10-05
REST API for account linking

alexeykazakov:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/134

- [x] Tests


design/token.go
```diff
@@ -50,6 +50,45 @@ var _ = a.Resource("token", func() {
...
+	})
+})
+
+var linkPayload = a.Type("LinkPayload", func() {
+	a.Attribute("for", d.String, "Resource we need to link accounts for", func() {
+		a.Example("https://github.com/somecoolrepo")
+	})
+	a.Attribute("token", d.String, "User's access token")
+	a.Attribute("redirect", d.String, "URL to be redirected to after successful account linking. If not set then will redirect to the referrer instead.")
+	//a.Attribute("scope", d.String, "Required scope. Multiple scopes can be specified by separating them with a space. If not defined then the default scope is used.")
```
sbose78: My not-so-strong opinion on this is : we can keep the `scope` but make it optional and always default it to our current scopes. :)

design/token.go
```diff
@@ -50,6 +50,45 @@ var _ = a.Resource("token", func() {
...
+	a.Required("token")
+	a.Required("for")
```
sbose78: this works too `a.Required("token", "for")`

design/token.go
```diff
@@ -50,6 +50,45 @@ var _ = a.Resource("token", func() {
...
+		a.Response(d.BadRequest, JSONAPIErrors)
+		a.Response(d.InternalServerError, JSONAPIErrors)
+	})
+})
+
+var linkPayload = a.Type("LinkPayload", func() {
+	a.Attribute("for", d.String, "Resource we need to link accounts for", func() {
+		a.Example("https://github.com/somecoolrepo")
+	})
+	a.Attribute("token", d.String, "User's access token")
```
sbose78: refresh my memory please - why aren't we picking up from the header?

token/provider/github.go
```diff
@@ -0,0 +1,23 @@
...
+type GitHubConfig struct {
+	oauth2.Config
+}
+
+func NewGitHubConfig(clientID string, clientSecret string, authURL string) *GitHubConfig {
+	provider := &GitHubConfig{}
+	provider.ClientID = clientID
+	provider.ClientSecret = clientSecret
+	provider.Endpoint = github.Endpoint
+	provider.RedirectURL = fmt.Sprintf("%s%s", authURL, client.CallbackTokenPath())
```
sbose78: curious, is this the only use-case we have for the `client/...` package?

token/provider/provider_service.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	redirectLocation := config.AuthCodeURL(stateID.String(), oauth2.AccessTypeOnline)
+	return redirectLocation, nil
+}
+
+// Callback returns a redirect URL after callback from an external oauth2 resource provider such as GitHub during user's account linking
+func (service *LinkService) Callback(ctx context.Context, req *goa.RequestData, state string, code string) (string, error) {
+	// validate known state
+	knownReferrer, err := LoadReferrer(ctx, service.db, state)
```
sbose78: before this, add a check that `state!="" and code!=""`

design/token.go
alexeykazakov: Because we can't use AJAX to initiate linking. So we have to use window.location = ... (GET) or submit a form (POST). In both cases we can't add headers to the request. Form/Post is preferable because we don't want to pass tokens as part of URL.

token/provider/github.go
alexeykazakov: Correct.

alexeykazakov:
Ready for review! But I still need to add tests.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-8`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-9`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-10`

token/link/link.go
```diff
@@ -0,0 +1,217 @@
...
+		log.Error(ctx, map[string]interface{}{
+			"state":       state,
+			"code":        code,
+			"provider_id": oauthProvider.ID(),
+		}, "access token return by provider is empty")
+		return "", errors.New("access token return by provider is empty")
+	}
+
+	err = application.Transactional(service.db, func(appl application.Application) error {
+		tokens, err := appl.ExternalTokens().LoadByProviderIDAndIdentityID(ctx, oauthProvider.ID(), identityUUID)
```
sbose78: Thinking we should check with "ProviderID + IdentityID + Scope" ? The scope will be default anyway for now. Let me update the repository API for that since this is the most popular use case "Get token for XXX provider for YYY identity for the scope ZZZ".

And if some token in the same scope was already present, we should overwrite it - else `create` a new entity in the table.

token/link/github.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/satori/go.uuid"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/github"
+)
+
+const (
+	gitHubProviderID = "2f6b7176-8f4b-4204-962d-606033275397" // Do not change! This ID is used as provider ID in the external token table
+)
+
+type GitHubConfig struct {
```
sbose78: Could you add a "Type" string which would be set to "git"  - we would need the provider type string when we return the header 
```
WWW-Authenticate: Link provider=git, url=https://auth.openshift.io/api/token/link?redirect=https%3A%2F%2Fopenshift.io?scope=user%20public_repo, description=”Git token is missing. Link Git account”
```

token/link/link.go
alexeykazakov: The scope thing is more complicated actually. For example git has limitation of number of tokens per user&app. So, if we just blindly add new tokens with new scopes we may invalidate the existing tokens. Another concern is UX around it. It would be a bad UX if user just linked GitHub and granted a write access to OSIO (let's say in new app wizard) and we ask the same user later for read access when another client (Che for example) asked for it.
Token scopes will be in DB so we could utilize it later but for now I wouldn't do any logic around it.

token/link/github.go
alexeykazakov: Sure, we can add a type string here. But it's rather a github type because different git providers will have different types: github, gitlab, etc.
But why would you need a provider type in WWW-Authenticate:? How it would help?

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-11`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-12`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-13`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-14`

alexeykazakov:
Simple app to test GitHub linking: https://github.com/alexeykazakov/linker

token/link/link.go
```diff
@@ -0,0 +1,233 @@
...
+func (service *LinkService) NewOauthProvider(ctx context.Context, req *goa.RequestData, config LinkConfig, forResource string) (ProviderConfig, error) {
+	authURL := rest.AbsoluteURL(req, "")
+
+	resourceURL, err := url.Parse(forResource)
+	if err != nil {
+		return nil, err
+	}
+	if resourceURL.Host == "github.com" {
+		return NewGitHubConfig(config.GetGitHubClientID(), config.GetGitHubClientSecret(), config.GetGitHubClientDefaultScopes(), authURL), nil
+	} else if strings.HasPrefix(forResource, config.GetOpenShiftClientApiUrl()) {
```
sbose78: This implies we are supporting : 
```
`for=https://<IP_Address>` 
```

Would be handy for local setups using minishift!

alexeykazakov: Yeah... this was intended :)

token/link/link.go
sbose78: > For example git has limitation of number of tokens per user&app. So, if we just blindly add new tokens with new scopes we may invalidate the existing tokens

haha yeah, this actually puts the whole per-scope-per-token plan in jeopardy :(

token/link/github.go
sbose78: I would need the type string to generate this message : 

"WWW-Authenticate: Link provider=github, url=https://auth.openshift.io/api/token/link?redirect=https%3A%2F%2Fopenshift.io?scope=user%20public_repo, description=”Github token is missing. Link Github account”

All the places were we use "github" would be replaced by "openshift" , which is going away but still might need in future.

controller/token.go
```diff
@@ -189,3 +191,50 @@ func GenerateUserToken(ctx context.Context, tokenEndpoint string, configuration
...
+		redirectURL = ctx.RequestData.Header.Get("Referer")
+		if redirectURL == "" {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("redirect", "empty").Expected("redirect param or Referer header should be specified"))
+		}
+	} else {
+		redirectURL = *ctx.Payload.Redirect
+	}
+
+	if !c.Configuration.IsOpenShiftLinkingEnabled() && strings.HasPrefix(ctx.Payload.For, c.Configuration.GetOpenShiftClientApiUrl()) {
+		// OSO account linking is disabled by default in Dev Mode.
```
sbose78: otherwise, we'd have to run minishift in a container via docker-compose :P

token/link/github.go
alexeykazakov: But how the provider type name would help the client? Why do you want to include it in the header?

configuration/configuration.go
```diff
@@ -178,6 +186,13 @@ func (c *ConfigurationData) setConfigDefaults() {
 	c.v.SetDefault(varKeycloakDomainPrefix, defaultKeycloakDomainPrefix)
 	c.v.SetDefault(varKeycloakTesUserName, defaultKeycloakTesUserName)
 	c.v.SetDefault(varKeycloakTesUserSecret, defaultKeycloakTesUserSecret)
+	c.v.SetDefault(varGitHubClientID, "c6a3a6280e9650ba27d8")
+	c.v.SetDefault(varGitHubClientSecret, "48d1498c849616dfecf83cf74f22dfb361ee2511")
+	c.v.SetDefault(varGitHubClientDefaultScopes, "admin:repo_hook read:org repo user gist")
+	c.v.SetDefault(varOSOClientApiUrl, "https://api.starter-us-east-2.openshift.com")
```
sbose78: We have to whitelist localhost:8089 here I suppose.

token/link/link.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	forResource := referrerURL.Query().Get(forParam)
+
+	oauthProvider, err := service.providerFactory.NewOauthProvider(ctx, req, service.config, forResource)
+	if err != nil {
+		return "", err
+	}
+
+	providerToken, err := oauthProvider.Exchange(ctx, code)
```
sbose78: For local minishift - had to add this.
```
tr := &http.Transport{
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
	}
	sslcli := &http.Client{Transport: tr}
	ctx = context.WithValue(ctx, oauth2.HTTPClient, sslcli)
```

sbose78: export AUTH_OSO_CLIENT_ID=openshift-v3-authentication
export AUTH_OSO_SECRET=1234
export AUTH_OSO_CLIENT_APIURL=https://192.168.99.100:8443
export AUTH_OSO_LINKING_ENABLED=true

I got past the login page and then gave permissions. 
but got this error ( which im guessing is an openshift/minishift setup issue ) 

```
{"errors":[{"code":"unknown_error","detail":"oauth2: cannot fetch token: 403 Forbidden\nResponse: {\n  \"kind\": \"Status\",\n  \"apiVersion\": \"v1\",\n  \"metadata\": {},\n  \"status\": \"Failure\",\n  \"message\": \"User \\\"system:anonymous\\\" cannot \\\"post\\\" on \\\"/oauth/access_token\\\"\",\n  \"reason\": \"Forbidden\",\n  \"details\": {},\n  \"code\": 403\n}\n","status":"500","title":"Unknown error"}]}
```

Could you tell me your setup config please ? That way I could test it out locally too.

token/link/link.go
sbose78: This worked!

token/link/github.go
alexeykazakov: I added provider type name - https://github.com/fabric8-services/fabric8-auth/pull/136/commits/83df82b919fa362f0da659472914802d6bb6f96c

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-16`

token/link/link.go
alexeykazakov: Addressed the TLS verification issue here - https://github.com/fabric8-services/fabric8-auth/pull/136/commits/50f3e1b723e3aa6797d9c6f998f7e13d87600967

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-136-17`

PR #141, opened 2017-10-09
Move remoteservices/wit.go to wit/wit.go

alexeykazakov:
IMO it makes sense to keep all wit related code in the wit package instead of spreading it between remoteservices and wit/witservices


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-141-1`

PR #142, opened 2017-10-10
External token provider ID index should be case sensitive

alexeykazakov:
It doesn't make any sense to convert provider ID to low case because we use UUID as provider IDs

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -3,6 +3,7 @@ package migration
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"database/sql"
 	"net/http"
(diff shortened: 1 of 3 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-142-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-142-2`

Issue #138, opened 2017-10-09
REST API for removing linked/external tokens


alexeykazakov:
We need to provide REST API so users can delete their external/linked tokens. This action should hard delete linked tokens with proper logging.
Related to https://github.com/fabric8-services/fabric8-auth/issues/132

PR #152, opened 2017-10-25
REST API for removing linked/external tokens

alexeykazakov:
- [x] Hard delete tokens in repo
- [x] Tests for hard deleting tokens in repo
- [x] REST API
- [x] Tests for REST API

Fixes https://github.com/fabric8-services/fabric8-auth/issues/138

### DIFF
--- a/application/repository/exister.go
+++ b/application/repository/exister.go
@@ -17,17 +17,26 @@ type Exister interface {
 	CheckExists(ctx context.Context, id string) error
 }
 
-// Exists returns true if an item exists in the database table with a given ID
-func Exists(ctx context.Context, db *gorm.DB, tableName string, id string) (bool, error) {
+// Exists returns true if a soft or hard deletable item exists in the database table with a given ID
+func Exists(ctx context.Context, db *gorm.DB, tableName string, id string, softDeletable bool) (bool, error) {
 	var exists bool
-	query := fmt.Sprintf(`
+	var query string
+	if softDeletable {
+		query = fmt.Sprintf(`
 		SELECT EXISTS (
 			SELECT 1 FROM %[1]s
 			WHERE
 				id=$1
 				AND deleted_at IS NULL
 		)`, tableName)
-
+	} else {
+		query = fmt.Sprintf(`
+		SELECT EXISTS (
+			SELECT 1 FROM %[1]s
+			WHERE
+				id=$1
+		)`, tableName)
+	}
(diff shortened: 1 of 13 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-152-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-152-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-152-3`

alexeykazakov:
[test]

kishansagathiya:
@alexeykazakov @sbose78 I tried it and the api works for me (i.e deletes the externel_token from the database).

alexeykazakov:
@kbsingh thanks for testing! Btw, there is a simple app to link GitHub accounts - https://github.com/alexeykazakov/linker (it won't work if you are not logged into GitHub because of some GitHub limitations of the length of Referer header but it should work in OSIO).

@sbryzak thanks for review!

application/repository/exister.go
```diff
@@ -17,17 +17,26 @@ type Exister interface {
 	CheckExists(ctx context.Context, id string) error
 }
 
-// Exists returns true if an item exists in the database table with a given ID
-func Exists(ctx context.Context, db *gorm.DB, tableName string, id string) (bool, error) {
+// Exists returns true if a soft or hard deletable item exists in the database table with a given ID
+func Exists(ctx context.Context, db *gorm.DB, tableName string, id string, softDeletable bool) (bool, error) {
```
sbose78: Could we something like `isSoftDeleted` ?
That would imply "check if object exists given that the object is soft deleted"

alexeykazakov:
> All looks good to me! My only feedback is what we already discussed about overloading the Exists() method to make it easier to call

The reason why added a new parameter instead of adding a new method something like ExistsHardDeletable() and keep Exists() for soft deletable only is that I think it's more error pron. If the current Exists() calls for a table which doesn't support soft deletes then it will fail in runtime. But with additional required parameter it's harder to misuse this method.

I had to add a new CheckHardDeletableExists() because the current CheckExists() is widely used and adding a new parameter would require refactoring all our repos.

Issue #153, opened 2017-10-25
Allow service accounts to be configured


sbryzak:
For the purpose of authenticating the resource servers (such as Work Item Tracker) that consume fabric8-auth services, it is necessary to be able to configure service accounts for these.

A new configuration file, **fabric8-auth-service-account-secrets.conf** will be used to configure the service accounts.  An entry in this file may look like this:

```
{
    "accounts": [
        {
            "name":"Work Item Tracker",
            "id":"wit-userid",
            "credentials":"0123456789abcdef"
        }
    ]
}
```

By default, fabric8-auth will look for this file in the **/etc/fabric8/** folder, failing to find it there it will search for it in the running directory, followed by the **/conf** directory (relative to the running dir), and finally failing that in the **../conf** directory.

The location of the configuration file may be overridden by an environment parameter, or a command line parameter.

This is a sub-issue of #72

alexeykazakov:
@sbryzak I think we should be able to configure a pair of credentials for each account, so, we could rotate them:
```

{
    "accounts": [
        {
            "name":"Work Item Tracker",
            "id":"wit-userid",
            "credentials": ["0123456789abcdef", "sdfhsldfjdfjlj"]
        }
    ]
}
```
1. AuthZ service checks all available credentials during SA "login".
2. When we need to update credentials we just add a new one to the array.
3. Update the credentials in the client service.
4. Remove the old credentials from the AuthZ service.

So, we have changed the credentials without any downtime. WDYT?

alexeykazakov:
I also renamed credentials to secrets ^^^

PR #158, opened 2017-10-27
ISSUE-153 Allow service accounts to be configured

sbryzak:
**WIP - There's still some problems with tests that I am working on.**

This PR adds support for configuring service accounts. See #153

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -96,6 +96,7 @@ type serviceAccountConfig struct {
 	Accounts []ServiceAccount
 }
 
+// ServiceAccount represents a service account configuration
 type ServiceAccount struct {
 	Name    string   `mapstructure:"name"`
 	ID      string   `mapstructure:"id"`
@@ -216,6 +217,7 @@ func getServiceAccountConfigFile() string {
 	return envServiceAccountConfigFile
 }
 
+// GetServiceAccounts returns a map of service account configurations by service account ID
 func (c *ConfigurationData) GetServiceAccounts() map[string]ServiceAccount {
 	return c.sa
 }
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -138,6 +138,7 @@ func NewConfigurationData(mainConfigFile string, serviceAccountConfigFile string
 	saViper.SetTypeByDefaultValue(true)
 
 	var err error
+	var etcSAConfigUsed bool
 	if serviceAccountConfigFile != "" {
 		// If a service account configuration file has been specified, check that it exists
 		if _, err := os.Stat(serviceAccountConfigFile); err != nil {
@@ -150,6 +151,13 @@ func NewConfigurationData(mainConfigFile string, serviceAccountConfigFile string
 		if err != nil {
 			return nil, err
 		}
+		etcSAConfigUsed = serviceAccountConfigFile != ""
+	}
+
+	if !etcSAConfigUsed {
+		log.WithFields(map[string]interface{}{
+			"default_sa_conf_path": defaultServiceAccountConfigPath,
+		}).Warningln("Default service account config file path is not used!")
 	}
 
 	saViper.SetConfigType("json")
@@ -162,6 +170,10 @@ func NewConfigurationData(mainConfigFile string, serviceAccountConfigFile string
 		saViper.ReadConfig(bytes.NewBuffer(data))
 	} else {
 		saViper.SetConfigFile(serviceAccountConfigFile)
+		err := saViper.ReadInConfig()
+		if err != nil {
+			return nil, errors.Errorf("failed to load the sa config file: %s \n", err)
+		}
 	}
 
 	var conf serviceAccountConfig
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -8,12 +8,12 @@ import (
 
 	"github.com/fabric8-services/fabric8-auth/rest"
 
+	"bytes"
 	"github.com/goadesign/goa"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
-	"path/filepath"
 )
 
 // String returns the current configuration as a string
@@ -90,9 +90,6 @@ const (
 	varLogJSON                              = "log.json"
 	varWITDomainPrefix                      = "wit.domain.prefix"
 	varWITURL                               = "wit.url"
-
-	varServiceAccountConfigName    = "service-account-secrets.conf"
-	varDefaultServiceAccountConfig = "/etc/fabric8/" + varServiceAccountConfigName
 )
 
 type ServiceAccountConfig struct {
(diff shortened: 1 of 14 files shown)


configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+		c.v.SetConfigFile(mainConfigFile)
 		err := c.v.ReadInConfig() // Find and read the config file
 		if err != nil {           // Handle errors reading the config file
 			return nil, errors.Errorf("Fatal error config file: %s \n", err)
 		}
 	}
+
+	// If a service account configuration file has been specified, check that it exists
+	if serviceAccountConfigFile != "" {
+		if _, err := os.Stat(serviceAccountConfigFile); os.IsNotExist(err) {
```
alexeykazakov: This code will fail if the file does exist because in this case err == nil
Should be something like this instead:
```
if serviceAccountConfigFile != "" {
    if _, err := os.Stat(serviceAccountConfigFile); err != nil {
        return nil, err // Return error for any reason if you can't get stat. If it doesn't exist or if it failed for another reason. 
    }
}
```

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+			// If it does not exist, return an error
+			return nil, errors.Errorf("Specified service account configuration file does not exist: %s \n", err)
+		}
+	}
+
+	// If the service account configuration file has not been specified
+	// then we default to /etc/fabric8/service-account-secrets.conf
+	if serviceAccountConfigFile == "" {
+		if _, err := os.Stat(varDefaultServiceAccountConfig); !os.IsNotExist(err) {
```
alexeykazakov: This code should fail if file exists as code above. Use something like this instead:
```
if serviceAccountConfigFile != "" {
    if _, err := os.Stat(serviceAccountConfigFile); err != nil {
        return nil, err
    }
} else {
     _, err := os.Stat(defaultServiceAccountConfigPath)
     if err == nil {
        serviceAccountConfigFile = defaultServiceAccountConfigPath
    } else if !os.IsNotExist(err) {
        return nil, err
    }
}
```

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+			serviceAccountConfigFile = varDefaultServiceAccountConfig
+		}
+	}
+
+	ex, err := os.Executable()
+	if err != nil {
+		panic(err)
```
alexeykazakov: Do not panic. Return err.

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+
+	// If the service account configuration file has not been specified
+	// then we default to /etc/fabric8/service-account-secrets.conf
+	if serviceAccountConfigFile == "" {
+		if _, err := os.Stat(varDefaultServiceAccountConfig); !os.IsNotExist(err) {
+			serviceAccountConfigFile = varDefaultServiceAccountConfig
+		}
+	}
+
+	ex, err := os.Executable()
```
alexeykazakov: From docs:

> Executable returns the path name for the executable that started the current process. There is no guarantee that the path is still pointing to the correct executable. If a symlink was used to start the process, depending on the operating system, the result might be the symlink or the path it pointed to. If a stable result is needed, path/filepath.EvalSymlinks might help.

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+	if err != nil {
+		panic(err)
+	}
+	exPath := filepath.Dir(ex)
+
+	// If we can't find it in the /etc/fabric8 directory, check the running directory
+	if serviceAccountConfigFile == "" {
+		checkPath := exPath + "/" + varServiceAccountConfigName
+		if _, err := os.Stat(checkPath); !os.IsNotExist(err) {
```
alexeykazakov: It will fail for the same reason. See above.

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+			serviceAccountConfigFile = checkPath
+		}
+	}
+
+	// Otherwise check the /conf subdirectory of the running directory
+	if serviceAccountConfigFile == "" {
+		checkPath := exPath + "/conf/" + varServiceAccountConfigName
+		if _, err := os.Stat(checkPath); !os.IsNotExist(err) {
```
alexeykazakov: It will fail for the same reason. See above.

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+			serviceAccountConfigFile = checkPath
+		}
+	}
+
+	// Otherwise as a last ditch attempt to locate the configuration, check the /conf subdirectory
+	// of the running directory's parent directory (../conf)
+	if serviceAccountConfigFile == "" {
+		checkPath := exPath + "/../conf/" + varServiceAccountConfigName
+		if _, err := os.Stat(checkPath); !os.IsNotExist(err) {
```
alexeykazakov: It will fail for the same reason. See above.

configuration/configuration.go
```diff
@@ -89,36 +90,114 @@ const (
...
+			serviceAccountConfigFile = checkPath
+		}
+	}
+
+	// If the service account configuration file has not been specified
+	if serviceAccountConfigFile == "" {
```
alexeykazakov: If the file still not found then we should return an error, shouldn't we? Otherwise load it:
```
if serviceAccountConfigFile == "" {
    return nil, errors.New(".....")
}
c.sa.SetConfigType("json")
c.sa.SetConfigFile(serviceAccountConfigFile)
```

alexeykazakov:
@sbryzak I wonder if we should actually store SA secrets as hashes in the service-account-secrets.conf file.

alexeykazakov:
@sbryzak I refactored the config loading. Instead of guessing the path to the default config file this PR now using data binding to generate a go file in "make". So, it doesn't matter what working dir you are using. The default config data will be always available. I also fixed a few issues with the original code.

alexeykazakov:
[test]

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-158-4`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-158-5`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-158-6`

alexeykazakov:
[test]

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-158-8`

vpavlin:
This looks ok from my point of view. I would like the configmap referenced in the PR/template to be added as well for the reference if someone wants to deploy this on their own. But feel free to merge this and open a new PR for that, I don't want to block this.

alexeykazakov:
@vpavlin we do have it referenced in the template - https://github.com/fabric8-services/fabric8-auth/pull/158/files#diff-8164f82f5c638cdf7da28b243478fe0eR26
Or you mean something else?

PR #164, opened 2017-11-03
Add cluster info to user data model

### LINKED ISSUE #165: data model changes for storing cluster information
part of https://github.com/fabric8-services/fabric8-auth/issues/162

sbose78:
In the first version of multi-cluster support, user will be enrolled to a single openshift cluster  ( which could be cluster A or cluster B or cluster C ).

Note: This design will not support when the same user would be enrolled to multiple clusters - with Cluster A , Cluster B and Custer C at the same time.

- [x] Refactor tests to test the return of cluster info in response.
- [x] Add the default cluster url to existing user entities.

Fixes #165


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-1`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-2`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-3`

design/account.go
```diff
@@ -170,6 +170,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("url", d.String, "The url")
 	a.Attribute("company", d.String, "The company")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
+	a.Attribute("cluster", d.String, "The cluster where the user has her OpenShift account is created")
```
alexeykazakov: Let's maybe more specific here what cluster means here. With examples. We will need to refer to the OpenShift API URL as a cluster I guess.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-4`

main.go
```diff
@@ -133,6 +133,12 @@ func main() {
 	if migrateDB {
 		os.Exit(0)
 	}
+	err = migration.UpdateUsersWithDefaultCluster(db, configuration.GetOpenShiftClientApiUrl())
```
alexeykazakov: It would be better to use a migration sql file for this but not sure how we could pass the default cluster URL...

design/account.go
sbose78: Updated this to be more descriptive: 7a9539d

main.go
sbose78: Yeah, that's why added this here - because the default url is cluster-specific. Also, this code could be taken out later on since every user would be seeded with the default cluster api url henceforth.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-5`

design/account.go
```diff
@@ -170,7 +170,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("url", d.String, "The url")
 	a.Attribute("company", d.String, "The company")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
-	a.Attribute("cluster", d.String, "The cluster where the user has her OpenShift account is created")
+	a.Attribute("cluster", d.String, "The OpenShift API URL of the cluster where the user's tenant services are being run")
```
alexeykazakov: Or rather "The OpenShift API URL of the cluster where the user was provisioned to". The tenant service is an implementation detail. The clients (the registration app) doesn't have to know such details.

main.go
alexeykazakov: This is a part of data migration. We should not remove this migration step. We should version this change in DB. We do have access to configuration in our migration code where we run our scripts. It looks like we support custom arguments in our sql templates.

sbose78: Thanks, trying out this
`m = append(m, steps{ExecuteSQLFile("010-add-cluster-to-user.sql", "my cluster")})`

design/account.go
```diff
@@ -170,6 +170,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("url", d.String, "The url")
 	a.Attribute("company", d.String, "The company")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
+	a.Attribute("cluster", d.String, "The OpenShift API URL of the cluster where the user's tenant services are being run")
```
alexeykazakov: "The OpenShift API URL of the cluster where the user is provisioned to"

migration/sql-files/010-add-cluster-to-user.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users ADD COLUMN cluster TEXT;
+UPDATE users SET cluster = '{{.}}'  WHERE cluster is null or cluster = '';
```
alexeykazakov: Let's add NOT NULL to the cluster field after updating the existing users.

account/user_blackbox_test.go
```diff
@@ -125,6 +127,7 @@ func createAndLoadUser(s *userBlackBoxTest) *account.User {
 		ImageURL: "someImageUrl" + uuid.NewV4().String(),
 		Bio:      "somebio" + uuid.NewV4().String(),
 		URL:      "someurl" + uuid.NewV4().String(),
+		Cluster:  "somecluster" + uuid.NewV4().String(),
```
alexeykazakov: Let's also add tests when we create or save a user without cluster then it fails (the cluster can't be null or empty)

sbose78: e13d57d71788b85bfa31808932b0554573218ffd

migration/sql-files/010-add-cluster-to-user.sql
sbose78: Thanks!
e13d57d71788b85bfa31808932b0554573218ffd

design/account.go
sbose78: Cool!
24929ff7e472058556a86f06d3db7c9ae886dc24

sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-9`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-10`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-11`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-12`

sbryzak:
The changes look good to me, however my one question would be whether we should be doing more to future-proof this now by at least building in support at the data schema level for multiple clusters per user?

migration/sql-files/010-add-cluster-to-user.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE users ADD COLUMN IF NOT EXISTS cluster TEXT;
+UPDATE users SET cluster = '{{ index . 0}}'  WHERE cluster is null or cluster = '';
+ALTER TABLE users ALTER COLUMN cluster set NOT NULL ,ADD CHECK (cluster <> '');
```
sbose78: Hi @sbryzak 
Do you seeing anything obvious which I should be taking care of , in my migration scripts to ensure that concurrent migration tests don't fail ? I see failure at the moment.. [1] 

Here are my database changes :  https://github.com/fabric8-services/fabric8-auth/pull/164/files#diff-7821b63d24bdd94b4c4cc05c4ff485c8

Here's the test that fails:
https://github.com/fabric8-services/fabric8-auth/blob/master/migration/migration_test.go

[1] https://ci.centos.org/job/devtools-fabric8-auth/362/console

sbose78:
> The changes look good to me, however my one question would be whether we should be doing more to future-proof this now by at least building in support at the data schema level for multiple clusters per user?

Good question!
At the moment - we'll only be supporting the assigning of users to a single cluster out of a pool of N clusters. That's what our multi-cluster implementation would be now ( at least for a while.. maybe a year ).

It will be quite a while before we have users registered to more than one cluster at the same time. I'm guessing we would need to store cluster info in a different table altogether then :)

alexeykazakov:
Yeah... we even don’t know yet how that multiple clusters per user support would look. If we try to guess and “support” it right now then we most likely will need to refactor it later anyway.
We do not need it right now, so, let’s not overengineer this module.

migration/sql-files/010-add-cluster-to-user.sql
sbryzak: I sent you some direct messages in Mattermost but it seems you're offline, anyway to repeat:

1. The DDL syntax looks to be only valid for postgresql 9.6 and higher: https://stackoverflow.com/questions/12597465/how-to-add-column-if-not-exists-on-postgresql

2. It looks like we are using Postgresql 9.5: 
(from the Jenkins log)
Pulling postgres_integration_test (centos/postgresql-95-centos7:latest)...

3. To fix the problem we could try updating to this postgres image instead:

https://hub.docker.com/r/centos/postgresql-96-centos7/

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-13`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-14`

sbose78:
[test]

sbose78:
[test]

migration/sql-files/010-add-cluster-to-user.sql
sbose78: @sbryzak Thanks a ton!
You were absolutely right. I removed the `IF NOT EXISTS` and it worked :)
since updating the version might not be an option.

Probably we should update the local postgres version to match it with ci.centos.

alexeykazakov:
[test]

alexeykazakov:
I added more tests https://github.com/fabric8-services/fabric8-auth/pull/164/commits/73734e29149bd9a3eab796b1cc90df44b5650751 to make sure we fails to set empty clusters because of the empty cluster field. Not because of other reasons.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-164-15`

alexeykazakov:
[test]

sbose78:
Thanks for the test, @alexeykazakov .

Issue #145, opened 2017-10-13
Switch to oauth2 flow for login


alexeykazakov:
We should switch to oauth2 flow for login in both UI/clients and auth backend.

- [x] Support oauth2 flow in Auth service - https://github.com/fabric8-services/fabric8-auth/issues/196
- [ ] Switch clients to ouath2 API

sbryzak:
Out of scope for auth service.

PR #169, opened 2017-11-04
Service Account token issuance (login) endpoint

### LINKED ISSUE #154: Service Account token issuance (login) endpoint
We need some REST API which can be used by services to "login" and obtain a service account token for service-to-service communications.

Part of https://github.com/fabric8-services/fabric8-auth/issues/72

alexeykazakov:
```
POST /api/token
    grant_type=client_credentials&
    client_id=SERVICE_ACCOUNT_ID&
    client_secret=SERVICE_ACCOUNT_SECRET
```
Response:
**200 OK**
```
{
    "access_token" : "SERVICE_ACCOUNT_TOKEN",
    "token_type" : "bearer"
}
```
**How to test:**
```
curl -d '{"grant_type": "client_credentials","client_id": "5dec5fdb-09e3-4453-b73f-5c828832b28e","client_secret": "witsecret"}' -H "Content-Type: application/json" -X POST http://localhost:8089/api/token
```
Fixes #154


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-3`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-4`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-5`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-169-6`

sbryzak:
Is this token also going to be used as a PAT?  If so, the scope should be set to uma_protection.

alexeykazakov:
> Is this token also going to be used as a PAT? If so, the scope should be set to uma_protection.

@sbryzak https://github.com/fabric8-services/fabric8-auth/pull/169/files#diff-6b1a7621961d1f6fe7463db59c5afef5R427

configuration/configuration_blackbox_test.go
```diff
@@ -341,11 +341,11 @@ func checkServiceAccountConfiguration(t *testing.T, accounts map[string]configur
 	checkServiceAccount(t, accounts, configuration.ServiceAccount{
 		ID:      "5dec5fdb-09e3-4453-b73f-5c828832b28e",
 		Name:    "fabric8-wit",
-		Secrets: []string{"witsecret"}})
+		Secrets: []string{"$2a$04$nI7z7Re4pbx.V5vwm14n5.velhB.nbMgxdZ0vSomWVxcct34zbH9e"}})
 	checkServiceAccount(t, accounts, configuration.ServiceAccount{
 		ID:      "c211f1bd-17a7-4f8c-9f80-0917d167889d",
 		Name:    "fabric8-tenant",
-		Secrets: []string{"tenantsecretOld", "tenantsecretNew"}})
+		Secrets: []string{"$2a$04$ynqM/syKMYowMIn5cyqHuevWnfzIQqtyY4m.61B02qltY5SOyGIOe", "$2a$04$sbC/AfW2c33hv8orGA.1D.LXa/.IY76VWhsfqxCVhrhFkDfL0/XGK"}})
```
sbryzak: Would it be useful to have these values defined in a const somewhere so we can re-use them? I imagine that integration tests for the permission API / resource registration will need to perform an authentication also.

controller/token.go
```diff
@@ -291,6 +292,48 @@ func (c *TokenController) Delete(ctx *app.DeleteTokenContext) error {
 	return ctx.OK([]byte{})
 }
 
+func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
```
sbryzak: The code here looks good, I have a question about the function name though - is "Exchange" significant in some way?  To me, it doesn't seem as intuitive as something like AuthenticateServiceAccount for example.

configuration/configuration_blackbox_test.go
alexeykazakov: The tests you are referring to will use the plain secrets. Not hashes. Right?
But these tests in configuration just check how we load the configuration file. We should not re-use the hashes itself outside of the configuration tests, shouldn't we?

sbryzak: Ah yes of course, if I had looked a little closer I would have noticed you were simply testing the configuration, not actually authenticating with these values. :)

controller/token.go
alexeykazakov: I used OAuth 2 term "Exchange" for token exchange action because I was thinking of re-using this action to support other "grant_type" too. Right now we support "client_credentials" only. But latter we should support "authorization_code" too. See https://github.com/fabric8-services/fabric8-auth/issues/145

sbryzak: No problem, perhaps this could be described in the function comment? E.g. for now we can have:

// This function provides token exchange, allowing clients to authenticate using an id and secret value
// May be expanded in the future to support other exchange types

alexeykazakov: Yeah.. I will add the comment. Good idea.

Issue #166, opened 2017-11-03
Add API for creating a user in AUTH 


sbose78:
Add API for creating a user in AUTH - which would be used by the Registration App to 'inform' Auth about the new provisioned user.

Part of https://github.com/fabric8-services/fabric8-auth/issues/162

Issue #167, opened 2017-11-03
Workflow for Auth service to add provisioned user in keycloak.


sbose78:
part of https://github.com/fabric8-services/fabric8-auth/issues/162

alexeykazakov:
Done.

PR #170, opened 2017-11-04
Support for creating users using service account tokens

sbose78:
fixes https://github.com/fabric8-services/fabric8-auth/issues/166 and #167 

- [x] Add API to accept user details - would be called from registration app.
- [x] Push user data to keycloak
- [x] link user to RHD
- [x] Create User in Auth DB
- [x] Create User in WIT DB
- [x] Tests

Appropriate changes need to be made in the registration app too.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-4`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-5`

controller/users.go
```diff
@@ -77,6 +78,130 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+			"err": err,
+		}, "failed to determine if account is a service account")
+		return jsonapi.JSONErrorResponse(ctx, goa.ErrUnauthorized(err))
+
+	}
+	if !isSvcAccount {
+		log.Error(ctx, map[string]interface{}{
+			"identity_id": ctx.ID,
+		}, "account used to call create api is not a service account")
+		return jsonapi.JSONErrorResponse(ctx, goa.ErrUnauthorized(errs.New("a non-service account tried to create a user.")))
```
alexeykazakov: If there is no token at all then Goa will generate 401. Which is correct.
But if the token doesn't have permissions to do this operation then it would be better to return 403 (Forbidden). And we don't have to be too specific in the error message we return to the client. Add details to the error log message instead.
Generally speaking 401 means we don't know who you are. And 403 means we know you but you don't have permissions. If I understand it correctly ;)

controller/users.go
```diff
@@ -77,6 +78,130 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	}
+
+	return c.createUserInDB(ctx)
+}
+
+func (c *UsersController) createUserInDB(ctx *app.CreateUserAsServiceAccountUsersContext) error {
+
+	userID, err := uuid.FromString(ctx.Payload.Data.Attributes.UserID)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, goa.ErrBadRequest(errs.New("invalid user id")))
```
alexeykazakov: Let's maybe finally add errors.NewBadRequest() function and start using it? Just for consistency.
But in this particular case I think we already have some error type which represents Bad Request in case invalid format...

controller/users.go
```diff
@@ -77,6 +78,130 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	}
+
+	id, err := uuid.FromString(ctx.ID)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, goa.ErrBadRequest(errs.New("incorrect identity id")))
```
alexeykazakov: As above.

controller/users.go
```diff
@@ -77,6 +78,130 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	}
+
+	returnResponse := application.Transactional(c.db, func(appl application.Application) error {
```
alexeykazakov: Convert the payload to user/identity objects outside of transaction (and verify if all required fields present in the payload). Then create them in the DB in a transaction.

design/account.go
```diff
@@ -109,6 +130,24 @@ var _ = a.Resource("users", func() {
 		a.Response(d.BadRequest, JSONAPIErrors)
 	})
 
+	a.Action("createUserAsServiceAccount", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.POST("/:id"),
```
alexeykazakov: What ID are you requesting here? It should be just `POST /users`, shouldn't it?

design/account.go
```diff
@@ -109,6 +130,24 @@ var _ = a.Resource("users", func() {
...
+		)
+		a.Description("create a user using a service account")
+		a.Payload(createUser)
+		a.Response(d.OK, func() {
+			a.Media(user)
+		})
+		a.Response(d.BadRequest, JSONAPIErrors)
+		a.Response(d.InternalServerError, JSONAPIErrors)
+		a.Response(d.NotFound, JSONAPIErrors)
```
alexeykazakov: NotFound?

design/account.go
```diff
@@ -109,6 +130,24 @@ var _ = a.Resource("users", func() {
...
+		a.Response(d.Unauthorized, JSONAPIErrors)
+		a.Response(d.Forbidden, JSONAPIErrors)
+		a.Response(d.Conflict, JSONAPIErrors)
```
alexeykazakov: Are we expecting any conflicts here?

design/account.go
```diff
@@ -109,6 +130,24 @@ var _ = a.Resource("users", func() {
 		a.Response(d.BadRequest, JSONAPIErrors)
 	})
 
+	a.Action("createUserAsServiceAccount", func() {
```
alexeykazakov: We don't have any REST API for creating users yet. Why do not call it Create (it will be still allowed to the service account only though). Later we will use our AuthZ service to determine who can do what.

design/account.go
```diff
@@ -206,3 +245,20 @@ var identityDataAttributes = a.Type("IdentityDataAttributes", func() {
 	a.Attribute("username", d.String, "The username")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
 })
+
+var createUserDataAttributes = a.Type("CreateIdentityDataAttributes", func() {
+	a.Attribute("userID", d.String, "The id of the corresponding User")
```
alexeykazakov: This PR should re-use https://github.com/fabric8-services/fabric8-auth/pull/164 and also add a cluster info to the payload/DB.

controller/users.go
```diff
@@ -77,6 +78,130 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	if err != nil {
+		return false, err
+	}
+	return (*tokenManager).IsServiceAccount(ctx), nil
+}
+
+// CreateUserAsServiceAccount creates a user when requested using a service account token
+func (c *UsersController) CreateUserAsServiceAccount(ctx *app.CreateUserAsServiceAccountUsersContext) error {
+
+	isSvcAccount, err := isServiceAccount(ctx)
```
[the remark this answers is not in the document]
sbose78: yeah, was wondering we need to check by some account name :)
2c2cd2e

controller/users.go
sbose78: Thanks!
7b462ed

I am usually undecided between
1) Forbidden or Unauthorized 
2) always say Unauthorized, that way we don't divulge information to the attacker. 

In case of (1)  "you are close, you got the right kind of token, but your identity is not as we expect it to be" :) 

but I understand if our services need to communicate with each other properly we need to be clear with the responses.

controller/users.go
sbose78: thanks.
now using `POST /users`  - therefore no BadRequest is being returned ever inside the controller.

controller/users.go
sbose78: using `POST /users`  - no BadRequest isn't being returned ever inside the controller.

controller/users.go
sbose78: Done.

design/account.go
sbose78: sounds right, thanks.
a076d63

design/account.go
sbose78: a076d63

Initially, I had designed it to accept the identity uuid and the user uuid ( in the payload ) but then I realized that the registration app shouldn't be burdened with the task of generating IDs. It should simply be passing on the user/identity data - and auth service should decide the IDs. 

Removed userID from payload and the identityID from the POST url as suggested you.

design/account.go
sbose78: fixed 210bfef

design/account.go
sbose78: nope, fixed 210bfef

auth/service_accounts.go
```diff
@@ -0,0 +1,4 @@
+package auth
+
+// OnlineRegistrationServiceAccount is the service account name of the online-registration app
+const OnlineRegistrationServiceAccount = "online-registration"
```
sbose78: if not here, we gotta put these somewhere in one place :)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-6`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-7`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-8`

controller/users.go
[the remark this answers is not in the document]
alexeykazakov: Well, we can update the secret on the OS side independently on this PR. It won't break anything.
I created a separate issue for that - https://github.com/fabric8-services/fabric8-auth/issues/171

openshift/auth.config.yaml
```diff
@@ -29,7 +29,7 @@ objects:
   type: Opaque
   data:
     service-account-secrets: >-
-      ewogICJhY2NvdW50cyI6IFsKICAgICAgewogICAgICAgICAgIm5hbWUiOiJmYWJyaWM4LXdpdCIsCiAgICAgICAgICAiaWQiOiI1ZGVjNWZkYi0wOWUzLTQ0NTMtYjczZi01YzgyODgzMmIyOGUiLAogICAgICAgICAgInNlY3JldHMiOlsid2l0c2VjcmV0Il0KICAgICAgfSwKICAgICAgewogICAgICAgICAgIm5hbWUiOiJmYWJyaWM4LXRlbmFudCIsCiAgICAgICAgICAiaWQiOiJjMjExZjFiZC0xN2E3LTRmOGMtOWY4MC0wOTE3ZDE2Nzg4OWQiLAogICAgICAgICAgInNlY3JldHMiOlsidGVuYW50c2VjcmV0T2xkIiwgInRlbmFudHNlY3JldE5ldyJdCiAgICAgIH0KICBdCn0=  
+      ew0KICAiYWNjb3VudHMiOiBbDQogICAgICB7DQogICAgICAgICAgIm5hbWUiOiJmYWJyaWM4LXdpdCIsDQogICAgICAgICAgImlkIjoiNWRlYzVmZGItMDllMy00NDUzLWI3M2YtNWM4Mjg4MzJiMjhlIiwNCiAgICAgICAgICAic2VjcmV0cyI6WyJ3aXRzZWNyZXQiXQ0KICAgICAgfSwNCiAgICAgIHsNCiAgICAgICAgICAibmFtZSI6ImZhYnJpYzgtdGVuYW50IiwNCiAgICAgICAgICAiaWQiOiJjMjExZjFiZC0xN2E3LTRmOGMtOWY4MC0wOTE3ZDE2Nzg4OWQiLA0KICAgICAgICAgICJzZWNyZXRzIjpbInRlbmFudHNlY3JldE9sZCIsICJ0ZW5hbnRzZWNyZXROZXciXQ0KICAgICAgfSwNCiAgICAgIHsNCiAgICAgICAgICAgICJuYW1lIjoib25saW5lLXJlZ2lzdHJhdGlvbiIsDQogICAgICAgICAgICAiaWQiOiJmODY3ZWM3Mi0zMTcxLTRiOGYtOGVlYy05MGEzMmVhYjZlMGIiLA0KICAgICAgICAgICAgInNlY3JldHMiOlsicmVnaXN0cmF0aW9uc2VjcmV0Il0NCiAgICAgIH0NCiAgXQ0KfQ==
```
alexeykazakov: We don't really need to update this every time we add a new service account. This particular secret is not used anywhere. It's just an example.

auth/service_accounts.go
alexeykazakov: Right now we just hardcode the name of the sa. Later we will associate a set of scopes for all service accounts. And instead of checking the SA name we will check the scopes via our standard authorization workflow.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-9`

openshift/auth.config.yaml
sbose78: Thanks - I was wondering we might need these for end-to-end testing?

auth/service_accounts.go
sbose78: Cool, removed - and hard-coded. 85ed444

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-10`

openshift/auth.config.yaml
alexeykazakov: openshift/auth.config.yaml is not used anywhere. It's just an example. Real secrets are set in our clusters.
So, for Auth tests (unit and integration) /configuration/config-files/service-account-secrets.conf is used. For end-to-end tests the secrets set in our preview will be used.

sbose78: Got it, 1947401788237363984b0b803839c29d74696441

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-11`

sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-12`

token/token.go
```diff
@@ -419,6 +419,20 @@ func (mgm *tokenManager) initServiceAccountToken(req *goa.RequestData) (string,
 	return mgm.serviceAccountToken, nil
 }
 
+func (mgm *tokenManager) IsServiceAccount(ctx context.Context, serviceAccountUser string) bool {
```
alexeykazakov: I've implemented this function in https://github.com/fabric8-services/fabric8-auth/blob/master/token/token.go#L438
You can re-use it now.

sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-14`

controller/users.go
```diff
@@ -84,7 +84,7 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
 	isSvcAccount := token.IsSpecificServiceAccount(ctx, "online-registration")
 	if !isSvcAccount {
 		log.Error(ctx, nil, "account used to call create api is not a service account")
-		return jsonapi.JSONErrorResponse(ctx, errors.NewForbiddenError("account not authorized to create users."))
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("account not authorized to create users."))
```
alexeykazakov: Any specific reason to use 401 vs 403 here?

sbose78:
```
func IsSpecificServiceAccount(ctx context.Context, name string) bool {
	accountName, ok := extractServiceAccountName(ctx)
	if !ok {
		return false
	}
	return accountName == name
}
```
There can be two conditions for the ` IsSpecificServiceAccount` to be false:
1. because it's a user token ( which is ~ 401 )
2. because it's a service account token but doesn't belong to `online-registration`  ( which is definitely 403 )

I wanted to hide out this detail externally - and do a blanket 401.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-15`

sbose78:
[test]

sbose78:
[test]

sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-16`

alexeykazakov:
BTW, as soon as start using this API to create users instead of using KC from the registration app. It would be interesting to see if the user has ever logged in. So, some "last time logged-in" field in our DB would be useful. We update it on every login. If it's nil then the user has never logged in. But it's for a separate PR/issue though.

controller/users_blackbox_test.go
```diff
@@ -984,3 +994,71 @@ func createDummyUserProfileResponse(updatedBio, updatedImageURL, updatedURL *str
...
+	}
+	secureService, secureController := s.SecuredServiceAccountController(testsupport.TestOnlineRegistrationAppIdentity)
+
+	// when
+	createUserPayload := createCreateUsersAsServiceAccountPayload(&user.Email, &user.FullName, &user.Bio, &user.ImageURL, &user.URL, &user.Company, &identity.Username, &identity.RegistrationCompleted, user.ContextInformation, user.Cluster)
+	_, appUser := test.CreateUsersOK(s.T(), secureService.Context, secureService, secureController, createUserPayload)
+	assertCreatedUser(s.T(), appUser.Data, user, identity)
+}
+
+func (s *TestUsersSuite) TestCreateUserAsServiceAccountUnauthorized() {
```
alexeykazakov: Good. You test if there is the expected SA or if there is a SA token but with unexpected service name.
But let's also test:
- if there is no token at all.
- if the token is not a service account token but a regular OSIO user token.

controller/users.go
```diff
@@ -77,6 +78,112 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+
+	return c.createUserInDB(ctx)
+}
+
+func (c *UsersController) createUserInDB(ctx *app.CreateUsersContext) error {
+
+	// can we specify what the identityID of a user can be , in keycloak?
+	// if not, then maybe we should first create the user in keycloak
+	// and then use that identityID here.
+	identityID := uuid.NewV4()
```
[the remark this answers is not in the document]
alexeykazakov: This actually the only major thing which blocks this PR. The rest of the comments are minor.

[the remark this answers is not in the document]
sbose78: Thanks.
Yeah I realized that a bit late. So effectively this task should include this task https://github.com/fabric8-services/fabric8-auth/issues/167 as well.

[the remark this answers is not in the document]
sbose78: So apart from persisting in the DB, we need to create in WIT also - have the code in place, adding tests & testing now.. :)

sbose78:
Tracking the "Whether the user has logged in.." : 
https://github.com/fabric8-services/fabric8-auth/issues/190

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-20`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-21`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-23`

controller/users.go
```diff
@@ -77,6 +83,195 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+}
+
+func (c *UsersController) createUserInKeycloak(ctx *app.CreateUsersContext) (*string, error) {
+
+	// All the below attributs are mandatory:
+	// "username", "email", "cluster", "company", "fullName", "approved", "rhd_username")
+
+	userAttributes := ctx.Payload.Data.Attributes
+
+	keycloakUser := login.KeytcloakUserRequest{
```
alexeykazakov: We also need to create a link to the user's RHD account in Keycloak. Otherwise the user won't able to login.
And let's also add the RHD username to User table.

sbose78:
Thanks for updating the branches ( and resolving the conflicts ) @alexeykazakov . 
I will now work on the review changes ( + RHD linking ).

controller/users.go
sbose78: umm. but the RHD username would be same as the identity username , right?  Then why are we gonna persist that separately ?

alexeykazakov: If we allow to change the username then it will be different than RHD name. We can address this in a separate issue/PR though.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-27`

controller/users_blackbox_test.go
sbose78: Thanks, added ab4347d

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-28`

sbose78:
Tried out the following payloads with `POST /api/users ` successfully - which includes
1. add user to keycloak
2. add user in db
3. link rhd in keycloak.
4. add user in WIT

```
{
  "data": {
    "attributes": {
      "approved": true,
      "enabled": true,
      "email": "sbosetest12345jhka@a.com",
      "rhd_username": "rhdusername",
      "fullName": "fullName B",
      "bio": "bio",
      "image_url": "imageURL",
      "url": "profileURL",
      "company": "company",
      "username": "sbosetest12345ausername",
      "registration_completed": true,
      "provider_type": "kc",
      "cluster": "cluster"
    },
    "type": "identities"
  }
}
```

and with mandatory attributes only.

```
{
  "data": {
    "attributes": {
      "approved": true,
      "email": "sbosetest123456jhka@a.com",
      "rhd_username": "rhdusername",
      "fullName": "fullName B",
      "company": "company",
      "username": "sbosetest123456ausername",
      "cluster" : "cluser"
    },
    "type": "identities"
  }
}
```
There are changes in the way the payload needs to be ( compared to what online-registration used to send as payload to keycloak ). This new format conforms to json-api as well as includes other optional fields which the user APIs in Auth has.

controller/users.go
```diff
@@ -77,6 +78,112 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	if contextInformation != nil {
+		if user.ContextInformation == nil {
+			user.ContextInformation = account.ContextInformation{}
+		}
+		for fieldName, fieldValue := range contextInformation {
+			user.ContextInformation[fieldName] = fieldValue
+		}
+	}
+
+	returnResponse := application.Transactional(c.db, func(appl application.Application) error {
```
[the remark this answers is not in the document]
sbose78: Thanks, done.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-29`

sbose78:
This PR is ready for review, though I'm still testing it - 

1. Could you please review https://github.com/fabric8-services/fabric8-auth/pull/170/files#diff-5a990e44c9a8391f2956c4b6d0128401R248  - wondering if the payload needs to be different. I had to find a good mix between
 the kind of payload our existing user endpoints have
+ the needs of the datamodel.
+ the info which gets passed from online-registration.

2. There are 4 major steps during the new user creation process. What's your suggestion on situations like 'keycloak user created but linking failed' . Were there any such scenarios possible in the existing online-registration workflow we have ? I'm trying to look for cases where persistence in the database might fail with any given specific payload ( will add tests if I do so ).

3. Should we add a different build task which runs remote tests ( to keycloak etc ) which tests workflows like we have here ? Asking this because this workflow is critical. I'm more concerned about situation like 'x out of y steps work, and then rollback of the x steps might not work because (maybe) keycloak is unstable :)

controller/users_blackbox_test.go
```diff
@@ -857,6 +860,17 @@ func findUser(id uuid.UUID, userData []*app.UserData) *app.UserData {
 	return nil
 }
 
+func assertCreatedUser(t *testing.T, actual *app.UserData, expectedUser account.User, expectedIdentity account.Identity) {
```
alexeykazakov: assert ContextInfo too

controller/users_blackbox_test.go
```diff
@@ -984,3 +1009,125 @@ func createDummyUserProfileResponse(updatedBio, updatedImageURL, updatedURL *str
...
+		"rate":         100.00,
+		"count":        3,
+	}
+	user.Company = "randomCompany"
+	secureService, secureController := s.SecuredServiceAccountController(testsupport.TestOnlineRegistrationAppIdentity)
+
+	// when
+	createUserPayload := createCreateUsersAsServiceAccountPayload(&user.Email, &user.FullName, &user.Bio, &user.ImageURL, &user.URL, &user.Company, &identity.Username, &identity.RegistrationCompleted, user.ContextInformation, user.Cluster)
+	_, appUser := test.CreateUsersOK(s.T(), secureService.Context, secureService, secureController, createUserPayload)
+	assertCreatedUser(s.T(), appUser.Data, user, identity)
```
alexeykazakov: Check if user is created correctly for a payload:
- with required fields only
- with all fields

Also check if user creation fails if some required info is missing/empty - username, email, etc.

login/link/keycloak_idp.go
```diff
@@ -0,0 +1,92 @@
...
+		log.Error(nil, map[string]interface{}{
+			"response_status":       resp.Status,
+			"response_body":         rest.ReadBody(resp.Body),
+			"keycloak_idp_link_url": keycloakIDPLinkURL,
+		}, "Unable to create idp link for RHD")
+
+		// Observed this error code when trying to create user
+		// with a token belonging to a different realm.
+		if resp.StatusCode == 403 {
+			return errors.NewUnauthorizedError(rest.ReadBody(resp.Body))
```
alexeykazakov: The body is already empty here because you just read it from the reader in the line 73. We should be careful with reading response body to avoid 1) file descriptor leaks 2) make sure we do not read from the reader multiple times.
A better pattern would be read the body right after you get the response (and if the error is nil) to a string and then use this string instead of the empty reader. Example - https://github.com/fabric8-services/fabric8-auth/pull/205/files#diff-b94dd2b70a7cf96de1555fd11e415f49R106

login/link/keycloak_idp.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return errors.NewInternalError(ctx, err)
+	}
+	req.Header.Add("Authorization", "Bearer "+protectedAccessToken)
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := c.client.Do(req)
+
+	if err != nil {
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

login/link/keycloak_idp.go
```diff
@@ -0,0 +1,92 @@
...
+			"err": err,
+		}, "Unable to create idp link for RHD")
+		return errors.NewInternalError(ctx, err)
+	} else if resp != nil {
+		defer resp.Body.Close()
+	}
+
+	if resp.StatusCode != 204 {
+
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+	if err != nil {
+		return nil, errors.NewInternalError(ctx, err)
+	}
+	req.Header.Add("Authorization", "Bearer "+protectedAccessToken)
+	req.Header.Add("Content-Type", "application/json")
+
+	resp, err := userProfileClient.client.Do(req)
+
+	if err != nil {
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+			"err": err,
+		}, "Unable to create Keycloak user")
+		return nil, errors.NewInternalError(ctx, err)
+	} else if resp != nil {
+		defer resp.Body.Close()
+	}
+
+	if resp.StatusCode != 201 {
+
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+			"response_status":           resp.Status,
+			"response_body":             rest.ReadBody(resp.Body),
+			"keycloak_user_profile_url": keycloakAdminUserAPIURL,
+		}, "Unable to create Keycloak user")
+
+		// Observed this error code when trying to create user
+		// with a token belonging to a different realm.
+		if resp.StatusCode == 403 {
+			return nil, errors.NewUnauthorizedError(rest.ReadBody(resp.Body))
```
alexeykazakov: The same as above (read the body once).

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+	createdUserURL, err := resp.Location()
+	if err != nil {
+		log.Error(nil, map[string]interface{}{
+			"keycloak_user_url": keycloakAdminUserAPIURL,
+			"err":               err,
+		}, "Unable to create Keycloak user")
+		return nil, errors.NewInternalError(ctx, err)
+	}
+	if createdUserURL == nil {
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+			"keycloak_user_url": keycloakAdminUserAPIURL,
+		}, "Unable to create Keycloak user")
+		return nil, errors.NewInternalError(ctx, errs.Errorf("user creation in keycloak might have failed."))
+	}
+
+	createdUserURLString := createdUserURL.String()
+	log.Info(nil, map[string]interface{}{
```
alexeykazakov: log.Info(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+	}
+	log.Info(nil, map[string]interface{}{
+		"response_status":           resp.Status,
+		"response_body":             rest.ReadBody(resp.Body),
+		"keycloak_user_profile_url": keycloakAdminUserAPIURL,
+	}, "Successfully create Keycloak user")
+
+	createdUserURL, err := resp.Location()
+	if err != nil {
+		log.Error(nil, map[string]interface{}{
```
alexeykazakov: log.Error(ctx, ...)

controller/users.go
```diff
@@ -77,6 +87,227 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	if !isSvcAccount {
+		log.Error(ctx, nil, "The account is not an authorized service account allowed to create a new user")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("account not authorized to create users."))
+	}
+
+	tokenEndpoint, err := c.config.GetKeycloakEndpointToken(ctx.RequestData)
+	if err != nil {
+		return errors.NewInternalError(ctx, err)
+	}
+	log.Info(nil, map[string]interface{}{
```
alexeykazakov: log.Info(ctx, ...)

login/link/keycloak_idp.go
```diff
@@ -0,0 +1,92 @@
...
+
+		// Observed this error code when trying to create user
+		// with a token belonging to a different realm.
+		if resp.StatusCode == 403 {
+			return errors.NewUnauthorizedError(rest.ReadBody(resp.Body))
+		}
+
+		return errors.NewInternalError(ctx, errs.Errorf("received a non-200 response %s while creating keycloak user :  %s", resp.Status, keycloakIDPLinkURL))
+	}
+	log.Info(nil, map[string]interface{}{
```
alexeykazakov: log.Info(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+		// Observed this error code when trying to create user with an existing username.
+		if resp.StatusCode == 409 {
+			// This isn't version conflict really,
+			// but helps us generate the final response code.
+			return nil, errors.NewVersionConflictError(fmt.Sprintf("user with username %s / email %s already exists", keycloakUserRequest.Username, keycloakUserRequest.Email))
+		}
+
+		return nil, errors.NewInternalError(ctx, errs.Errorf("received a non-200 response %s while creating keycloak user :  %s", resp.Status, keycloakAdminUserAPIURL))
+	}
+	log.Info(nil, map[string]interface{}{
```
alexeykazakov: log.Info(ctx, ...)

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+		log.Error(nil, map[string]interface{}{
+			"keycloak_user_url": keycloakAdminUserAPIURL,
+		}, "Unable to create Keycloak user")
+		return nil, errors.NewInternalError(ctx, errs.Errorf("user creation in keycloak might have failed."))
+	}
+
+	createdUserURLString := createdUserURL.String()
+	log.Info(nil, map[string]interface{}{
+		"keycloak_user_url": keycloakAdminUserAPIURL,
+		"err":               err,
```
alexeykazakov: Remove "err"

login/profile.go
```diff
@@ -77,6 +95,86 @@ func NewKeycloakUserProfileClient() *KeycloakUserProfileClient {
...
+		"user_url":          createdUserURLString,
+	}, "Unable to create Keycloak user")
```
alexeykazakov: I guess you wanted to report a successful creation. Not an error.

login/profile_user_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+	testsuite "github.com/fabric8-services/fabric8-auth/test/suite"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/fabric8-services/fabric8-auth/auth"
+	keycloakLink "github.com/fabric8-services/fabric8-auth/login/link"
+
+	"github.com/goadesign/goa"
+	_ "github.com/lib/pq"
```
alexeykazakov: ?

login/profile_user_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+// SetupSuite overrides the RemoteTestSuite's function but calls it before doing anything else
+// The SetupSuite method will run before the tests in the suite are run.
+func (s *ProfileUserBlackBoxTest) SetupSuite() {
+	resource.Require(s.T(), resource.Remote)
+	var err error
+	s.Config, err = configuration.GetConfigurationData()
+	if err != nil {
+		panic(fmt.Errorf("Failed to setup the configuration: %s", err.Error()))
+	}
```
alexeykazakov: Replace lines 43-48 by s.RemoteTestSuite.SetupSuite()

login/profile_user_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+	r := &goa.RequestData{
+		Request: &http.Request{Host: "api.example.org"},
+	}
+
+	s.tokenEndpoint, err = s.Config.GetKeycloakEndpointToken(r)
+
+	// http://sso.prod-preview.openshift.io/auth/admin/realms/fabric8/users"
+	s.userAPIFOrAdminURL, err = s.Config.GetKeycloakEndpointUsers(r)
+
+	token, err := s.generateProtectedAccessToken()
```
alexeykazakov: While you are skipping the test above the SetupSuite is not skipped and you call KC to generate a PAT. Let's move this code into the test method so we can skill all interactions with KC in the test suite.

login/profile_user_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+	assert.Nil(s.T(), err)
+	require.NotNil(s.T(), token)
+	s.protectedAccessToken = *token
+}
+
+func (s *ProfileUserBlackBoxTest) TearDownTest() {
```
alexeykazakov: Just remove this method.

login/profile_user_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+	clientSecret := s.Config.GetKeycloakSecret()
+	token, err := auth.GetProtectedAPIToken(context.Background(), s.tokenEndpoint, clientID, clientSecret)
+	require.Nil(s.T(), err)
+	require.NotNil(s.T(), token)
+
+	return &token, err
+}
+
+func (s *ProfileUserBlackBoxTest) TestPATGenerated() {
+	assert.NotEmpty(s.T(), s.protectedAccessToken)
```
alexeykazakov: Skip it.

controller/users.go
```diff
@@ -77,6 +87,227 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	// finally, if all works, we create a user in WIT too.
+	witURL, err := c.config.GetWITURL(ctx.RequestData)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "failed to create user in WIT")
+
+		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+	}
+	identity.User = *user // being explicit
```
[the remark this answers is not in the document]
sbose78: not sure I understand, why inside `createUserInDB()` ? we call WIT after this..

controller/users.go
```diff
@@ -77,6 +87,227 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+	identity, user, err := c.createUserInDB(ctx, identityID)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "failed to create user in DB")
+
+		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+	}
+
+	err = c.linkUserToRHD(ctx, *identity, protectedAccessToken)
```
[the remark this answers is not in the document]
sbose78: was ensuring we first sync kc <--> auth-db first before anything else :)

controller/users.go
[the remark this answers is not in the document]
alexeykazakov: we create identity and user inside createUserInDB() so why not setting them up fully in that function? why you don't do identity.User = *user inside createUserInDB()?

controller/users.go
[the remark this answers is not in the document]
alexeykazakov: If we do all the KC part first then Auth/WIT part then if anything bad happens in KC we will have a broken stuff in KC only (easy to clean up manually). Otherwise we may have broken stuff everywhere.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-30`

controller/users.go
```diff
@@ -77,6 +87,227 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+
+	userURL, err := c.userProfileService.Create(ctx, &keycloakUser, protectedAccessToken, usersEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Handle error, check if there was actually a URL returned.
+	userURLComponents := strings.Split(*userURL, "/")
+
+	identityID := userURLComponents[len(userURLComponents)-1]
```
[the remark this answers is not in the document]
alexeykazakov: NM. it's always > 0

controller/users.go
```diff
@@ -77,6 +87,227 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
...
+		ID:       userID,
+		Email:    ctx.Payload.Data.Attributes.Email,
+		Cluster:  ctx.Payload.Data.Attributes.Cluster,
+		Company:  ctx.Payload.Data.Attributes.Company,
+		FullName: ctx.Payload.Data.Attributes.FullName,
+	}
+	identity = &account.Identity{
+		ID:           identityID,
+		Username:     ctx.Payload.Data.Attributes.Username,
+		ProviderType: "kc", // replace with a string constant
```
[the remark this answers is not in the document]
alexeykazakov: It's actually "kc" but anyway worth using the constant.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-31`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-170-32`

sbose78:
Thanks a lot @alexeykazakov for contributing the changes to this PR - I will manually test it before I merge this.

sbose78:
Tested out the new minimal set of attributes

```
 {
  "data": {
    "attributes": {
      "email": "sbose2test123456jhka@a.com",
      "username": "sbosetest02123456ausername",
      "cluster" : "cluser"
    },
    "type": "identities"
  }
}
```

Works fine , verified the link, WIT data and also whether the user is approved=true

sbose78:
( and I take responsibility of cleaning up keycloak as well ;) )

PR #175, opened 2017-11-07
ISSUE-163 Set up documentation build with asciidoc

sbryzak:
This is just the initial setup for building documentation.  It's a little hackish as each documentation source file is specified separately in the Makefile (asciidoc doesn't seem to support processing docs for an entire directory) so this could be improved with some kind of file iteration that runs the tool for each *.adoc source file found.  

Anyway this is enough to get us up and running - I've configured Github Pages for our project so any changes to the /docs directory gets automatically published [here](https://fabric8-services.github.io/fabric8-auth).


sbryzak:
Actually, the one-file limitation was only in the original asciidoc tool, it seems that asciidoctor (which builds on top of asciidoc) is capable of processing an entire source directory.  I've updated the PR with the changes.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-175-1`

sbryzak:
Some more useful information - to build the docs, simply run:

`make docs`

in the root project directory.  This will generate the docs from source (/docs/source) and replace the existing html files in /docs.

PR #181, opened 2017-11-09
Clean up test identity in account linking tests

alexeykazakov:
Fixes the issue with leftovers in identities table after account linking tests.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-181-1`

Issue #186, opened 2017-11-13
It's not possible to set empty string in User's profile for optional fields


alexeykazakov:
Currently we use db.Model(&user).Updates(model) to save users. But gorm doesn't update fields with blank values (it treats them as not-changed). We should use db.Save(&user) instead. Otherwise we can't really set empty values to User's optional fields like BIO or URL.

See http://jinzhu.me/gorm/crud.html#update for details.

PR #189, opened 2017-11-15
Enabling updates to accept empty strings for User's profile for optional fields

kishansagathiya:
Made changes such that gorm can update fields with blank values as well
It was treating them as not-changed before

Fixes(#186)

### DIFF
--- a/account/identity.go
+++ b/account/identity.go
@@ -211,7 +211,7 @@ func (m *GormIdentityRepository) Save(ctx context.Context, model *Identity) erro
 		}, "unable to update the identity")
 		return errs.WithStack(err)
 	}
-	err = m.db.Model(obj).Updates(model).Error
+	err = m.db.Save(obj).Error
 
 	log.Debug(ctx, map[string]interface{}{
 		"identity_id": model.ID,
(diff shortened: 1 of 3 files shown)


fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-189-1`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-189-2`

kishansagathiya:
[test]

kishansagathiya:
[test]

account/identity.go
```diff
@@ -202,7 +202,7 @@ func (m *GormIdentityRepository) Lookup(ctx context.Context, username, profileUR
 func (m *GormIdentityRepository) Save(ctx context.Context, model *Identity) error {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "save"}, time.Now())
 
-	obj, err := m.Load(ctx, model.ID)
+	_, err := m.Load(ctx, model.ID)
```
alexeykazakov: Do we really need to load it with the new approach? What happen if you do m.db.Save(model) for unknown model? It should generate the same not-found error I guess?

kishansagathiya: Why should we be concerned with the not-found error? to run update we need user token, and if we have the user token that means the account exist, hence found.

kishansagathiya: Should be safe to remove that error check

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-189-3`

PR #191, opened 2017-11-16
Retrieve external token from Auth DB if Keycloak failed 

alexeykazakov:
We should try to load the token from Auth DB if Keycloak fails for any reason. Not just 401.

Fixes https://github.com/openshiftio/openshift.io/issues/1379

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-191-2`

PR #155, opened 2017-10-27
Check local DB first when retrieving external tokens

alexeykazakov:
When retrieving a GitHub or OS token check the local DB first. If there is nothing then check Keycloak. If Keycloak has a token then save it in the local DB and return.

Also improved Delete token API. Now it deletes all the tokens from the local DB for the given provider ID and identity. It should not normally happen (multiple tokens) but technically it's possible if there are a few linking requests in the same time.
We also now remove the link from Keycloak when deleting tokens.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/156

PR #192, opened 2017-11-16
Delete account links from Keycloak

alexeykazakov:
We should delete the Account links from Keycloak when deleting them via Auth API.

Extracted from https://github.com/fabric8-services/fabric8-auth/pull/155

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -208,6 +208,11 @@ func (c *TokenController) getKeycloakExternalTokenURL(providerName string) strin
 	return fmt.Sprintf("%s/auth/realms/%s/broker/%s/token", c.Configuration.GetKeycloakURL(), c.Configuration.GetKeycloakRealm(), providerName)
 }
 
+func (c *TokenController) getKeycloakIdentityProviderURL(identityID string, providerName string) string {
+	// not moving this to config because this is temporary.
+	return fmt.Sprintf("%s/auth/admin/realms/%s/users/%s/federated-identity/%s", c.Configuration.GetKeycloakURL(), c.Configuration.GetKeycloakRealm(), identityID, providerName)
+}
+
 // Retrieve fetches the stored external provider token.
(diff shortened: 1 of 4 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-192-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-192-3`

controller/token.go
```diff
@@ -280,6 +284,18 @@ func (c *TokenController) Delete(ctx *app.DeleteTokenContext) error {
...
 
+	// Delete from Keycloak
+	err = c.keycloakExternalTokenService.Delete(ctx, c.getKeycloakIdentityProviderURL(currentIdentity.String(), providerConfig.TypeName()))
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"error":         err,
+			"provider_name": providerConfig.TypeName(),
+			"identity_id":   currentIdentity,
+		}, "Unable to remove Identity Provider link from Keycloak.")
+		// Not critical. Log the error and proceed.
```
sbose78: :+1:  When this happens, users are gonna see that their account is still connected as long as we check keycloak first.

controller/token.go
```diff
@@ -290,7 +306,12 @@ func (c *TokenController) Delete(ctx *app.DeleteTokenContext) error {
 			return err
 		}
 		if len(tokens) > 0 {
-			return appl.ExternalTokens().Delete(ctx, tokens[0].ID)
+			for _, token := range tokens {
```
sbose78: This situation shouldn't arise in a ideal world , right?

alexeykazakov: That’s right. But theoretically it may happen if there are many concurrent link requests from the same users (or even GET token requests until we stop get token from KC first and then save them in DB).

Issue #201, opened 2017-11-21
Improve readiness prob/status to make sure all critical CM/Secrets are set


alexeykazakov:
Currently we are using /api/status endpoint to check if the service is ready and alive but we only check DB connection there.
We also should expend it and also check if all the critical configuration like env vars and config files are set correctly and we don't use the default values in production.
It's probably make sense to split the Readiness and Liveness prob endpoints instead of using /status for everything.
So, if some critical config maps or secrets are not set in prod/preview and only the default one are available  then the Auth service should not pass the readness prob.

PR #206, opened 2017-11-23
Improve readiness prob/status

alexeykazakov:
Improves readiness prob/status to make sure all critical CM/Secrets are set

**New status format**

In case of DB and Configuration errors (**Status 503**):
```
{
  "buildTime": "2017-11-23T19:49:25Z",
  "commit": "03faa353d315f73668eea52260b2aafe52b25a45-dirty",
  "configurationStatus": "Error: /etc/fabric8/service-account-secrets.conf is not used; /etc/fabric8/oso-clusters.conf is not used; default service account private key is used; default service account private key ID is used; default DB password is used; default Keycloak client secret is used; default GitHub client secret is used",
  "databaseStatus": "Error: dial tcp 127.0.0.1:5433: getsockopt: connection refused",
  "startTime": "2017-11-23T19:52:03Z"
}
```
When there is no error (**Status 200**):
```
{
  "buildTime": "2017-11-23T19:49:25Z",
  "commit": "03faa353d315f73668eea52260b2aafe52b25a45-dirty",
  "configurationStatus": "OK",
  "databaseStatus": "OK",
  "startTime": "2017-11-23T19:52:03Z"
}
```

In Dev Mode the configuration error is added by the status is 200 (not 503). In Dev Mode we also now have `"devMode" : true` added to the status JSON.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/201


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-206-1`

sbose78:
Do we know why the latest issues with deployment happened ? 
This readiness effort looks really good.

controller/status.go
```diff
@@ -17,33 +18,77 @@ var (
...
 // Show runs the show action.
 func (c *StatusController) Show(ctx *app.ShowStatusContext) error {
-	res := &app.Status{}
-	res.Commit = Commit
-	res.BuildTime = BuildTime
-	res.StartTime = StartTime
+	res := &app.Status{
+		Commit:    Commit,
+		BuildTime: BuildTime,
+		StartTime: StartTime,
```
sbose78: ( Now that we are refactoring ) 
Since we are doing the db test + configuration tests, should we report that in the status response too ?

In fact, instead of just returning the error in the response ( when there's an error ), could we stick to this format and specify the error in the json with the appropriate http response code.
example,

```
{
  "commit":"xyz123",
  "build_time":"xyz123",
  "start_time":"xyz123",
   "database_status":"Error connecting to host",
   "configuration_status":"Error: default configuration is being read."
}
```

This will help us confirm which is the build/commit which is breaking.
What do you think?

alexeykazakov:
No we don’t know what was the issue. It just suddenly started to work after N deployment. But it’s pretty dangerous to use some default config values in prod which should be used only in dev env. Plus it gives us some safety net for cases when secrets are changed unexpectedly.

controller/status.go
alexeykazakov: We do return the whole json with the commit hash, build time, start time and error message in case of errors. But we return it as 503 instead of 200 if there is no error.

But we could refactor our error key in that JSON and use db_status + config_status instead. Let’s do it.

alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/206/commits/e9a711bb60558fe6b9383d268ce559da7c23f945

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-206-2`

controller/status.go
alexeykazakov: I also updated the description of the PR above. Added examples of the new status output.

PR #208, opened 2017-11-24
Allow to overwrite valid redirect URL in prod and staging env

alexeykazakov:
Currently the deployment fails in preview because we have the valid redirect URL there is different from the default one and prod.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-208-1`

Issue #196, opened 2017-11-21
Support Oauth2 flow for user login


alexeykazakov:
This is about supporting `grant_type=authorization_code` for users login flow.

For user login, here are the steps:
1. Client sends a generated state. Auth "proxies" it to KC and thereby initiates the first in oauth2
2. KC redirects to IdP etc. User logs in..
3. KC `calls back` Auth with a `code`
4. Auth passes the `code` to Client ( UI or another services )
5. Client then calls KC to do the `code` <-> `token` exchange.
Auth works as a smart proxy here between client and KC.

Part of https://github.com/fabric8-services/fabric8-auth/issues/145

kishansagathiya:
@sbose78 @alexeykazakov 
I am thinking a flow as follows
- To get authorization_code. We need to create /api/login/authorize endpoint
``` 
GET /api/login/authorize
        response_type=code&
        client_id=SERVICE_ACCOUNT_ID&
        redirect_uri=CALLBACK_URL&
        scope=read
```

- To get the access token
```
POST /api/token
    grant_type=authorization_code&
    client_id=SERVICE_ACCOUNT_ID&
    client_secret=SERVICE_ACCOUNT_SECRET&
    code=AUTHORIZATION_CODE&
    redirect_uri=CALLBACK_URL
```

alexeykazakov:
Looks good. But I think we can support public clients only for now. It doesn’t make much sense to use client secrets in UI.

sbose78:
There should be approx two APIs on the Auth side to make Auth act like a soft proxy : 

> Client sends a generated state. Auth "proxies" it to KC and thereby initiates the first in oauth2
This should be API 1 .
Auth needs to accept the `state` here from the UI ( or any other client ).

> KC redirects to IdP etc. User logs in..
> KC calls back Auth with a code
this should be API 2, which will be called (back) by KC.

> Auth passes the code to Client ( UI or another services )
> Client then calls KC to do the code <-> token exchange.
> Auth works as a smart proxy here between client and KC.

Are we on the same page @alexeykazakov , @kishansagathiya  ? :)

alexeykazakov:
If by API 1/2 you mean endpoint 1/2 then yes.

sbose78:
Right, endpoint. Thanks.

kishansagathiya:
@sbose78 @alexeykazakov 
[...]

PR #210, opened 2017-11-26
Endpoints for supporting oauth2.0 friendly login

kishansagathiya:
- [x] Create /api/authorize endpoint to authorize service client to login through fabric8-auth
```
GET /api/authorize
        response_type=code&
        client_id=SERVICE_ACCOUNT_ID&
        redirect_uri=CALLBACK_URL&
        scope=SCOPE&
	api_client=API_CLIENT&
	state=STATE_UUID
```
Response:
	code, state

- [x] Modify /api/token endpoint to enable grant_type=authorization_code
```
POST /api/token
    grant_type=authorization_code&
    client_id=SERVICE_ACCOUNT_ID&
    code=AUTHORIZATION_CODE&
    redirect_uri=CALLBACK_URL
```
- [x] Add tests
(#196)


fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-2`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-3`

kishansagathiya:
@alexeykazakov @sbose78 
- Does not support api_client so far
- Token is returned as (AccessToken, Tokentype) to be consistent with implementation of grant_type=client_credentials, what about using (AccessToken, expiry, RefreshToken, TokenType) for both?

login/service.go
```diff
@@ -337,6 +339,180 @@ func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, confi
...
+	}
+
+	// store referrer in a state reference to redirect later
+
+	log.Debug(ctx, map[string]interface{}{
+		"referrer": referrer,
+		"redirect": redirect,
+	}, "Got Request from!")
+
+	stateID := *ctx.State //uuid.NewV4()
```
sbose78: Check for ctx.State != nil ?

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-4`

controller/authorize.go
```diff
@@ -0,0 +1,85 @@
...
+		if ctx.ClientID == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_id", "nil").Expected("Service Account ID"))
+		}
+		if ctx.RedirectURI == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("redirect_uri", "nil").Expected("Redirect URI"))
+		}
+		if ctx.ResponseType == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("response_type", "nil").Expected("Response Type"))
+		}
+		if ctx.Scope == nil {
```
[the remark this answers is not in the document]
kishansagathiya: Added in new commit

sbose78:
Used the following to test:

Step 1:

In the browser:
```
http://localhost:8089/api/authorize?response_type=code&client_id=341c283f-0cd7-48a8-9281-4583aceb3617&redirect_uri=http://localhost:8089/api/status&scope=user:email&state=84f31f02-b0c7-4111-8c3b-a924cba9d8b8
```

gives a response:
```
{"code":"uss.8H8TUjZI6LT9bz0ySxgfpp5EWLnf-Sr6FILu8M-Ipj0.4731bb8c-14ee-451b-b82f-42423cda9185.239ed057-eec1-425b-a7eb-f4b338c94cdd","state":"84f31f02-b0c7-4111-8c3b-a924cba9d8b6"}
```

Step 2 : 

```
http://localhost:8089/api/token?grant_type=authorization_code&client_id=341c283f-0cd7-48a8-9281-4583aceb3617&code=uss.PhlzSqmVhCk5IDqFecMBh22Yq7cc_8XGIb05BcBMsdQ.c1b472b3-19eb-4b94-abce-a9b8da053183.239ed057-eec1-425b-a7eb-f4b338c94cdd&redirect_uri=http://localhost:8089/api/status
```

controller/authorize.go
```diff
@@ -0,0 +1,89 @@
...
+			}, "Unknown Service Account ID")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("invalid Service Account ID"))
+		}
+
+	}
+	authEndpoint, err := c.Configuration.GetKeycloakEndpointAuth(ctx.RequestData)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "Unable to get Keycloak Auth endpoint URL")
```
sbose78: error messages in small letters ( we try to follow this convention )

controller/authorize.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if ctx.Scope != nil {
+		authEndpoint = fmt.Sprintf("%s?scope=%s", authEndpoint, *ctx.Scope)
+	}
+
+	oauth := &oauth2.Config{
+		ClientID:     c.Configuration.GetKeycloakClientID(),
+		ClientSecret: c.Configuration.GetKeycloakSecret(),
+		Scopes:       []string{"user:email"},
+		Endpoint:     oauth2.Endpoint{AuthURL: authEndpoint, TokenURL: tokenEndpoint},
```
sbose78: do we need the scope in both places ( in `authEndpoint` and `Scopes`) ? Also, handle the situation where `scope` wasn't passed.

controller/token.go
```diff
@@ -342,41 +343,104 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
 	if payload == nil {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("payload", "nil").Expected("not empty payload"))
 	}
-	if payload.ClientID == nil {
-		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_id", "nil").Expected("Service Account ID"))
-	}
-	if payload.ClientSecret == nil {
-		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_secret", "nil").Expected("Service Account secret"))
-	}
+	if payload.GrantType == "client_credentials" {
```
sbose78: add this string as a constant at the top of this file?

controller/token.go
```diff
@@ -342,41 +343,104 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
...
+		if payload.ClientID == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_id", "nil").Expected("Service Account ID"))
+		}
+		if payload.ClientSecret == nil {
```
sbose78: we could make this optional for now, for `grant_type=authorization_code`

kishansagathiya: We need it here, for `grant_type=client_credentials`
it is optional for `grant_type=authorization_code`

controller/token.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/391604b2affa2b852131a55a5f94a7d1b120c10c

controller/authorize.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/391604b2affa2b852131a55a5f94a7d1b120c10c

controller/authorize.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/391604b2affa2b852131a55a5f94a7d1b120c10c

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-6`

controller/authorize.go
```diff
@@ -0,0 +1,93 @@
...
+
+	tokenEndpoint, err := c.Configuration.GetKeycloakEndpointToken(ctx.RequestData)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "unable to get keycloak token endpoint url")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, errs.Wrap(err, "unable to get keycloak token endpoint url")))
+	}
+	if ctx.Scope != nil {
+		authEndpoint = fmt.Sprintf("%s?scope=%s", authEndpoint, *ctx.Scope)
```
alexeykazakov: You are adding the scope via oauth2.Config object. No need to duplicate it here.

login/service.go
```diff
@@ -337,6 +339,166 @@ func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, confi
...
+		log.Debug(ctx, map[string]interface{}{
+			"code":  code,
+			"state": state,
+		}, "redirected from oauth provider")
+
+		authCode := &app.AuthorizationCode{
+			Code:  *ctx.Code,
+			State: *ctx.State,
+		}
+		return ctx.OK(authCode)
```
alexeykazakov: We should redirect to the original redirect URL and pass the code there instead of returning 200.

controller/authorize.go
```diff
@@ -0,0 +1,93 @@
...
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("response_type", "nil").Expected("response type"))
+		}
+
+		if ctx.Scope == nil {
+			scope = []string{"user:email"}
+		} else {
+			scope = []string{*ctx.Scope}
+		}
+
+		_, found := c.Configuration.GetServiceAccounts()[*ctx.ClientID]
```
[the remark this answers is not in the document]
kishansagathiya: Did you mean https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/conf-files/service-account-secrets.conf this file?

login/service.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/6597103ce44a1b789b4de9f9c257a3ae1532b596

controller/authorize.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/6597103ce44a1b789b4de9f9c257a3ae1532b596

controller/authorize.go
```diff
@@ -0,0 +1,93 @@
...
+// Authorize runs the authorize action.
+func (c *AuthorizeController) Authorize(ctx *app.AuthorizeAuthorizeContext) error {
+
+	if ctx.State == nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("state", "nil").Expected("state"))
+	}
+	var scope []string
+
+	if ctx.Code == nil {
+		if ctx.ClientID == nil {
```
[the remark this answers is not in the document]
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/6597103ce44a1b789b4de9f9c257a3ae1532b596

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-8`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-9`

controller/authorize.go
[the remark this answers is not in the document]
kishansagathiya: @alexeykazakov @sbose78 
since you want types, I think It would need everything to be refactored/renamed, sort of like
service-account-secrets.conf --> account-secrets.conf
type ServiceAccount --> type Account
and so on. 
I did a quick search of occurences of "serviceAccount". Quite a bit of renaming. I just want to make sure that I am going the right way.  ...?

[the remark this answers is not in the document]
alexeykazakov: We can do renaming in a separate PR.

controller/authorize.go
```diff
@@ -0,0 +1,85 @@
...
+	oauth := &oauth2.Config{
+		ClientID:     c.Configuration.GetKeycloakClientID(),
+		ClientSecret: c.Configuration.GetKeycloakSecret(),
+		Scopes:       scope,
+		Endpoint:     oauth2.Endpoint{AuthURL: authEndpoint, TokenURL: tokenEndpoint},
+		RedirectURL:  rest.AbsoluteURL(ctx.RequestData, "/api/authorize"),
+	}
+
+	ctx.ResponseData.Header().Set("Cache-Control", "no-cache")
+	return c.Auth.PerformAuthorize(ctx, oauth, c.Configuration)
```
alexeykazakov: Let's use the same approach as we do in token/link/ling.go instead of what we do in our login service.
Let's have to methods there: 1. Get redirect URL for authorize endpoint. 2. A method for the Call back endpoint.
Both methods return URL for redirect or an error. And controller handle them and create the response.

login/service.go
```diff
@@ -337,6 +339,187 @@ func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, confi
...
+		RefreshToken: &keycloakToken.RefreshToken,
+		TokenType:    &keycloakToken.TokenType,
+	}
+
+	log.Debug(ctx, map[string]interface{}{
+		"code":         code,
+		"redirect_uri": redirect,
+	}, "exchanged code to access token")
+
+	identity, newUser, err := keycloak.CreateOrUpdateIdentity(ctx, keycloakToken.AccessToken, serviceConfig)
```
alexeykazakov: I see that there is a lot of copy&paste from Perform(). Reuse the common code instead.

login/service.go
```diff
@@ -377,6 +560,28 @@ func (keycloak *KeycloakOAuthProvider) saveParams(ctx *app.LoginLoginContext, re
 	return &redirect, nil
 }
 
+func (keycloak *KeycloakOAuthProvider) saveParamsForAuthorize(ctx *app.AuthorizeAuthorizeContext, redirect string) (*string, error) {
```
alexeykazakov: Re-use saveParams() instead of duplicating the code.

login/service.go
```diff
@@ -752,6 +957,11 @@ func redirectWithError(ctx *app.LoginLoginContext, knownReferrer string, errorSt
 	return ctx.TemporaryRedirect()
 }
 
+func redirectWithErrorForAuthorize(ctx *app.AuthorizeAuthorizeContext, knownReferrer string, errorString string) error {
```
alexeykazakov: Duplication again. Let's not use this function at all. Just return an JSON error when needed.

login/service.go
kishansagathiya: I can modify perform function and break it into smaller functions so that can be used for both

controller/authorize.go
```diff
@@ -0,0 +1,85 @@
...
+func NewAuthorizeController(service *goa.Service, auth *login.KeycloakOAuthProvider, tokenManager token.Manager, configuration LoginConfiguration) *AuthorizeController {
+	return &AuthorizeController{Controller: service.NewController("AuthorizeController"), Auth: auth, TokenManager: tokenManager, Configuration: configuration}
+}
+
+// Authorize runs the authorize action.
+func (c *AuthorizeController) Authorize(ctx *app.AuthorizeAuthorizeContext) error {
+
+	var scope []string
+
+	if ctx.Code == nil {
```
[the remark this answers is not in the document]
kishansagathiya: I can create /api/login/callback endpoint as well for login. It would keep a common flow.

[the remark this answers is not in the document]
alexeykazakov: +1 (it could be in a separate PR though)

login/service.go
alexeykazakov: +1

design/login.go
```diff
@@ -30,6 +48,35 @@ var _ = a.Resource("login", func() {
...
+			a.Param("state", d.UUID, "")
+			a.Param("code", d.String, "authorization_code")
+			a.Param("api_client", d.String, "The name of the api client which is requesting a token")
+			a.Required("state")
+		})
+		a.Description("Authorize service client")
+		a.Response(d.Unauthorized, JSONAPIErrors)
+		a.Response(d.TemporaryRedirect)
+		a.Response(d.InternalServerError, JSONAPIErrors)
+		a.Response(d.BadRequest, JSONAPIErrors)
```
sbose78: Please add at least one test for each of the above response possibilities.

login/service_blackbox_test.go
```diff
@@ -690,3 +691,156 @@ func (c *dummyOauth2Config) Exchange(ctx netcontext.Context, code string) (*oaut
...
+
+func (s *serviceBlackBoxTest) TestInvalidOAuthStateForAuthorize() {
+
+	rw, callbackCtx := s.AuthorizeCallback("invalid_state", make(map[string]string))
+	_, err := s.loginService.VerifyState(callbackCtx, callbackCtx.State.String(), callbackCtx.Code)
+	require.NotNil(s.T(), err)
+	assert.Equal(s.T(), 401, rw.Code)
+}
+
+func (s *serviceBlackBoxTest) AuthorizeCallback(testType string, extraParams map[string]string) (*httptest.ResponseRecorder, *app.CallbackAuthorizeContext) {
```
sbose78: This method doesn't need to be exposed outside the package.

controller/authorize.go
```diff
@@ -0,0 +1,88 @@
...
+	ctx.ResponseData.Header().Set("Cache-Control", "no-cache")
+	return c.Auth.PerformAuthorize(ctx, oauth, c.Configuration)
+}
+
+// Callback takes care of Authorize callback
+func (c *AuthorizeController) Callback(ctx *app.CallbackAuthorizeContext) error {
+
+	referrerURL, err := c.Auth.VerifyState(ctx, ctx.State.String(), ctx.Code)
+	if err != nil {
+		return err
```
sbose78: Since this is being returned by the controller, let's do a 
`return jsonapi.JSONErrorResponse(ctx,err)`

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,198 @@
...
+	test.AuthorizeAuthorizeTemporaryRedirect(t, svc.Context, svc, ctrl, nil, clientID, redirect, responseType, nil, state)
+}
+
+func (rest *TestAuthorizeREST) TestAuthorizeCallbackOK() {
+	t := rest.T()
+	_, ctrl := rest.UnSecuredController()
+
+	rw := httptest.NewRecorder()
+	u := &url.URL{
+		Path: fmt.Sprintf("/api/authorize"),
```
sbose78: I think you need to split your test into two tests, one for `/api/authorize` and another for `/api/authorize/callback` .

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,198 @@
...
+	prms := url.Values{}
+	prms.Add("code", "SOME_OAUTH2.0_CODE")
+	prms.Add("state", uuid.NewV4().String())
+
+	ctx := context.Background()
+
+	prms.Del("code")
+	goaCtx := goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
+	_, err = app.NewAuthorizeAuthorizeContext(goaCtx, req, goa.New("LoginService"))
+	require.NotNil(t, err)
```
sbose78: The code isn't checking if a 400 Bad request has been returned. the `rw` object should the response code.

login/service_blackbox_test.go
```diff
@@ -543,10 +543,11 @@ func (s *serviceBlackBoxTest) TestUnapprovedUserLoginUnauthorized() {
 	}
 
 	err = s.loginService.Perform(authorizeCtx, dummyOauth, s.Configuration)
-	require.NotNil(s.T(), err)
+	require.Nil(s.T(), err)
```
sbose78: This changes the test case, right?

controller/authorize_blackbox_test.go
kishansagathiya: Both controllers are in the same file. Doesn't it make sense to keep the tests also in the same file?

sbose78: yes same file is fine. But have different test methods.

controller/authorize.go
kishansagathiya: VerifyState returns the error as `jsonapi.JSONErrorResponse(ctx,err)`, do I still need to do this again?

controller/authorize_blackbox_test.go
kishansagathiya: We have made these parameters compulsory in design, so that we don't need to check.Since we have no checks, parameters will be accepted. It will result in 200. Do we want that?

login/service_blackbox_test.go
kishansagathiya: seems wrong to me. It is passing cause of dummyOauth. Will change this.

login/service_blackbox_test.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/24b633c4334765f4623bff85a089c206bae145b8

controller/authorize_blackbox_test.go
sbose78: I'm not suggesting you to not do `app.NewAuthorizeAuthorizeContext`, rather I'm suggesting that you could see if the `rw` object has information about the response?

controller/authorize_blackbox_test.go
kishansagathiya: I am not sure what you want. I think tests are already split. Lets discuss this.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-14`

controller/authorize.go
sbose78: You could have `VerifyState(..)` return a usual error and always ensure that the controller does the `jsonapi.JSONErrorResponse(ctx,...)` conversion ( kind of  responsibility :) )

controller/authorize_blackbox_test.go
sbose78: You should have 2 tests:

`TestAuthorizeOK()`
and
`TestAuthorizeCallbackOK()`

sbose78: So effectively you needn't split since you are using the response `/api/authorize` in `/api/authorize/callback` .
would be nice to have an explicit test written for `/api/authorize` even if it duplicates the first half of `TestAuthorizeCallbackOK`

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,198 @@
...
+	assert.Contains(t, rw.Header().Get("Location"), redirectURI)
+
+	locationString = rw.HeaderMap["Location"][0]
+	locationUrl, err = url.Parse(locationString)
+	require.Nil(t, err)
+
+	allQueryParameters = locationUrl.Query()
+
+	assert.NotNil(t, allQueryParameters)
+	assert.NotNil(t, allQueryParameters["state"][0])
```
sbose78: this should be equal to `returnedState` right ? would be nice to have an assert.Equal(..)

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	assert.Contains(t, rw.Header().Get("Location"), redirectURI)
+
+	locationString = rw.HeaderMap["Location"][0]
+	locationUrl, err = url.Parse(locationString)
+	require.Nil(t, err)
+
+	allQueryParameters = locationUrl.Query()
+
+	assert.NotNil(t, allQueryParameters)
```
sbose78: require.NotNil(..)

controller/authorize_blackbox_test.go
kishansagathiya: app.NewAuthorizeAuthorizeContext is setting compulsory fields to empty strings, so we won't see bad parameter error on this api. I can add checks for empty strings.

controller/token.go
```diff
@@ -354,41 +354,114 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
 	if payload == nil {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("payload", "nil").Expected("not empty payload"))
 	}
-	if payload.ClientID == nil {
-		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_id", "nil").Expected("Service Account ID"))
-	}
-	if payload.ClientSecret == nil {
-		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("client_secret", "nil").Expected("Service Account secret"))
-	}
+	const clientCredentials = "client_credentials"
```
sbose78: move the `const` declarations to the top of the file.

controller/token.go
```diff
@@ -354,41 +354,114 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
...
+			ClientSecret: c.Configuration.GetKeycloakSecret(),
+			Endpoint:     oauth2.Endpoint{AuthURL: authEndpoint, TokenURL: tokenEndpoint},
+			RedirectURL:  rest.AbsoluteURL(ctx.RequestData, "/api/authorize/callback"),
+		}
+
+		ctx.ResponseData.Header().Set("Cache-Control", "no-cache")
+		keycloakToken, err := c.Auth.GetTokenFromAuthorizationCode(ctx, *payload.Code, oauth)
+
+		if err != nil || keycloakToken == nil {
+			return err
```
sbose78: You need to do a `return jsonapi.JSONErrorResponse(..)`

controller/token.go
```diff
@@ -354,41 +354,114 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
...
+				"redirectURL": oauth.RedirectURL,
+				"err":         err,
+			}, "failed to parse referrer")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+		}
+
+		_, err = c.Auth.CreateOrUpdateIdentityAndUser(ctx, *payload.Code, redirectURL, keycloakToken, ctx.RequestData, oauth, c.Configuration)
+
+		if err != nil {
+			return err
```
sbose78: should do a `return jsonapi.JSONErrorResponse(..)`

controller/token.go
```diff
@@ -354,41 +354,114 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
...
+
 	log.Error(ctx, map[string]interface{}{
-		"client_id":     *payload.ClientID,
-		"client_secret": *payload.ClientSecret,
-	}, "Service Account secret doesn't match")
-	return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("invalid Service Account ID or secret"))
+		"grant_type": payload.GrantType,
+	}, "Invalid grant_type")
+
+	return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("grant_type", "nil").Expected("grant_type=client_credentials or grant_type=authorization_code"))
```
sbose78: Pass `payload.GrantType` instead of "nil" ?

sbose78:
@kishansagathiya could you please document in comments the workflows you have tested out manually ?

I can think of 
1. Login ( existing way ) with & without `link`. ( using the API like `/api/login?redirect=http://localhost:8089/api/status` )
2. /api/token/exchange for service accounts

let's ensure there's no regression on these use cases :)

kishansagathiya:
@sbose78 Just test all mentioned cases to be working

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-15`

login/service.go
```diff
@@ -328,15 +181,222 @@ func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, confi
...
+		log.Info(ctx, map[string]interface{}{
+			"referrerURL": referrerURL.String(),
+			"user_name":   identity.Username,
+			"api_client":  apiClient,
+		}, "all good; redirecting back to referrer")
+		return &redirectTo, nil
+	}
+
+	// If the 'initlinking" param == true then initiate account linking if not already linked
+	linked, err := keycloak.checkAllFederatedIdentities(ctx, keycloakToken.AccessToken, brokerEndpoint)
```
alexeykazakov: We can now remove the all code related to linking accounts from this method. We do not use auto-linking (to github or OSO) during login anyway. Just return the redirect url without checking if the account is linked or not.

login/service.go
```diff
@@ -64,7 +64,12 @@ type KeycloakOAuthProvider struct {
...
+	GetTokenFromAuthorizationCode(ctx context.Context, code string, config oauth.OauthConfig) (*oauth2.Token, error)
+	VerifyState(ctx context.Context, state string, code string) (*url.URL, error)
 	CreateOrUpdateIdentity(ctx context.Context, accessToken string, configuration LoginServiceConfiguration) (*account.Identity, bool, error)
+	CreateOrUpdateIdentityAndUser(ctx context.Context, code string, referrerURL *url.URL, keycloakToken *oauth2.Token, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error)
+	BeforeRedirectToLogin(ctx context.Context, redirect *string, link *bool, apiClient *string, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error)
 	Link(ctx *app.LinkLinkContext, brokerEndpoint string, clientID string, validRedirectURL string) error
 	LinkSession(ctx *app.SessionLinkContext, brokerEndpoint string, clientID string, validRedirectURL string) error
 	LinkCallback(ctx *app.CallbackLinkContext, brokerEndpoint string, clientID string) error
```
alexeykazakov: I would suggest:
1. Rename Perform() -> Login()
2. PerformAuthorize() -> AuthCodeURL() and just return an URL here or error. The controller will create a proper redirect response. See token/link/link.go for example. Login() could re-use AuthCodeURL().
3. GetTokenFromAuthorizationCode() -> Exchange() 
So, AuthCodeURL() and Exchange() will match names used by oauth2.Config
4. VerifyState() do you need to export it? Looks like an internal service method. Create a new AuthCodeCallback() and call it from the controller. You can verify state inside this method so you don't have to expose it. Similar to what we have in token/link/link.go
5. VerifyState() is not accurate anyway. This method is actually removing the existing state too. Maybe something like reclaimReferrer() or something like this?
6. BeforeRedirectToLogin() why it's exported? Why it's in the interface? Looks like an internal method. saveAuthParams() ?
7. CreateOrUpdateIdentityAndUser vs CreateOrUpdateIdentity is misleading. Both methods create both user and identity. Maybe we should rename CreateOrUpdateIdentity() to CreateOrUpdateIdentityInDB() ? @sbose78 WDYT?

kishansagathiya: 2. We have a AuthCodeURL() method in OauthConfig, wont it be confusing?
5. reclaimReferrer() seems like a bettter name, but I dont see how it is removing existing state?
7. Probably we should rename the new method CreateOrUpdateIdentityAndUser() instead?

login/service.go
sbose78: ^ Good idea :) 
@kishansagathiya you no longer need to worry about his ;)

login/service.go
sbose78: @alexeykazakov sounds good - now that we aren't using the *linking during login* anymore, CreateOrUpdateIdentityInDB() looks fine. 

@kishansagathiya 
> 5 -> reclaimReferrer() seems like a bettter name, but I dont see how it is removing existing state?
the problem with the name `VerifyState()` was that it looked read-only and it does a lot more actually. 
`reclaimReferrer()` ensures the illusion of read-only is gone :)

> 2 -> We have a AuthCodeURL() method in OauthConfig, wont it be confusing?
Should be fine since the interfaces are different.
and the idea is to actually *match* the names with OauthConfig for uniformity's sake. :)

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-16`

login/service.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/ec079fc047eefd3dff08b139dd5344b90e52defe

controller/authorize.go
```diff
@@ -0,0 +1,88 @@
...
+
+	var scope []string
+
+	if ctx.Scope == nil {
+		scope = []string{"user:email"}
+	} else {
+		scope = []string{*ctx.Scope}
+	}
+
+	_, found := c.Configuration.GetServiceAccounts()[ctx.ClientID]
```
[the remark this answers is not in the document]
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/ec079fc047eefd3dff08b139dd5344b90e52defe

controller/authorize.go
```diff
@@ -0,0 +1,88 @@
...
+		}, "unable to get keycloak token endpoint url")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, errs.Wrap(err, "unable to get keycloak token endpoint url")))
+	}
+
+	oauth := &oauth2.Config{
+		ClientID:     c.Configuration.GetKeycloakClientID(),
+		ClientSecret: c.Configuration.GetKeycloakSecret(),
+		Scopes:       scope,
+		Endpoint:     oauth2.Endpoint{AuthURL: authEndpoint, TokenURL: tokenEndpoint},
+		RedirectURL:  rest.AbsoluteURL(ctx.RequestData, "/api/authorize/callback"),
```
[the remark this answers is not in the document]
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/ec079fc047eefd3dff08b139dd5344b90e52defe

login/service.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/210/commits/ec079fc047eefd3dff08b139dd5344b90e52defe

controller/authorize.go
```diff
@@ -38,12 +38,12 @@ func (c *AuthorizeController) Authorize(ctx *app.AuthorizeAuthorizeContext) erro
...
-	_, found := c.Configuration.GetServiceAccounts()[ctx.ClientID]
-	if !found {
+	// Default value of this public client id is set to "740650a2-9c44-4db5-b067-a3d1b2cd2d01"
+	if ctx.ClientID != c.Configuration.GetPublicOauthClientID() {
 		log.Error(ctx, map[string]interface{}{
 			"client_id": ctx.ClientID,
-		}, "unknown service account id")
-		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("invalid service account id"))
+		}, "unknown oauth client id")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("invalid oauth client id"))
```
alexeykazakov: Looks good now.

controller/authorize.go
```diff
@@ -67,22 +67,14 @@ func (c *AuthorizeController) Authorize(ctx *app.AuthorizeAuthorizeContext) erro
 		ClientSecret: c.Configuration.GetKeycloakSecret(),
 		Scopes:       scope,
 		Endpoint:     oauth2.Endpoint{AuthURL: authEndpoint, TokenURL: tokenEndpoint},
-		RedirectURL:  rest.AbsoluteURL(ctx.RequestData, "/api/authorize/callback"),
+		RedirectURL:  rest.AbsoluteURL(ctx.RequestData, client.CallbackAuthorizePath()),
 	}
 
 	ctx.ResponseData.Header().Set("Cache-Control", "no-cache")
-	return c.Auth.PerformAuthorize(ctx, oauth, c.Configuration)
+	return c.Auth.AuthCodeURL(ctx, oauth, c.Configuration)
```
alexeykazakov: You still set the location header in the service? As I mentioned a few times it's better if the service (c.Auth.AuthCodeURL(), AuthCodeCallback()) returns URLs only. Then you set the location header in the controller.

login/service.go
```diff
@@ -138,16 +137,16 @@ func (keycloak *KeycloakOAuthProvider) Perform(ctx *app.LoginLoginContext, confi
...
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 	ctx.ResponseData.Header().Set("Location", *redirectURL)
 	return ctx.TemporaryRedirect()
 }
 
-// BeforeRedirectToLogin takes care of things parameter and state saving
-func (keycloak *KeycloakOAuthProvider) BeforeRedirectToLogin(ctx context.Context, redirect *string, link *bool, apiClient *string, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error) {
+// beforeRedirectToLogin saves referer and api_client parameter
+func (keycloak *KeycloakOAuthProvider) beforeRedirectToLogin(ctx context.Context, redirect *string, apiClient *string, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error) {
```
alexeykazakov: Still don't like the method name. saveAuthParams() or saveParamsAndReferrer() ?

login/service.go
```diff
@@ -316,59 +308,36 @@ func (keycloak *KeycloakOAuthProvider) CreateOrUpdateIdentityAndUser(ctx context
...
-	log.Debug(ctx, map[string]interface{}{
-		"code":        code,
-		"referrerURL": referrerURL.String(),
-		"user_name":   identity.Username,
-		"linked":      linked,
-	}, "linking identities...")
-
-	return keycloak.autoLinkProvidersDuringLogin(ctx, request, keycloakToken.AccessToken, referrerStr)
-
+	ctx.ResponseData.Header().Set("Location", *redirectURL)
```
alexeykazakov: This is what I was referring to above. Le't set the header and return the redirect in the controller. Here, just return the location URL.

login/service.go
```diff
@@ -316,59 +308,36 @@ func (keycloak *KeycloakOAuthProvider) CreateOrUpdateIdentityAndUser(ctx context
...
+// AuthCodeCallback takes care of authorization callback.
+// When authorization_code is requested with /api/authorize, keycloak would return authorization_code at /api/authorize/callback,
+// which would pass on the code along with the state to client using this method
+func (keycloak *KeycloakOAuthProvider) AuthCodeCallback(ctx *app.CallbackAuthorizeContext) error {
+	referrerURL, err := keycloak.reclaimReferrer(ctx, ctx.State.String(), ctx.Code)
 	if err != nil {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
-	ctx.ResponseData.Header().Set("Location", *redirectURL)
+	ctx.ResponseData.Header().Set(
```
alexeykazakov: The same as above.

controller/authorize.go
kishansagathiya: And same with the Login()(Perform() earlier) method?

alexeykazakov: No. Perform/Login is very old method which handed redirection. Leave it as is.
But new methods like AuthCodeURL() and AuthCodeCallback() should return URLs as we do in https://github.com/fabric8-services/fabric8-auth/blob/master/token/link/link.go#L127 or https://github.com/fabric8-services/fabric8-auth/blob/master/token/link/link.go#L255

kishansagathiya:
@sbose78 @alexeykazakov Thinking of movies lot of authorize related tests from service_blackbox_test.go to authorize_blackbox_test.go. I guess that would be a better place. WDYT?

sbose78:
@kishansagathiya If you wish to avoid having files which grow too big and wish to group them in a different `.go` file, that's encouraged :) Let's name it `service_authorize_blackbox_test.go`

kishansagathiya:
@sbose78 It's not just that. I think it would make more sense to have it in controller tests. Since I am not just testing methods. We would like to check status code as well and now none of the methods(except Login()) are setting status codes at method level. Status codes are being set in Controllers.

sbose78:
@kishansagathiya status codes are verified inside the auto-generated tests like `test.ListUsersOK()` method, would that work for you?

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-17`

login/service.go
```diff
@@ -137,16 +137,16 @@ func (keycloak *KeycloakOAuthProvider) Login(ctx *app.LoginLoginContext, config
...
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 	ctx.ResponseData.Header().Set("Location", *redirectURL)
 	return ctx.TemporaryRedirect()
 }
 
-// beforeRedirectToLogin saves referer and api_client parameter
-func (keycloak *KeycloakOAuthProvider) beforeRedirectToLogin(ctx context.Context, redirect *string, apiClient *string, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error) {
+// saveParamsAndReferrer saves referer and api_client parameter
+func (keycloak *KeycloakOAuthProvider) saveParamsAndReferrer(ctx context.Context, redirect *string, apiClient *string, request *goa.RequestData, config oauth.OauthConfig, serviceConfig LoginServiceConfiguration) (*string, error) {
```
alexeykazakov: I'm sorry. I didn't actually look at this method closely before. saveParamsAndReferrer is not accurate name for this method. This method saves the param & referrer but also returns the keycloak redirect URL.

This method actually should be the AuthCodeURL() method. It's the same method. saveParamsAndReferrer() == AuthCodeURL(). So just rename it to AuthCodeURL().

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-18`

login/service.go
```diff
@@ -331,7 +321,12 @@ func (keycloak *KeycloakOAuthProvider) AuthCodeCallback(ctx *app.CallbackAuthori
 		return nil, err
 	}
 
-	redirectTo := referrerURL.String() + "?code=" + ctx.Code + "&state=" + ctx.State.String()
+	parameters := url.Values{}
```
alexeykazakov: `parameters := referrerURL.Query()`
Otherwise you will lost existing params in referrerURL.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	ctx := context.Background()
+
+	prms.Del("response_type")
+	goaCtx := goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
+	authorizeCtx, err := app.NewAuthorizeAuthorizeContext(goaCtx, req, goa.New("LoginService"))
+	require.NotNil(t, err)
+	require.Contains(t, err.Error(), "400 invalid_request: missing required parameter")
+	jsonapi.JSONErrorResponse(authorizeCtx, err)
+	require.Equal(t, 400, rw.Code)
+	prms.Add("response_type", "code")
```
alexeykazakov: Lines 71-78 are mostly duplicated a few times for every param. Extract this code to a separate function and re-use it in the test for all params.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	goaCtx := goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
+	authorizeCtx, err := app.NewAuthorizeAuthorizeContext(goaCtx, req, goa.New("LoginService"))
+	require.Nil(t, err)
+
+	err = ctrl.Authorize(authorizeCtx)
+	require.Nil(t, err)
+
+	require.Equal(t, 307, rw.Code) // redirect to keycloak login page.
+
+	locationString := rw.HeaderMap["Location"][0]
```
alexeykazakov: Verify that the locationString == the expected URL.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	require.Nil(t, err)
+
+	err = ctrl.Callback(callbackCtx)
+	require.Nil(t, err)
+
+	require.Equal(t, 307, rw.Code) // redirect to keycloak login page.
+
+	require.Contains(t, rw.Header().Get("Location"), redirectURI)
+
+	locationString = rw.HeaderMap["Location"][0]
```
alexeykazakov: Verify that the locationString == the expected URL.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	locationUrl, err = url.Parse(locationString)
+	require.Nil(t, err)
+
+	allQueryParameters = locationUrl.Query()
+
+	require.NotNil(t, allQueryParameters)
+	require.NotNil(t, allQueryParameters["state"][0])
+	require.NotNil(t, allQueryParameters["code"][0])
+	require.Equal(t, returnedState, allQueryParameters["state"][0])
+}
```
alexeykazakov: Call the same call back one more time to make sure it now fails (because the corresponding state does not exist anymore).

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	locationString = rw.HeaderMap["Location"][0]
+	locationUrl, err = url.Parse(locationString)
+	require.Nil(t, err)
+
+	allQueryParameters = locationUrl.Query()
+
+	require.NotNil(t, allQueryParameters)
+	require.NotNil(t, allQueryParameters["state"][0])
+	require.NotNil(t, allQueryParameters["code"][0])
```
alexeykazakov: Check if the code == the expected code.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	require.Contains(t, rw.Header().Get("Location"), redirectURI)
+
+	locationString = rw.HeaderMap["Location"][0]
+	locationUrl, err = url.Parse(locationString)
+	require.Nil(t, err)
+
+	allQueryParameters = locationUrl.Query()
+
+	require.NotNil(t, allQueryParameters)
+	require.NotNil(t, allQueryParameters["state"][0])
```
alexeykazakov: Check if the state == the expected state.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	returnedState := allQueryParameters["state"][0]
+
+	u = &url.URL{
+		Path: fmt.Sprintf(client.CallbackAuthorizePath()),
+	}
+
+	prms = url.Values{
+		"state": {returnedState},
+		"code":  {"SOME_OAUTH2.0_CODE"},
```
alexeykazakov: Generate a new code here instead of using a hardcoded value. And check it below.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	req, err := http.NewRequest("GET", u.String(), nil)
+	require.Nil(t, err)
+
+	prms := url.Values{}
+	prms.Add("code", "SOME_OAUTH2.0_CODE")
+	prms.Add("state", uuid.NewV4().String())
+
+	ctx := context.Background()
+
+	prms.Del("code")
```
alexeykazakov: What's the point of adding a param in line 213 and then remove it here? :) Just don't add it in the first place.

controller/authorize_blackbox_test.go
```diff
@@ -0,0 +1,301 @@
...
+	goaCtx := goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
+	callbackCtx, err := app.NewAuthorizeAuthorizeContext(goaCtx, req, goa.New("LoginService"))
+	require.NotNil(t, err)
+	require.Contains(t, err.Error(), "400 invalid_request: missing required parameter")
+	jsonapi.JSONErrorResponse(callbackCtx, err)
+	require.Equal(t, 400, rw.Code)
+
+	prms.Del("state")
+	goaCtx = goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
```
alexeykazakov: Add code param to make sure it fails because of missing state. Not code. And extract this code actually in a separate function and reuse for both params.

controller/token_blackbox_test.go
```diff
@@ -135,6 +169,37 @@ func (rest *TestTokenREST) checkServiceAccountCredentials(name string, id string
...
+	saToken, err := rest.dummyOauth.Exchange(service.Context, code)
+	assert.Nil(rest.T(), err)
+	assert.NotNil(rest.T(), saToken.TokenType)
+	assert.Equal(rest.T(), "bearer", saToken.TokenType)
+	assert.NotNil(rest.T(), saToken.AccessToken)
+}
+
+func (c *dummyOauth2Config) Exchange(ctx netcontext.Context, code string) (*oauth2.Token, error) {
+	claims := make(map[string]interface{})
+	accessToken, err := testtoken.GenerateTokenWithClaims(claims)
```
alexeykazakov: Can be just some string actually. String will be easier to validate.

controller/token_blackbox_test.go
```diff
@@ -135,6 +169,37 @@ func (rest *TestTokenREST) checkServiceAccountCredentials(name string, id string
 	assert.True(rest.T(), token.IsSpecificServiceAccount(ctx, []string{name}))
 }
 
+func (rest *TestTokenREST) checkAuthorizationCode(name string, code string) {
+	service, _ := rest.SecuredController()
+	saToken, err := rest.dummyOauth.Exchange(service.Context, code)
+	assert.Nil(rest.T(), err)
+	assert.NotNil(rest.T(), saToken.TokenType)
+	assert.Equal(rest.T(), "bearer", saToken.TokenType)
+	assert.NotNil(rest.T(), saToken.AccessToken)
```
alexeykazakov: Check if saToken.AccessToken and saToken.RefreshToken == expected tokens.

controller/token_blackbox_test.go
```diff
@@ -135,6 +169,37 @@ func (rest *TestTokenREST) checkServiceAccountCredentials(name string, id string
...
+	if err != nil {
+		panic(err)
+	}
+
+	var thirtyDays int64
+	thirtyDays = 60 * 60 * 24 * 30
+	token := &oauth2.Token{
+		TokenType:    "bearer",
+		AccessToken:  accessToken,
+		RefreshToken: "someRefreshToken",
```
alexeykazakov: Generate some random strings (in suite or test setup) and use them as RefreshToken and AssessToken. Then validate these tokens in checkAuthorizationCode()

login/service_blackbox_test.go
```diff
@@ -690,3 +692,171 @@ func (c *dummyOauth2Config) Exchange(ctx netcontext.Context, code string) (*oaut
...
+
+	prms.Add("response_type", "code")
+	prms.Add("redirect_uri", "https://openshift.io/somepath")
+	prms.Add("client_id", "740650a2-9c44-4db5-b067-a3d1b2cd2d01")
+	prms.Add("state", uuid.NewV4().String())
+
+	ctx := context.Background()
+	goaCtx := goa.NewContext(goa.WithAction(ctx, "AuthorizeTest"), rw, req, prms)
+	authorizeCtx, err := app.NewAuthorizeAuthorizeContext(goaCtx, req, goa.New("LoginService"))
+	if err != nil {
```
alexeykazakov: Remove this block.

controller/authorize_blackbox_test.go
kishansagathiya: Isn't ```require.Contains(t, rw.Header().Get("Location"), redirectURI)``` enough here?

alexeykazakov: No, it's not enough. The location should be something like this: https://sso.openshift.io/auth/realm/....?redirect={redirectURI}
and you check if redirectURI is part of this URL. But we should also check if Location starts with https://sso.openshift.io/auth/realm/ or whatever is used in our test env. The URL to our Identity Provider which is Keycloak in this case.

kishansagathiya: The location is redirectURI?code={code}&state={state} (which is correct as per the way I have implemented). If it should be the way you are suggesting I will have to change the implementation. Callback should redirect to redirectURI along with code and state. I don't see a reason to redirect to https://sso.openshift.io/auth/realm/....?redirect={redirectURI} .

alexeykazakov: Ah, I didn't notice you asked about callback check. Yes, for callback URL it should be the original redirect. Not sso. SSO URL should be checked for authorize calls.

alexeykazakov: But it would be better if you check that the URL string starts with the expected string. Not just contains it (because in this case it will pass even if it's just in a param).

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-19`

controller/authorize_blackbox_test.go
```diff
@@ -152,8 +152,8 @@ func (rest *TestAuthorizeREST) TestAuthorizeCallbackOK() {
 	require.Equal(t, 307, rw.Code)
 
 	locationString = rw.HeaderMap["Location"][0]
-	require.Contains(t, locationString, redirectURI)
 	locationUrl, err = url.Parse(locationString)
+	require.Contains(t, redirectURI, locationUrl.Host+locationUrl.Path)
```
alexeykazakov: require.True(t, strings.HasPrefix(locationString, redirectURI))

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-20`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-21`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-22`

alexeykazakov:
api/authorize requires state param to be UUID. It should accept any string as state. We could address this in a follow up PR though.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-210-23`

Issue #212, opened 2017-12-01
Fetch user's username for linked Github/OpenShiftOnline account


sbose78:
Return the username in the existing API, in a new json field "username"

PR #214, opened 2017-12-02
Fetch user's username for linked Github/OpenShiftOnline account

alexeykazakov:
- [x] Add new username field to the External Tokens table
- [x] Load usernames from IdPs when linking accounts and save them in the token table
- [x] When retrieving the token check if the username is missing for old accounts, load username from IdP and save it in the DB
- [x] Return username as part of the token payload
- [x] Tests (fix existing, add new)

Fixes https://github.com/fabric8-services/fabric8-auth/issues/212


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-214-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-214-2`

token/oauth/oauth2.go
```diff
@@ -20,6 +22,64 @@ type OauthConfig interface {
...
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err":         err.Error(),
+			"profile_url": provider.ProfileURL,
+		}, "unable to read user profile payload")
+		return body, err
+	}
+	if res.StatusCode < 200 || res.StatusCode > 299 {
```
sbose78: :D nice.

errors/errors.go
```diff
@@ -31,6 +32,11 @@ func NewInternalError(ctx context.Context, err error) InternalError {
 	return InternalError{err}
 }
 
+// NewInternalErrorFromString returns the custom defined error of type InternalError.
+func NewInternalErrorFromString(ctx context.Context, errorMessage string) InternalError {
```
sbose78: Method overloading :) could think about something like this http://changelog.ca/log/2015/01/30/golang but doesn't look very clean to me.

token/link/github.go
```diff
@@ -1,45 +1,65 @@
...
 	provider.ClientID = clientID
 	provider.ClientSecret = clientSecret
 	provider.Endpoint = github.Endpoint
 	provider.RedirectURL = authURL + client.CallbackTokenPath()
-	provider.scopeStr = scopes
+	provider.ScopeStr = scopes
 	provider.Config.Scopes = strings.Split(scopes, " ")
-	provider.providerID, _ = uuid.FromString(gitHubProviderID)
+	provider.ProviderID, _ = uuid.FromString(GitHubProviderID)
+	provider.ProfileURL = "https://api.github.com/user"
```
sbose78: Should be in configuration.

token/link/link.go
```diff
@@ -205,6 +209,7 @@ func (service *LinkService) Callback(ctx context.Context, req *goa.RequestData,
 			// It was re-linking. Overwrite the existing link.
 			externalToken := tokens[0]
 			externalToken.Token = providerToken.AccessToken
+			externalToken.Username = userProfile.Username
```
sbose78: so reconnecting is how the username will be updated.

sbryzak:
@alexeykazakov Should I update the UI to display the new username field?

token/link/github.go
alexeykazakov: I don't think so. This is GitHub. It's not supposed to be configurable in different environments.

token/link/link.go
alexeykazakov: Yes.

errors/errors.go
alexeykazakov: It's not overloading actually. It's about allowing both string and error types in the param of this function. Alternative to this approach (two different functions) would be using interface{} as a param but not sure it's better because it's more error prone.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-214-3`

alexeykazakov:
The PR is ready for review now.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-214-4`

sbose78:
Planning to update the API in another pull request with a parameter something like "refresh_status=true"
that would try to fetch the username from github.

PR #218, opened 2017-12-06
Add option to pull externally linked account details

sbose78:
Add support for 
`/api/token?force_pull=true&for=https://github.com`

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -252,14 +252,11 @@ func (c *TokenController) Retrieve(ctx *app.RetrieveTokenContext) error {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 	if externalToken != nil {
-		if externalToken.Username == "" || (ctx.ForcePull != nil && *ctx.ForcePull == true) {
-			updatedToken, err := c.updateProfileIfEmpty(ctx, providerConfig, externalToken)
-			if err != nil {
-				return jsonapi.JSONErrorResponse(ctx, err)
-			}
-			appResponse = modelToAppExternalToken(updatedToken)
+		updatedToken, err := c.updateProfileIfEmpty(ctx, providerConfig, externalToken, ctx.ForcePull)
+		if err != nil {
+			return jsonapi.JSONErrorResponse(ctx, err)
 		}
-		appResponse = modelToAppExternalToken(*externalToken)
+		appResponse = modelToAppExternalToken(updatedToken)
 		return ctx.OK(&appResponse)
 	}
 
@@ -287,7 +284,7 @@ func (c *TokenController) Retrieve(ctx *app.RetrieveTokenContext) error {
 	if err != nil {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
-	updatedToken, err := c.updateProfileIfEmpty(ctx, providerConfig, externalToken)
+	updatedToken, err := c.updateProfileIfEmpty(ctx, providerConfig, externalToken, ctx.ForcePull)
 	if err != nil {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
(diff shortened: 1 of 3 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-218-1`

sbose78:
@alexeykazakov let me know your thoughts on having something like this. 
This piece can be used to even test out whether the token is valid.

controller/token.go
```diff
@@ -250,11 +252,14 @@ func (c *TokenController) Retrieve(ctx *app.RetrieveTokenContext) error {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 	if externalToken != nil {
-		updatedToken, err := c.updateProfileIfEmpty(ctx, providerConfig, externalToken)
-		if err != nil {
-			return jsonapi.JSONErrorResponse(ctx, err)
+		if externalToken.Username == "" || (ctx.ForcePull != nil && *ctx.ForcePull == true) {
```
alexeykazakov: 1. `if *ctx.ForcePull == true {...}` looks weird :) `if *ctx.ForcePull {...}` would look better ;)
2. Let's move the check of this param to the `updateProfileIfEmpty(ctx, config, token, force)` method?

controller/token_storage_blackbox_test.go
```diff
@@ -93,7 +93,7 @@ func (rest *TestTokenStorageREST) checkRetrieveOSOServiceAccountToken(saName str
 	service, controller := rest.SecuredControllerWithServiceAccount(sa)
 	require.True(rest.T(), len(rest.Configuration.GetOSOClusters()) > 0)
 	for _, cluster := range rest.Configuration.GetOSOClusters() {
-		_, tokenResponse := test.RetrieveTokenOK(rest.T(), service.Context, service, controller, cluster.URL)
+		_, tokenResponse := test.RetrieveTokenOK(rest.T(), service.Context, service, controller, cluster.URL, nil)
```
alexeykazakov: Add some tests when the force param is not nil and handled properly.

design/token.go
```diff
@@ -38,6 +38,7 @@ var _ = a.Resource("token", func() {
 		)
 		a.Params(func() {
 			a.Param("for", d.String, "The resource for which the external token is being fetched, example https://github.com/fabric8-services/fabric8-auth or https://api.starter-us-east-2.openshift.com")
+			a.Param("force_pull", d.Boolean, "Pull the user's details for the specific connected account, example, the user's updated github username would be fetched from github")
```
alexeykazakov: Mention that if it's not set or "false" then the user profile will be pulled only if it's missing in Auth. Otherwise the profile stored in Auth will be used instead of updating it from the provider.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-218-2`

design/token.go
sbose78: ad55329

Done.

controller/token_storage_blackbox_test.go
sbose78: Added 2 tests:
06656cecf2653d2e3d7fdf6d064543083f9a0323

1. Token present in DB, but error during interaction with github.com
2. Token present in DB, & success during interaction with github.com

design/token.go
sbose78: Done in 06656cecf2653d2e3d7fdf6d064543083f9a0323

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-218-3`

PR #227, opened 2017-12-14
Email verification

### LINKED ISSUE #62: User's emails should not be updated without verification
Currently users can change email without verification

sbose78:
fixes #62 

- [x] Add template in fabric8-services/fabric8-notification
- [x] Send out email 
- [x] Controller for verifying and generation code.
- [x] Data model
- [x] Tests

Will make *keycloak interaction* changes in a separate PR


account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
+package email
+
+import "github.com/fabric8-services/fabric8-auth/application"
```
alexeykazakov: Let's use import() for consistency ;)

account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
...
+type EmailVerificationService interface {
+	SendVerificationCode(ctx context.Context, user account.User) error
+	VerifyCode(ctx context.Context, code string) (bool, error)
+}
+
+type EmailVerificationClient struct {
+	db application.DB
+}
+
+// NewKeycloakIDPServiceClient creates a new Keycloakc
```
alexeykazakov: ? :)

account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Info(ctx, map[string]interface{}{
+			"email": user.Email,
+		}, "verification code to be sent")
+		err := appl.VerificationCodes().Create(ctx, &newVerificationCode)
+		/*
+			TODO: Invoke the EmailService to send out an email
+		*/
+		return err
+	})
+	return nil
```
alexeykazakov: it will never happen, right?

account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
...
+		Verified: false,
+	}
+
+	return application.Transactional(c.db, func(appl application.Application) error {
+		log.Info(ctx, map[string]interface{}{
+			"email": user.Email,
+		}, "verification code to be sent")
+		err := appl.VerificationCodes().Create(ctx, &newVerificationCode)
+		/*
+			TODO: Invoke the EmailService to send out an email
```
alexeykazakov: Send email outside of the transaction.

account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		user.EmailVerified = true
+		err = appl.Users().Save(ctx, user)
+		if err != nil {
+			return err
+		}
+
+		verificationCode.Verified = true
+		err = appl.VerificationCodes().Save(ctx, verificationCode)
```
alexeykazakov: Verification code is a temporal thing.
- Create it & Save in DB
- Send email
- Check it & hard delete from DB

account/verification_code.go
```diff
@@ -0,0 +1,207 @@
...
+	errs "github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+type VerificationCode struct {
+	gormsupport.Lifecycle
+	ID       uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+	User     User
+	Verified bool // The verification status of the updated email.
+	Code     string
```
alexeykazakov: Let's maybe use ID as a code?

account/verification_code.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/jinzhu/gorm"
+	errs "github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+type VerificationCode struct {
+	gormsupport.Lifecycle
+	ID       uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+	User     User
+	Verified bool // The verification status of the updated email.
```
alexeykazakov: We don't really need verified field, do we?

account/verification_code.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/fabric8-services/fabric8-auth/log"
+
+	"github.com/goadesign/goa"
+	"github.com/jinzhu/gorm"
+	errs "github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+type VerificationCode struct {
+	gormsupport.Lifecycle
```
alexeykazakov: Should be hard-deletable.

account/verification_code.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// TableName overrides the table name settings in Gorm to force a specific table name
+// in the database.
+func (m VerificationCode) TableName() string {
+	return "verification_codes"
+}
+
+// GetETagData returns the field values to use to generate the ETag
+func (m VerificationCode) GetETagData() []interface{} {
```
alexeykazakov: No any ETags.

account/verification_code.go
```diff
@@ -0,0 +1,207 @@
...
+
+// VerificationCodeRepository represents the storage interface.
+type VerificationCodeRepository interface {
+	repository.Exister
+	Load(ctx context.Context, id uuid.UUID) (*VerificationCode, error)
+	LoadByCode(ctx context.Context, code string) (*VerificationCode, error)
+	Create(ctx context.Context, VerificationCode *VerificationCode) error
+	Save(ctx context.Context, VerificationCode *VerificationCode) error
+	Delete(ctx context.Context, id uuid.UUID) error
+	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]VerificationCode, error)
```
alexeykazakov: No need for Query. Create(), Load() and Delete() I thin is the only methods we need.

design/account.go
```diff
@@ -114,6 +114,20 @@ var _ = a.Resource("user", func() {
...
 	a.BasePath("/users")
 
+	a.Action("verifyEmail", func() {
+		a.Routing(
+			a.GET("/verifyemail"),
+		)
+		a.Params(func() {
+			a.Param("code", d.String, "code")
+		})
+		a.UseTrait("conditional")
```
alexeykazakov: ?

migration/sql-files/012-add-email-verified.sql
```diff
@@ -0,0 +1,29 @@
+ALTER TABLE users ADD COLUMN email_verified BOOLEAN DEFAULT false;
+UPDATE users SET email_verified = true;
```
alexeykazakov: +1 I think it's OK if we just set all existing emails to true for now. IMO we can afford skip Auth-Keycloak migration at this point.

migration/sql-files/012-add-email-verified.sql
```diff
@@ -0,0 +1,29 @@
...
+--	gormsupport.Lifecycle
+--	ID       uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+-- 	User     User
+--	Verified bool // The verification status of the updated email.
+--	Code     string
+-- }
+--
+
+
+CREATE TABLE external_provider_tokens (
```
alexeykazakov: ? :)

migration/sql-files/012-add-email-verified.sql
```diff
@@ -0,0 +1,29 @@
...
+    id uuid primary key DEFAULT uuid_generate_v4() NOT NULL,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    deleted_at timestamp with time zone,
+    user_id uuid REFERENCES users(id),
+    verified boolean DEFAULT false,
+    code text not null,
```
alexeykazakov: As I already mentioned above I think we can have only id (==code) and user_id in this table.

account/email/verification_service.go
```diff
@@ -0,0 +1,95 @@
...
+		/*
+			TODO: Invoke the EmailService to send out an email
+		*/
+		return err
+	})
+	return nil
+}
+
+// VerifyCode validates whether the code is present in our database.
+func (c *EmailVerificationClient) VerifyCode(ctx context.Context, code string) (bool, error) {
```
alexeykazakov: I think VerifyEmail() would be a better name. This method should:
- Load the code
- If found, set user.email_verfied = true
- Hard delete the code

account/email/verification_service.go
sbose78: Since emails are a critical piece, I was thinking of keeping the code ( database row ) after user *confirms* . That'll help us audit when the user actually approved his email change.

account/verification_code.go
sbose78: In the event where we don't hard-delete ( my initial thought as per https://github.com/fabric8-services/fabric8-auth/pull/227#discussion_r157127249 ) , this field would be needed to ensure one doesn't use the same confirmation url twice.

account/verification_code.go
sbose78: We could and it would work, but didn't want to mix-up the entity ID with the what the entity contains 
:(

account/verification_code.go
sbose78: Had kept Query() for using with Filter..() .. in case we ever wanted to audit 'how many times did a user update his email and when'

account/email/verification_service.go
sbose78: and how many times did he attempt a change to be precise.

alexeykazakov: OK. Good point. But let's maybe at least do a soft delete when the email is approved? No need to use any verified field, right?
- Create code & Save in DB
- Send email
- Check it & soft delete from DB

We can always audit soft deleted records if needed. We can also clean up unused codes later (old codes which were never deleted).

account/verification_code.go
alexeykazakov: He couldn't because the code would be soft deleted, could he?

account/email/verification_service.go
sbose78: Agreed. Soft delete sounds perfect. And `verified` field wouldn't be needed.

account/verification_code.go
sbose78: Right - now that we decided to soft-delete. 
I will remove the `verified` field.

account/verification_code.go
alexeykazakov: OK. Let's stick with separate code field. We will have more flexible implementation and can change the code generation anytime (use some more complex/long strings for example).

account/verification_code.go
alexeykazakov: We can always add it to the code if/when we need it, right? Don't want to add blow up our code when not necessary. I would expect we will need to audit the DB directly if we need it. Query&Filters are useless unless we add some secured API or even GUI to it.

account/email/verification_service.go
sbose78: Will happen after we add the `send email` code outside of the transaction block.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-3`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-4`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-5`

sbose78:
Need to add a new template for an email verification notification:
https://github.com/fabric8-services/fabric8-notification/tree/master/template

and then call the notification service the way it's been done here:
https://github.com/fabric8-services/fabric8-wit/commit/b38d8c00fe31743c2152d2206046e8304e6abe9d

account/verification_code.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/jinzhu/gorm"
+	errs "github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+type VerificationCode struct {
+	gormsupport.Lifecycle
+	ID     uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+	User   User
+	UserID NullUUID `sql:"type:uuid"`
```
alexeykazakov: UserID can't be null, right? So UserID should be uuid.UUID

account/verification_code.go
```diff
@@ -0,0 +1,189 @@
...
+
+// LoadByCode loads result by filtering with respect to the verificate code.
+func (m *GormVerificationCodeRepository) LoadByCode(ctx context.Context, code string) ([]VerificationCode, error) {
+	return m.Query(VerificationCodeWithUser(), VerificationCodeFilterByCode(code)) // maybe load with user?
+}
+
+// CheckExists returns nil if the given ID exists otherwise returns an error
+func (m *GormVerificationCodeRepository) CheckExists(ctx context.Context, id string) error {
+	defer goa.MeasureSince([]string{"goa", "db", "VerificationCode", "exists"}, time.Now())
+	return repository.CheckHardDeletableExists(ctx, m.db, m.TableName(), id)
```
alexeykazakov: We soft delete.

account/verification_code.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Save modifies a single record.
+func (m *GormVerificationCodeRepository) Save(ctx context.Context, model *VerificationCode) error {
+	defer goa.MeasureSince([]string{"goa", "db", "VerificationCode", "save"}, time.Now())
+
+	obj, err := m.Load(ctx, model.ID)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"verification_code_id": model.ID,
+			"ctx": ctx,
```
alexeykazakov: No need to add ctx explicitly. It will be added by log.Error()

controller/users.go
```diff
@@ -588,6 +595,10 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 
+	if isEmailVerificationNeeded {
+		c.sendVerificationEmail(ctx, *user)
```
alexeykazakov: Check the error. But we probably should just log the error here and proceed...

controller/users.go
```diff
@@ -632,6 +643,12 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
 	return ctx.OK(ConvertToAppUser(ctx.RequestData, user, identity))
 }
 
+func (c *UsersController) sendVerificationEmail(ctx context.Context, user account.User) error {
```
alexeykazakov: Any specific reason to define this method instead of just using c.EmailVerificationService.SendVerificationCode(ctx, user) directly where needed?

design/account.go
```diff
@@ -114,6 +114,19 @@ var _ = a.Resource("user", func() {
 var _ = a.Resource("users", func() {
 	a.BasePath("/users")
 
+	a.Action("verifyEmail", func() {
+		a.Routing(
```
alexeykazakov: Add description.

design/account.go
```diff
@@ -114,6 +114,19 @@ var _ = a.Resource("user", func() {
...
+			a.GET("/verifyemail"),
+		)
+		a.Params(func() {
+			a.Param("code", d.String, "code")
+			a.Required("code")
+		})
+		a.Response(d.OK, emailApprovedData)
```
alexeykazakov: We should return a redirect instead. User opens a link in a browser https://auth.openshift.io/api/users/verifyemail?code=... and after successful verification we redirect to "your email has been verified" page. Please create two issues for UX team: 
1) Email template
2) "your email has been verified" page (we should add the relative link to this page to our configuration and set it by default so we don't need to configure it for every cluster (preview/prod)). For now it can be _home page until we have this new page available.

migration/sql-files/012-add-email-verified.sql
```diff
@@ -0,0 +1,15 @@
...
+
+CREATE TABLE verification_codes (
+    id uuid primary key DEFAULT uuid_generate_v4() NOT NULL,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    deleted_at timestamp with time zone,
+    user_id uuid REFERENCES users(id),
+    code text
+);
+
```
alexeykazakov: Add indexes for ID and Code.

test/account.go
```diff
@@ -82,3 +104,38 @@ func CreateTestIdentityForAccountIdentity(db *gorm.DB, identity *account.Identit
...
+		log.Error(nil, map[string]interface{}{
+			"err":      err,
+			"identity": identity.User,
+		}, "unable to create user")
+		return err
+	}
+	log.Info(nil, map[string]interface{}{"user_id": identity.User.ID}, "created user")
+
+	err = models.Transactional(db, func(tx *gorm.DB) error {
+		return identityRepository.Create(context.Background(), identity)
```
alexeykazakov: Why in a separate transaction?

account/email/verification_service_blackbox_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (s *verificationServiceBlackboxTest) SetupTest() {
+	s.DBTestSuite.SetupTest()
+	s.repo = account.NewVerificationCodeRepository(s.DB)
+	s.verificationService = email.NewEmailVerificationClient(s.Application)
+}
+
+func (s *verificationServiceBlackboxTest) TestSendVerificationCodeOK() {
+	identity, err := test.CreateTestIdentity(s.DB, uuid.NewV4().String(), "kc")
+	require.Nil(s.T(), err)
```
alexeykazakov: Let's start using require.Error() and NotError() instead. See https://github.com/fabric8-services/fabric8-wit/pull/1813

account/email/verification_service_blackbox_test.go
```diff
@@ -0,0 +1,68 @@
...
+	require.Equal(s.T(), identity.ProviderType, "kc")
+
+	generatedCode := uuid.NewV4().String()
+	newVerificationCode := account.VerificationCode{
+		User: identity.User,
+		Code: generatedCode,
+	}
+
+	fmt.Println(identity.User.ID)
```
alexeykazakov: Remove println

account/email/verification_service_blackbox_test.go
```diff
@@ -0,0 +1,68 @@
...
+	err = s.Application.VerificationCodes().Create(context.Background(), &newVerificationCode)
+	codeOK, err := s.verificationService.VerifyCode(context.Background(), generatedCode)
+	require.Nil(s.T(), err)
+	require.NotNil(s.T(), codeOK)
+
+	// let's check if it's present in the db - should be deleted
+	verificationCodes, err := s.Application.VerificationCodes().LoadByCode(context.Background(), generatedCode)
+	require.Nil(s.T(), err)
+	require.Empty(s.T(), verificationCodes)
+}
```
alexeykazakov: Also test if verification of unknown code fails.

controller/users_blackbox_test.go
```diff
@@ -69,6 +71,38 @@ func (s *TestUsersSuite) SecuredServiceAccountController(identity account.Identi
...
+	}
+	//secureController, secureService := createSecureController(t, identity)
+	updateUsersPayload := createUpdateUsersPayload(&newEmail, &newFullName, &newBio, &newImageURL, &newProfileURL, &newCompany, nil, nil, contextInformation)
+	test.UpdateUsersOK(s.T(), secureService.Context, secureService, secureController, updateUsersPayload)
+
+	codes, err := s.Application.VerificationCodes().Query(account.VerificationCodeWithUser(), account.VerificationCodeFilterByUserID(user.ID))
+	require.Nil(s.T(), err)
+	require.Len(s.T(), codes, 1)
+	verificationCode := codes[0].Code
+	test.VerifyEmailUsersOK(s.T(), secureService.Context, secureService, secureController, verificationCode)
```
alexeykazakov: Check if the code was deleted from DB.

controller/users_blackbox_test.go
```diff
@@ -69,6 +71,38 @@ func (s *TestUsersSuite) SecuredServiceAccountController(identity account.Identi
...
+}
+
```
alexeykazakov: Add some very simple tests (no users, no codes) to check if it fails when we verify unknown code.

account/email/verification_service.go
```diff
@@ -0,0 +1,91 @@
...
+		err = appl.Users().Save(ctx, &user)
+		if err != nil {
+			return err
+		}
+
+		err = appl.VerificationCodes().Delete(ctx, verificationCode.ID)
+		return err
+
+	})
+	if err != nil {
```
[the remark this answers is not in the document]
sbose78: Actually, plan to add logging in a number of places where I'm checking for an error - this too is one such place.

migration/sql-files/012-add-email-verified.sql
sbose78: yeah. that would be a fair assumption to make for now.
>we can afford skip Auth-Keycloak migration at this point.

Do you mean, the part where we need to refresh the token?

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-6`

controller/users_blackbox_test.go
sbose78: thanks , 48469ff

controller/users_blackbox_test.go
sbose78: 50cd526433557dc0ac357bcb468bd2862b6ac833 thanks,done.

design/account.go
sbose78: 2754fab  thanks, done.

migration/sql-files/012-add-email-verified.sql
sbose78: thanks, 575e2bb

test/account.go
sbose78: fixed 261d298

controller/users.go
sbose78: Yeah it's a one liner, was expecting to do other things too :)  
removed e8b8863

account/verification_code.go
sbose78: 19e25e8 done.

account/verification_code.go
sbose78: ah yes, my bad! 
19e25e8

account/verification_code.go
sbose78: Right, we dont need NullUUID , 
6d6d759

account/email/verification_service_blackbox_test.go
sbose78: cbdb4ad

account/email/verification_service_blackbox_test.go
sbose78: 01322b3

account/email/verification_service_blackbox_test.go
sbose78: 01322b3 thanks, done in multiple places.

account/email/verification_service.go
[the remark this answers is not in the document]
sbose78: f3ef72f  removed it, and added logging.

account/email/verification_service.go
```diff
@@ -0,0 +1,91 @@
...
+func (c *EmailVerificationClient) SendVerificationCode(ctx context.Context, user account.User) (*account.VerificationCode, error) {
+
+	generatedCode := uuid.NewV4().String()
+	newVerificationCode := account.VerificationCode{
+		User: user,
+		Code: generatedCode,
+	}
+
+	err := application.Transactional(c.db, func(appl application.Application) error {
+		log.Info(ctx, map[string]interface{}{
```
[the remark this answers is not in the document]
sbose78: f3ef72f

sbose78:
Thanks @alexeykazakov for your detailed review - I'm waiting on this to get merged https://github.com/fabric8-services/fabric8-notification/pull/51 so that I can consume the updated API here.

migration/sql-files/012-add-email-verified.sql
alexeykazakov: No, no. We have to do the token refresh :)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-7`

design/account.go
sbose78: sure will make it `https://prod-preview.openshift.io/_gettingstarted` by default.

sbose78: @alexeykazakov there would be an error page and a success page ( as per UX's designs ) 
https://redhat.invisionapp.com/share/U2F2YVAT5

I was planning to have a UI redirect with a param `verified=true` and `verified=false` to the same URL. Based on the param the UI content in the page would be rendered.

What do you think ?
@alexeykazakov  @mindreeper2420 ( cant find Catherine here )

alexeykazakov: The same URL with different params (verified=true/false) would simplify things on the back-end side. But not sure if it's OK for the front-end side ;)
cc: @catrobson

catrobson: @alexeykazakov @sbose78 That sounds good to me, I think this should be fine for the front-end side.

sbose78: Thanks @catrobson and @alexeykazakov !

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-12`

notification/notification.go
```diff
@@ -0,0 +1,144 @@
...
+			log.Error(ctx, map[string]interface{}{
+				"status":     resp.StatusCode,
+				"message_id": msg.MessageID,
+				"type":       msg.MessageType,
+				"target_id":  msg.TargetID,
+				"custom":     msg.Custom,
+				"err":        err,
+			}, "unexpected response code")
+		}
+		defer resp.Body.Close()
```
alexeykazakov: If you use "defer rest.CloseResponse()" instead then it will also read the response body. It will help avoid FD leaks if response body is not empty.

notification/notification.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+		)
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"custom":     msg.Custom,
+				"message_id": msg.MessageID,
+				"type":       msg.MessageType,
+				"target_id":  msg.TargetID,
+				"err":        err,
+			}, "unable to send notification")
```
alexeykazakov: return

notification/notification.go
```diff
@@ -0,0 +1,144 @@
...
+		} else if resp.StatusCode >= 400 {
+			log.Error(ctx, map[string]interface{}{
+				"status":     resp.StatusCode,
+				"message_id": msg.MessageID,
+				"type":       msg.MessageType,
+				"target_id":  msg.TargetID,
+				"custom":     msg.Custom,
+				"err":        err,
```
alexeykazakov: err will be nil here

test/notification.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/fabric8-services/fabric8-auth/notification"
+)
+
+// NotificationChannel is a simple Sender impl that records the notifications for later verification
+type NotificationChannel struct {
+	Messages []notification.Message
```
alexeykazakov: Doesn't look like you use these messages in any tests?

sbose78: I'm using `test.NotificationChannel` but not the `Messages` yet - been trying to figure out some pointer indirection challenges with `Channel`.

Initial idea was(is) to inspect the `Messages` array in tests.

sbose78:
[test]

notification/notification.go
```diff
@@ -0,0 +1,144 @@
...
+		MessageID:   uuid.NewV4(),
+		MessageType: "user.email.updated",
+		TargetID:    userID,
+		Custom:      custom,
+	}
+}
+
+func setCurrentIdentity(ctx context.Context, msg *Message) {
+	currentUserIdentityID, err := login.ContextIdentity(ctx)
+	if err != nil {
```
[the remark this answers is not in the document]
sbose78: I'm reconsidering having this method since we aren't using the user info explicitly.

sbose78:
Local testing - haven't figured out a way to send out emails since the notification service on prod-preview is not exposed. 

Alternative would be to lay hands on a mailchimp account key for local testing.


**Update the user email** 

```
curl --request PATCH \
  --header 'authorization: Bearer eyJhbGciOh93vQ' \
  --data '{\n  "data": {\n    "attributes": {\n      "cluster": "cluster",\n      "email":"sbose0708@gmail.com"\n    },\n    "type": "identities"\n  }\n}'
```

This should return a usual 200 OK

Then, dig up the most recent code from the database 
`select code from verification_codes;` 

**Construct the verification url**

http://localhost:8089/api/users/verifyemail?code=e313d6d7-12b9-499d-b1e4-b958aa7a4fa4

This should redirect to: 
https://prod-preview.openshift.io/_home?verified=true&error=true

Try hitting the same url again and you would get a 
https://prod-preview.openshift.io/_home?verified=false&error=code%20with%20id%20%27e313d6d7-12b9-499d-b1e4-b958aa7a4fa4%27%20not%20found

notification/notification.go
sbose78: Thanks for pointing out - should get into the habit of using this :)  ad43a64

notification/notification.go
sbose78: ad43a64

notification/notification.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+// DevNullChannel is the default configured channel. It does nothing.
+type DevNullChannel struct{}
+
+// Send NO-OP
+func (d *DevNullChannel) Send(context.Context, Message) {}
+
+// ServiceConfiguration holds configuration options required to interact authh the fabric8-notification API
```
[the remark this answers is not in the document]
sbose78: 2e08da2

notification/notification.go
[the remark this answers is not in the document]
sbose78: ce63fdf

main.go
```diff
@@ -140,6 +141,19 @@ func main() {
 	identityRepository := account.NewIdentityRepository(db)
 	userRepository := account.NewUserRepository(db)
 
+	var notificationChannel notification.Channel = &notification.DevNullChannel{}
+	if config.GetNotificationServiceURL() != "" {
```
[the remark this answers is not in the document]
sbose78: thanks, done. 18ef67da8b8d9536168a97d2e58c66e19fcd830d

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-14`

notification/notification.go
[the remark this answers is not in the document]
sbose78: 13e5e1d  using for logging.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-15`

notification/notification.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+	return &Service{config: config}, nil
+}
+
+// Send invokes the fabric8-notification API
+func (s *Service) Send(ctx context.Context, msg Message) {
+	go func(ctx context.Context, msg Message) {
+
+		u, err := url.Parse(s.config.GetNotificationServiceURL())
```
alexeykazakov: You already parsed it once in NewServiceChannel (line 65). Why don't save it as an URL in the Service struct (instead of ServiceConfiguration) instead of re-parsing it every time we are sending a message?

sbose78: We could do that for sure - but it (kind of) deviates from how we use configuration information.
In all places, 
1. We initialize the config object in the constructor helper ( `NewService..` ).
2. We consume the content the config object in the service/struct methods.

but I'm trying to avoid deviating into initializing the config value directly inside the constructor. 
WDYT ?
Dont have too strong feelings about this :)

alexeykazakov: No, no. You still pass the configuration to the constructor. But then get the URL from it and parse it (as you already do) then save it instead of throwing it away. Then just reuse it every time you need it instead of re-parsing over and over again every time.

sbose78: Sure!

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-227-16`

Issue #61, opened 2017-08-17
Move PATCH /users to /user


alexeykazakov:
It doesn't make sense to keep PATCH under /users/:userID instead of /user

alexeykazakov:
We will need to keep PATCH /users working too until we migrate all existing clients to /user

kishansagathiya:
@alexeykazakov @sbose78 If I were to copy the code used for user update in /users/:userID  to /user, I would end up doing lot of code repetition. Instead I am thinking of creating a new package say user package where we would keep all the methods (kinda like login package) and we would be able to use them in userController and usersController both. WDYT?

alexeykazakov:
We need to move existing PATCH /users/:userID code to /user and proxy /users/:userID requests to /user.
Example of such proxy - https://github.com/fabric8-services/fabric8-wit/blob/master/controller/login.go#L53
(you will need to copy this file from WIT to Auth - https://github.com/fabric8-services/fabric8-wit/blob/master/rest/proxy/proxy.go)

PR #228, opened 2017-12-15
Move PATCH /users to /user

kishansagathiya:
Created PATCH /user (Similar to PATCH /users)
- action=Update ctrl=User route="PATCH /api/user" security=jwt
- Code for Update(ctx *app.UpdateUserContext) borrowed/copied from Update action of UsersController (https://github.com/fabric8-services/fabric8-auth/blob/master/controller/users.go#L420)

(#61)


fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-228-2`

controller/users.go
```diff
@@ -338,6 +339,7 @@ func (c *UsersController) createUserInDB(ctx *app.CreateUsersContext, identityID
 	return identity, user, nil
 }
 
+/*
```
alexeykazakov: Do not comment old code. Remove it.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-228-3`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-228-4`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-228-5`

sbose78:
Doing a `make dev` wouldn't work as the proxy returns a 502 bad gateway but when I ran a 
```
AUTH_DEVELOPER_MODE_ENABLED=true ./bin/auth
``` 

with a 

```
curl --request PATCH   --url http://localhost:8089/api/users   --header 'authorization: Bearer eFZ9dH0g'   --header 'cache-control: no-cache'   --header 'content-type: application/json'   --header 'postman-token: 60633f04-2ccb-f3d6-bdaa-71045e223c7a'   --data '{"data":{"attributes":{"contextInformation":{"boosterCatalog":null,"tenantConfig":null}},"type":"identities"}}'
```

it worked!

This means, that we need to test these changes in minishift before we merge.

@kishansagathiya  could you please setup the auth stack in minishift https://github.com/fabric8-services/fabric8-auth/tree/master/minishift  and test out your changes? We want to avoid a situation where this works locally but fails on openshift.

kishansagathiya:
The PATCH request was converting to POST  because I was using inappropriate curl command. It had nothing to do with the code. 
@alexeykazakov From What I assume (thanks to @jarifibrahim for finding this out), in goa, when we declare payload in the design, request.body becomes empty. Our proxy function makes use of only request, thus proxy does not pass data. Thus the request would fail. Today I spent quite some time trying to add body to the request, but was met with errors all the time.  Below is an example
```
	var body bytes.Buffer
	/* The line below fails for invalid memory address or nil pointer dereference */
	client.Client.Encoder.Encode(ctx.Payload, &body, "*/*")
	req, err := http.NewRequest("POST", target.String(), body)
```
The proxy function works in WIT, because they have not  have payload in the api design of update users, because of which request.Body is not emptied. The problem with this is that they have an api design without payload, but while using the api we will have to pass the payload. Because of this inconsistency we won't be able to use autogenerated tests. Writing tests for this approach would be very lengthy. They don't have tests for update users. If we decide to skip tests as well, then this method is handy and I don't see much problem. In the latest commit, users update works, but the WIT way.

A sample curl command 
```
curl -X PATCH -H "Authorization: Bearer $TKN" -H "Content-Type: application/json" "http://localhost:8089/api/users" -d '{"data":{"attributes":{"bio": "123"}, "type":"identities"}}' -v
```

alexeykazakov:
@kishansagathiya yeah.. I guess it happens because Goa reads the request stream to populate the payload in the context if it declared in the design. And once it's read it gets empty.

That's OK if we remove almost everything from the deprecated endpoint (payload, attributes, etc) in the design and move all tests to the new endpoint. Basically we can do the same we did for WIT proxy.

alexeykazakov:
Could you please fix the conflicts?

kishansagathiya:
@sbose78 I had to move lot of your code. Though I have checked twice if I have missed anything, but you should also see once.

controller/user_blackbox_test.go
```diff
@@ -390,7 +390,7 @@ func newGormTestBase(identity *account.Identity, user *account.User) *GormTestBa
 		UserRepository:     TestUserRepository{User: user}}
 }
 
-func (rest *TestUserREST) createRandomUser(fullname string) account.User {
+func (rest *TestUserREST) giveUser(fullname string) account.User {
```
alexeykazakov: Let's not rename methods/functions in this PR if don't have to. It will help us to make sure that we just are moving code from one file to another.

kishansagathiya: @alexeykazakov then we will have to create an issue and PR just to rename this.

alexeykazakov: That's OK.

sbose78:
We should test this on minishift/openshift before we merge this.

kishansagathiya:
@sbose78 Many conflicts now. Let me continue on this after che issue and other che related auth-issues are done?

sbose78:
Does the spec need it to be `.../userinfo` ?

kishansagathiya:
@sbose78 openid specs docs are using `/userinfo`, but I think specs don't put restrictions on paths. We could have kept `/gettoken` instead of `/token` and we would still be following openid-specs. This is what my understanding is. I have seen this mentioned explicitly, so It would be better to confirm with @alexeykazakov

sbose78:
@kishansagathiya let's keep it as is - I would suggest. Worst case if it doesn't work for the clients, we might have to proxy.

alexeykazakov:
Closing as outdated.

PR #236, opened 2017-12-21
Update kedge configurations based on upstream

surajssd:
This change is inspired from the configs that platform uses to deploy the auth with few changes that makes latest auth to run.

This also makes sure these are openshift deploymentconfigs and not kubernetes deployments.

Link to upstream configs: http://central.maven.org/maven2/io/fabric8/platform/packages/fabric8-system/4.0.208/fabric8-system-4.0.208-openshift.json


Makefile
```diff
@@ -321,20 +321,17 @@ dev-db-openshift: prebuild-check deps generate $(FRESH_BIN)
 
 .PHONY: dev-openshift
 dev-openshift: prebuild-check deps generate build bin/docker/fabric8-auth-linux
-	minishift start --cpus 4
+	minishift start --cpus 3
```
surajssd: For auth to run we don't need so much cpu so reduced it.

Makefile
```diff
@@ -321,20 +321,17 @@ dev-db-openshift: prebuild-check deps generate $(FRESH_BIN)
...
-	kedge apply -f minishift/kedge/db-auth.yml
-	sleep 5s
-	-eval `minishift docker-env` && docker login -u developer -p $$(oc whoami -t) $$(minishift openshift registry) && docker build -t fabric8/fabric8-auth:dev bin/docker
-	-kedge delete -f minishift/kedge/auth-local.yml
-	kedge apply -f minishift/kedge/auth-local.yml
+	-eval `minishift oc-env` && oc login -u developer -p developer && oc new-project auth-openshift
+	kedge apply -f minishift/kedge/auth-db.yml
+	-eval `minishift docker-env` && docker build -t fabric8/fabric8-auth:dev bin/docker
+	IMAGE_TAG=dev AUTH_DEVELOPER_MODE_ENABLED="'true'" kedge apply -f minishift/kedge/auth-local.yml
+	-oc rollout latest dc/auth
```
surajssd: this command will automatically pick up the latest built image, but this might fail for the first time we do it so can be ignored

Makefile
```diff
@@ -321,20 +321,17 @@ dev-db-openshift: prebuild-check deps generate $(FRESH_BIN)
...
-	AUTH_DEVELOPER_MODE_ENABLED=true \
-	kedge apply -f minishift/kedge/db-auth.yml
-	sleep 5s
-	-eval `minishift docker-env` && docker login -u developer -p $$(oc whoami -t) $$(minishift openshift registry) && docker build -t fabric8/fabric8-auth:dev bin/docker
-	-kedge delete -f minishift/kedge/auth-local.yml
-	kedge apply -f minishift/kedge/auth-local.yml
+	-eval `minishift oc-env` && oc login -u developer -p developer && oc new-project auth-openshift
+	kedge apply -f minishift/kedge/auth-db.yml
+	-eval `minishift docker-env` && docker build -t fabric8/fabric8-auth:dev bin/docker
+	IMAGE_TAG=dev AUTH_DEVELOPER_MODE_ENABLED="'true'" kedge apply -f minishift/kedge/auth-local.yml
```
surajssd: I have parameterized the configs to change the image tag and also we can switch between dev mode and disable it when needed

Makefile
```diff
@@ -321,20 +321,17 @@ dev-db-openshift: prebuild-check deps generate $(FRESH_BIN)
...
 .PHONY: dev-openshift
 dev-openshift: prebuild-check deps generate build bin/docker/fabric8-auth-linux
-	minishift start --cpus 4
+	minishift start --cpus 3
 	./minishift/check_hosts.sh
-	-eval `minishift oc-env` &&  oc login -u developer -p developer && oc new-project auth-openshift
-	AUTH_DEVELOPER_MODE_ENABLED=true \
-	kedge apply -f minishift/kedge/db-auth.yml
-	sleep 5s
-	-eval `minishift docker-env` && docker login -u developer -p $$(oc whoami -t) $$(minishift openshift registry) && docker build -t fabric8/fabric8-auth:dev bin/docker
```
surajssd: we don't need to login to internal docker registry so removed it

Makefile
```diff
@@ -321,20 +321,17 @@ dev-db-openshift: prebuild-check deps generate $(FRESH_BIN)
...
+	-eval `minishift oc-env` && oc login -u developer -p developer && oc new-project auth-openshift
+	kedge apply -f minishift/kedge/auth-db.yml
+	-eval `minishift docker-env` && docker build -t fabric8/fabric8-auth:dev bin/docker
+	IMAGE_TAG=dev AUTH_DEVELOPER_MODE_ENABLED="'true'" kedge apply -f minishift/kedge/auth-local.yml
+	-oc rollout latest dc/auth
 	
 .PHONY: clean-openshift
 clean-openshift:
 	-eval `minishift oc-env` &&  oc login -u developer -p developer
-	-kedge delete -f minishift/kedge/auth.yml -f minishift/kedge/db-auth.yml
```
surajssd: no need of delete, since deleting the entire project will delete everything inside it

minishift/kedge/auth-local.yml
```diff
@@ -1,16 +1,99 @@
...
-  - name: AUTH_POSTGRES_PORT
-    value: "5432"
+    value: [[ AUTH_DEVELOPER_MODE_ENABLED ]]
+  envFrom:
+  - configMapRef:
+      name: auth
+  - secretKeyRef:
+      name: auth
+
+  health:
```
surajssd: since liveness and readiness checks are exactly the same, so we can use the kedge's shortcut called health which populates the liveness and readiness both for us.

minishift/kedge/auth-local.yml
```diff
@@ -1,16 +1,99 @@
...
+
+routes:
+- to:
+    kind: Service
+    name: auth
```
surajssd: this will create a route for us

minishift/kedge/db-auth.yml
```diff
@@ -1,12 +0,0 @@
-name: db-auth
```
surajssd: renamed the file name to be consistent with what is used in platform

Makefile
sbryzak: How can we build an image from compiled local source in the internal docker registry if we don't connect to it?

surajssd: so when you do `eval ` on minishift docker-env you already have access to the docker environment of the minishift machine, so when you do `docker build ` it is happening inside the VM so the image is already available to the openshift cluster inside minishift so no need to login to the registry.

But if you want to use the registry just logging in is not enough because you will also have to push the image to the registry and then use it.

sbryzak: Thanks for the clarification, I thought the docker login was necessary to be able to build the new docker image inside minishift's docker registry, but if setting the env variables is sufficient then that is fine.

minishift/kedge/auth-local.yml
```diff
@@ -1,16 +1,99 @@
...
+secrets:
 - name: auth
-  type: NodePort
-  ports:
-  - port: 8089
-    nodePort: 31000
+  data:
+    AUTH_POSTGRES_USER: cG9zdGdyZXM=
+    AUTH_POSTGRES_PASSWORD: bXlzZWNyZXRwYXNzd29yZA==
+    AUTH_TOKEN_PUBLICKEY: LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlHZU1BMEdDU3FHU0liM0RRRUJBUVVBQTRHTUFEQ0JpQUtCZ0VqbFpOMVE3d2c1UU9IelhHcXlOZi9xL1ZIUTBvSjJOZGVPRHMxOHh1WkpHSkxEZVA4ZU5QY2trbGdWb1RJaFZzOTlaNXRMTTFoUCs5R213ODI1dVVXRnZsNHg1MGoyYW0yNUFIOGlpUGhreUw2aEx6dThzYnd3clU5R1RpcllqRVM5RlNRVWF6TmlWMVhYSGR5TGRvRm03dzMzZDVobFdpT1M4NmNmRnNGN0FnTUJBQUU9Ci0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQo=
```
alexeykazakov: What is this? We do not use any hardcoded public key in Auth.

minishift/kedge/auth-local.yml
```diff
@@ -1,16 +1,99 @@
...
+    AUTH_TENANT_SERVICEURL: aHR0cDovL2Y4dGVuYW50Ojgw
```
alexeykazakov: http://f8tenant:80 looks wrong to me... If we expect the tenant service to be run in minitshift too I would expect another URL for it?

minishift/kedge/auth-local.yml
```diff
@@ -1,16 +1,99 @@
...
...fSwKICAgICAgICB7CiAgICAgICAgICAgICJuYW1lIjoiZmFicmljOC10ZW5hbnQiLAogICAgICAgICAgICAiaWQiOiJjMjExZjFiZC0xN2E3LTRmOGMtOWY4MC0wOTE3ZDE2Nzg4OWQiLAogICAgICAgICAgICAic2VjcmV0cyI6WyIkMmEkMDQkeW5xTS9zeUtNWW93TUluNWN5cUh1ZXZXbmZ6SVFxdHlZNG0uNjFCMDJxbHRZNVNPeUdJT2UiLCAiJDJhJDA0JHNiQy9BZlcyYzMzaHY4b3JHQS4xRC5MWGEvLklZNzZWV2hzZnF4Q1ZocmhGa0RmTDAvWEdLIl0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICAgIm5hbWUiOiJmYWJyaWM4LWplbmtpbnMtaWRsZXIiLAogICAgICAgICAgICAiaWQiOiIzNDFjMjgzZi0wY2Q3LTQ4YTgtOTI4MS00NTgzYWNlYjM2MTciLAogICAgICAgICAgICAic2VjcmV0cyI6WyIkMmEkMDQkaGJHSEFWS29ocGVEZ0h6YWZuTHdkTzRaemhFbjl1a1ZQLzZDYU90ZjVvM0J0cC5yNnRYVEciXQogICAgICAgIH0sCiAgICAgICAgewogICAgICAgICAgICAibmFtZSI6ImZhYnJpYzgtb3NvLXByb3h5IiwKICAgICAgICAgICAgImlkIjoiOTZlZTNmOTEtZjk4MC00MjlmLWE0ZTgtMDYwYzMyNTgwMzFkIiwKICAgICAgICAgICAgInNlY3JldHMiOlsiJDJhJDA0JGhiR0hBVktvaHBlRGdIemFmbkx3ZE80WnpoRW45dWtWUC82Q2FPdGY1bzNCdHAucjZ0WFRHIl0KICAgICAgICB9LAogICAgICAgIHsKICAgICAgICAgICAgIm5hbWUiOiJvbmxpbmUtcmVnaXN0cmF0aW9uIiwKICAgICAgICAgICAgImlkIjoiZjg2N2VjNzItMzE3MS00YjhmLThlZWMtOTBhMzJlYWI2ZTBiIiwKICAgICAgICAgICAgInNlY3JldHMiOlsiaklSNUZwWXEwUVVVek1FS3FpQUlWbG9OQkNMM3Yxbk9QeHE5V20wN3ZUc0poS0dOQlJXV3Nka25LNHg0ZWwzIl0KICAgICAgICB9CiAgICBdCn0=
```
alexeykazakov: Why not re-using the default config here? If these files are not mounted then https://github.com/fabric8-services/fabric8-auth/tree/master/configuration/conf-files will be used. The default config is preferable in dev env.
The health check will fail if the files are not mounted and service is not run in Dev Mode (AUTH_DEVELOPER_MODE_ENABLED = true) though.

alexeykazakov:
make dev-openshift fails for me:
```
kedge apply -f minishift/kedge/auth-db.yml
unable to perform controller operations: unable to unmarshal data: could not unmarshal into internal struct: error unmarshaling JSON: json: cannot unmarshal number into Go struct field ServiceSpecMod.portMappings of type string
Makefile:324: recipe for target 'dev-openshift' failed
```

sbose78:
Let's not define runtime env vars for which there are default values in https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration.go#L350

unless its necessary :)

minishift/kedge/auth-local.yml
surajssd: not sure, it is used in http://central.maven.org/maven2/io/fabric8/platform/packages/fabric8-system/4.0.208/fabric8-system-4.0.208-openshift.json

this is direct conversion from configs used there! I can remove that if it is not used anymore

minishift/kedge/auth-local.yml
surajssd: what should be the URL be?

minishift/kedge/auth-local.yml
surajssd: I had issues running without these files being mounted, but let me give it a try and get back to this comment.

surajssd:
It would be really awesome to see same configs used in both(dev/prod) places. With few switches which when turned on/off it can become ready for one environment.

surajssd:
@alexeykazakov do you mind updating your kedge binary? The portMappings issue has been fixed in the newer version of kedge!

sbryzak:
@surajssd does the fixed portMappings issue mean that the service ports will be correctly freed now when doing a kedge delete?

alexeykazakov:
We can’t use the same config in prod and dev environment. This is the point of having config vs hardcoding so we can configure the service independently in different environments, right? :)
Or you meant something else?

Regarding the version of kedge... please update the documentation to reflect these requirements.

minishift/kedge/auth-local.yml
alexeykazakov: Yes, please remove this.

minishift/kedge/auth-local.yml
alexeykazakov: It's supposed to be the URL to the instance of https://github.com/fabric8-services/fabric8-tenant service. So, for now http://f8tenant:80 may be OK.

alexeykazakov:
@surajssd do you plan to update this PR to address the comments and answer the questions?

sbryzak:
@surajssd any update on this PR?

alexeykazakov:
Closing as abandoned.

fabric8cd:
@surajssd snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-236-3`

Issue #35, opened 2017-07-28
Support privacy properties


alexeykazakov:
Need to support privacy properties as shown in the following user profile mockup.
https://redhat.invisionapp.com/share/3ZBFATW28#/screens

The properties appear to be booleans for:

1. Display name in your public profile
2. Display email in your public profile
3. Display activity in your public profile
4. Display Spaces in your public profile
5. Display work items assigned in your public profile

We could temporarily store some extra properties in the user contextInformation object. However, I'm thinking we may want to make this part of the user's profile, along with properties like email, bio, etc.?

Moved from https://github.com/fabric8-services/fabric8-wit/issues/1288

mindreeper2420:
@alexeykazakov @sbose78 For the user's email address, do we want it to be displayed by default, or hidden by default?

alexeykazakov:
Good question... Hidden/private by default I guess?

mindreeper2420:
I've created a wireframe that both matches the current User Profile layout and includes the ability to select Private or Public for the Email address. We are still reviewing it in UX, but feel free to add any comments that you have as well: https://redhat.invisionapp.com/share/F4F0D8FZA.

alexeykazakov:
Thanks @mindreeper2420 ! Looks good.

PR #237, opened 2017-12-22
Allow email privacy to be configured

sbose78:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/35

The objective of this PR is to provide a way to make email private using the `PATCH /api/users`

Make your email private : 
```
curl --request POST \
  --header 'authorization: Bearer eyJhbGciOiJSUzI1NiIajg' \
  --data '{"data": { "attributes": {"email_private":true},"type": "identities"}}'
```

The email should not be returned ( will be an empty string ) when a call similar to 
```
https://auth.openshift.io/api/users?filter[username]=shbose
```
is made.

However, a call to to 
```
curl --request GET \
  --url http://localhost:8089/api/user \
  --header 'authorization: Bearer eyJhbGciOiJSUzI1NiIajg' \
```

would return the email address irrespective of the `email_private` value.

### DIFF
--- a/account/user.go
+++ b/account/user.go
@@ -25,7 +25,7 @@ type User struct {
 	gormsupport.Lifecycle
 	ID                 uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
 	Email              string    `sql:"unique_index"`
-	EmailHidden        bool
+	EmailPrivate       bool
 	FullName           string             // The fullname of the User
 	ImageURL           string             // The image URL for the User
 	Bio                string             // The bio of the User
--- a/account/user.go
+++ b/account/user.go
@@ -23,8 +23,9 @@ import (
 // User describes a User account. A few identities can be assosiated with one user account
 type User struct {
 	gormsupport.Lifecycle
-	ID                 uuid.UUID          `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
-	Email              string             `sql:"unique_index"`                                            // This is the unique email field
+	ID                 uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+	Email              string    `sql:"unique_index"`
+	EmailHidden        bool
 	FullName           string             // The fullname of the User
 	ImageURL           string             // The image URL for the User
 	Bio                string             // The bio of the User
--- a/controller/search_user_blackbox_test.go
+++ b/controller/search_user_blackbox_test.go
@@ -151,7 +151,7 @@ func (s *TestSearchUserSearch) TestEmailPrivateSearchOK() {
 		Cluster:      "default Cluster",
 	}
 
-	_, err := testsupport.CreateTestUser(s.DB, user)
+	_, err := testsupport.CreateTestUser(s.DB, &user)
 	require.Nil(s.T(), err)
 
 	offset := "0"
@@ -172,7 +172,7 @@ func (s *TestSearchUserSearch) TestEmailNotPrivateSearchOK() {
 		Email:        uuid.NewV4().String(),
 		Cluster:      "default Cluster",
 	}
-	_, err := testsupport.CreateTestUser(s.DB, user)
+	_, err := testsupport.CreateTestUser(s.DB, &user)
 	require.Nil(s.T(), err)
 
 	offset := "0"
--- a/controller/search_user_blackbox_test.go
+++ b/controller/search_user_blackbox_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/fabric8-services/fabric8-auth/gormapplication"
 	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
 	"github.com/fabric8-services/fabric8-auth/resource"
+	testsupport "github.com/fabric8-services/fabric8-auth/test"
 	"github.com/goadesign/goa"
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/require"
@@ -142,26 +143,16 @@ func (s *TestSearchUserSearch) createTestData() []account.Identity {
 
 func (s *TestSearchUserSearch) TestEmailPrivateSearchOK() {
 	randomName := uuid.NewV4().String()
-	application.Transactional(s.Application, func(app application.Application) error {
-		user := account.User{
-			EmailPrivate: true,
-			FullName:     randomName,
-			ImageURL:     "http://example.org/" + randomName + ".png",
-			Email:        uuid.NewV4().String(),
-			Cluster:      "default Cluster",
-		}
-		err := app.Users().Create(context.Background(), &user)
-		require.Nil(s.T(), err)
+	user := account.User{
+		EmailPrivate: true,
+		FullName:     randomName,
+		ImageURL:     "http://example.org/" + randomName + ".png",
+		Email:        uuid.NewV4().String(),
+		Cluster:      "default Cluster",
+	}
 
-		ident := account.Identity{
-			User:         user,
-			Username:     uuid.NewV4().String(),
-			ProviderType: "kc",
-		}
-		err = app.Identities().Create(context.Background(), &ident)
-		require.Nil(s.T(), err)
-		return err
-	})
+	_, err := testsupport.CreateTestUser(s.DB, user)
+	require.Nil(s.T(), err)
 
 	offset := "0"
 	pageLimit := 1
@@ -174,26 +165,15 @@ func (s *TestSearchUserSearch) TestEmailPrivateSearchOK() {
 
 func (s *TestSearchUserSearch) TestEmailNotPrivateSearchOK() {
 	randomName := uuid.NewV4().String()
-	application.Transactional(s.Application, func(app application.Application) error {
-		user := account.User{
-			EmailPrivate: false,
-			FullName:     randomName,
-			ImageURL:     "http://example.org/" + randomName + ".png",
-			Email:        uuid.NewV4().String(),
-			Cluster:      "default Cluster",
-		}
-		err := app.Users().Create(context.Background(), &user)
-		require.Nil(s.T(), err)
-
-		ident := account.Identity{
-			User:         user,
-			Username:     uuid.NewV4().String(),
-			ProviderType: "kc",
-		}
-		err = app.Identities().Create(context.Background(), &ident)
-		require.Nil(s.T(), err)
-		return err
-	})
+	user := account.User{
+		EmailPrivate: false,
+		FullName:     randomName,
+		ImageURL:     "http://example.org/" + randomName + ".png",
+		Email:        uuid.NewV4().String(),
+		Cluster:      "default Cluster",
+	}
+	_, err := testsupport.CreateTestUser(s.DB, user)
+	require.Nil(s.T(), err)
 
 	offset := "0"
 	pageLimit := 1
--- a/controller/search_user_blackbox_test.go
+++ b/controller/search_user_blackbox_test.go
@@ -140,6 +140,70 @@ func (s *TestSearchUserSearch) createTestData() []account.Identity {
 	return idents
 }
 
+func (s *TestSearchUserSearch) TestEmailPrivateSearchOK() {
+	randomName := uuid.NewV4().String()
+	application.Transactional(s.Application, func(app application.Application) error {
+		user := account.User{
+			EmailPrivate: true,
+			FullName:     randomName,
+			ImageURL:     "http://example.org/" + randomName + ".png",
+			Email:        uuid.NewV4().String(),
+			Cluster:      "default Cluster",
+		}
+		err := app.Users().Create(context.Background(), &user)
+		require.Nil(s.T(), err)
+
+		ident := account.Identity{
+			User:         user,
+			Username:     uuid.NewV4().String(),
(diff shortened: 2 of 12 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-1`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-2`

migration/sql-files/012-hide-email.sql
```diff
@@ -0,0 +1,3 @@
+-- Store identity provider username for the external token
+ALTER TABLE users ADD COLUMN email_hidden boolean;
```
alexeykazakov: email_private would be better. Hidden/shown email is more UI/UX thing. Private/public is a model thing.

controller/users.go
```diff
@@ -82,7 +82,7 @@ func (c *UsersController) Show(ctx *app.ShowUsersContext) error {
 			}
 		}
 		return ctx.ConditionalRequest(*user, c.config.GetCacheControlUser, func() error {
-			return ctx.OK(ConvertToAppUser(ctx.RequestData, user, identity))
+			return ctx.OK(ConvertToAppUser(ctx.RequestData, user, identity, true))
```
alexeykazakov: should be false, right?

controller/users.go
```diff
@@ -863,7 +868,7 @@ func loadKeyCloakIdentity(appl application.Application, user account.User) (*acc
 }
 
 // ConvertToAppUser converts a complete Identity object into REST representation
-func ConvertToAppUser(request *goa.RequestData, user *account.User, identity *account.Identity) *app.User {
+func ConvertToAppUser(request *goa.RequestData, user *account.User, identity *account.Identity, isAuthenticated bool) *app.User {
```
alexeykazakov: Add the attribute description to the function comment what isAuthenticated means please.

design/account.go
```diff
@@ -203,6 +203,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("username", d.String, "The username")
 	a.Attribute("registrationCompleted", d.Boolean, "Whether the registration has been completed")
 	a.Attribute("email", d.String, "The email")
+	a.Attribute("email_hidden", d.Boolean, "Whether the email address would be visible to other users.")
```
alexeykazakov: email_private

controller/users.go
sbose78: nope, should be a `true` since this is the `GET /api/user` with `Authorization..`

sbose78: My bad, this is the `/api/users/<ID>` .

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-3`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-4`

controller/users.go
sbose78: thanks 287005aa5744cef0eeea2c44c3c42b6f462d2d97

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-5`

design/account.go
```diff
@@ -203,6 +203,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("username", d.String, "The username")
 	a.Attribute("registrationCompleted", d.Boolean, "Whether the registration has been completed")
 	a.Attribute("email", d.String, "The email")
+	a.Attribute("email_private", d.Boolean, "Whether the email address would be visible to other users.")
```
kishansagathiya: email_private=true means not visible,
email_private=false means visible, right?
With that logic the proper description should be "Whether the email address wouldn't be visible to other users" or  "Whether the email address would be private"

design/account.go
```diff
@@ -221,6 +222,7 @@ var updateUserDataAttributes = a.Type("UpdateIdentityDataAttributes", func() {
 	a.Attribute("email", d.String, "The email")
 	a.Attribute("bio", d.String, "The bio")
 	a.Attribute("url", d.String, "The url")
+	a.Attribute("email_private", d.Boolean, "Whether the email address would be visible to other users.")
```
kishansagathiya: Same as earlier,
"Whether the email address wouldn't be visible to other users" or "Whether the email address would be private"

account/user.go
```diff
@@ -23,8 +23,9 @@ import (
 // User describes a User account. A few identities can be assosiated with one user account
 type User struct {
 	gormsupport.Lifecycle
-	ID                 uuid.UUID          `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
-	Email              string             `sql:"unique_index"`                                            // This is the unique email field
+	ID                 uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"` // This is the ID PK field
+	Email              string    `sql:"unique_index"`
+	EmailPrivate       bool
```
kishansagathiya: Since it is a boolean instead of EmailPrivate, IsEmailPrivate would make more sense. But it might involve quite a bit of refactoring, so you can ignore this.

sbose78:
@kishansagathiya Addressed your comments and updated description too - if you could test and give me a Green signal, I will proceed with merging this :)

account/user.go
sbose78: I agree with you - but in most places we aren't using the term `is..`  - hence will keep it this way for consistency. :)

alexeykazakov: I don't think we should use "Is" in bool field names at all.

controller/users_blackbox_test.go
```diff
@@ -804,6 +804,53 @@ func (s *TestUsersSuite) TestListUsersByEmailOKEmptyResult() {
...
+	require.Equal(s.T(), user1.Email, *updateResult.Data.Attributes.Email)
+
+	// But when you try to access the same with an API which doesn't respect auth,
+	// it wouldn't be visible.
+	_, result = test.ListUsersOK(s.T(), nil, nil, s.controller, &email, nil, nil, nil)
+	returnedUserResult := result.Data[0]
+	require.Equal(s.T(), "", *returnedUserResult.Attributes.Email)
+
+	// even though the email_private=true,
+	// the email address is visible to the user if her user token is passed.
```
alexeykazakov: The comment and test do not match. This is not what you are testing here. You are testing that api/users/:identityID doesn't return private emails. We should keep this test but fix the comment and add another test to the user controller test to actually test what your describe in the comment :)

controller/users_blackbox_test.go
```diff
@@ -804,6 +804,53 @@ func (s *TestUsersSuite) TestListUsersByEmailOKEmptyResult() {
...
+
+	// given user1
+	user1 := s.createRandomUser("TestListUsersOK1")
+	identity := s.createRandomIdentity(user1, account.KeycloakIDP)
+
+	// when
+	email := user1.Email
+
+	// by default, email is public.
+	_, result := test.ListUsersOK(s.T(), nil, nil, s.controller, &email, nil, nil, nil)
```
alexeykazakov: Also check if test.ShowUsersOK() returns the public email.

controller/users_blackbox_test.go
```diff
@@ -822,12 +869,13 @@ func (s *TestUsersSuite) TestListUsersByEmailNotModifiedUsingIfNoneMatchHeader()
 
 func (s *TestUsersSuite) createRandomUser(fullname string) account.User {
```
alexeykazakov: Maybe we can replace this function by test.CreateTestUser() too?

migration/sql-files/012-hide-email.sql
```diff
@@ -0,0 +1,3 @@
+-- Store identity provider username for the external token
+ALTER TABLE users ADD COLUMN email_private boolean;
+UPDATE USERS SET email_private = false;
```
alexeykazakov: Maybe set it to NOT_NULL after updating the table?

controller/users_blackbox_test.go
sbose78: For consistency's sake, I will keep it as it is - because if change here, it *should* be changed in all other places in this file. 
In fact in this PR, https://github.com/fabric8-services/fabric8-auth/pull/227/files#diff-cc740fdbd6f22db6da588050e1b7d5a4R78
I have included a util method there too :)

controller/users_blackbox_test.go
sbose78: Agreed, looks like the /api/user and the /api/users confused me.
Thanks!

6583a918d37a389ab2aba36e546e01c89965f419

controller/search_user_blackbox_test.go
```diff
@@ -140,6 +140,70 @@ func (s *TestSearchUserSearch) createTestData() []account.Identity {
...
+	_, results := test.UsersSearchOK(s.T(), context.Background(), s.svc, s.controller, &pageLimit, &offset, randomName)
+
+	for _, result := range results.Data {
+		require.Equal(s.T(), "", *result.Attributes.Email)
+	}
+}
+
+func (s *TestSearchUserSearch) TestEmailNotPrivateSearchOK() {
+	randomName := uuid.NewV4().String()
+	application.Transactional(s.Application, func(app application.Application) error {
```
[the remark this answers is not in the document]
sbose78: 2f43cfbac1c8050cc6aa6952a26b04d62d7aab9a

controller/search_user_blackbox_test.go
```diff
@@ -140,6 +140,70 @@ func (s *TestSearchUserSearch) createTestData() []account.Identity {
 	return idents
 }
 
+func (s *TestSearchUserSearch) TestEmailPrivateSearchOK() {
+	randomName := uuid.NewV4().String()
+	application.Transactional(s.Application, func(app application.Application) error {
```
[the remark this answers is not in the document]
sbose78: 2f43cfbac1c8050cc6aa6952a26b04d62d7aab9a

migration/sql-files/012-hide-email.sql
sbose78: 7c096653afb69415af9e1ad559255cddbfba6e04

controller/users_blackbox_test.go
sbose78: 6583a918d37a389ab2aba36e546e01c89965f419

sbose78:
All review comments taken care of. Thanks @alexeykazakov !

controller/users_blackbox_test.go
alexeykazakov: So? Is this a huge refactoring? :) I mean we should try to keep our tests clean. And try to improve our tests when it makes sense as well as we improve our main code. And if you already changing these tests it's fine to do some moderate test refactoring of the related code here. But it's also OK to do the test refactoring in the separate PR but I don't want us to create report more technical debt issues and forget about them ;)

sbose78: Cool, I will make refactor in this PR :)

alexeykazakov:
This PR will break email notifications for users with private email.
What if we check if there is a service account token passed to /api/users and /api/space/:spaceID/collaborators endpoints and then include private emails to the result? WDYT? Then notification service could use it's service account to get users.

alexeykazakov:
We could address the email notification for private emails in a separate issue/PR though.

sbose78:
Agreed, was wondering the same. 
Here's what we could do

1. When notification service comes up, call Auth service API to request for service account token, and keep it in memory.

2. Subsequently use the token from configuration memory object.

3. On the auth side, check for service account token, and based on that show the email .

How does this sound to you?

alexeykazakov:
> How does this sound to you?

Sounds good to me.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-237-9`

controller/users_blackbox_test.go
sbose78: done - https://github.com/fabric8-services/fabric8-auth/pull/237/commits/f826965bb991df019f209ecb399ab385edbfab4e

controller/users_blackbox_test.go
```diff
@@ -72,7 +72,9 @@ func (s *TestUsersSuite) SecuredServiceAccountController(identity account.Identi
 func (s *TestUsersSuite) TestUpdateUserOK() {
 	// given
 	user := s.createRandomUser("TestUpdateUserOK")
-	identity := s.createRandomIdentity(user, account.KeycloakIDP)
+	identity, err := testsupport.CreateTestUser(s.DB, &user)
```
kishansagathiya: Isn't this creating user twice?

sbose78: `createRandomUser` only creates the user object ( but not in db ).
`CreateTestUser` creates the user in the db. 

( we should refactor the method name maybe )

kishansagathiya: Cool , i am gonna rename this to `giveUser`. There are other files which has the same method, but that saves user in the database as well

PR #238, opened 2017-12-22
Manage the feature group a user belongs to (#1717)

xcoulon:
Add a field in the `user` struct and a matching column
in the `users` table.
Support set/unset(using empty ("") value.

Also, refactor Users Controller update tests for clarity

Fixes openshiftio/openshift.io#1717

* Please describe what your change is about.
* What issues does it solve? (Use [autoreferencing for this](https://help.github.com/articles/basic-writing-and-formatting-syntax/#referencing-issues-and-pull-requests))
* If you are still working on the change please prefix the pull request title with "WIP"


fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-238-1`

alexeykazakov:
What about setting groups which should not be manageable by the user? Or at least not by any user. Like “internal”?

xcoulon:
As far as I know, an internal user may opt-in for the "internal" level of features, so this remains something she/he can manage as well.

migration/sql-files/012-add-user-feature-group.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN feature_level TEXT;
```
sbose78: Would we want to make it `not null` and also add a default value to this column for all existing rows?

xcoulon: no, I don't think so: it's like a program that users can opt-in/opt-out, but it's not mandatory, so I'd rather keep the column `nullable` for now, if it's ok for you.

sbose78: Isn't there a default group where all of us would be in , after we introduce this?

xcoulon: weel, in that case, that would be a group in which we only use "released" features. So that could be the default level, if you prefer it that way.

xcoulon:
@alexeykazakov @sbose78: as discussed on MM, I updated the PR with the following change:
add a validation against the email address when the level is internal only, and accept any other value as-is, so we will not depend on a redeployment of auth if an extra level is needed in the future. If a value is unknown/invalid, then the user won't have any toggles feature, that's all. Not a big damage for him/her, AFAICT (but at least auth is a bit permissive)

About @sbose78's concern about the default value, this PR keeps the field as a `*string` so it's `NULLABLE` in the database (if `NULL`, then the user has not opted for any toggled feature)

controller/users.go
```diff
@@ -23,7 +23,7 @@ import (
 	goajwt "github.com/goadesign/goa/middleware/security/jwt"
 	"github.com/jinzhu/gorm"
 	errs "github.com/pkg/errors"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
```
alexeykazakov: Not sure why you need this ( uuid "github.com/satori/go.uuid" == "github.com/satori/go.uuid") but OK :)

controller/users.go
```diff
@@ -580,6 +580,33 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
...
+			log.Debug(ctx, map[string]interface{}{"current_feature_level": currentFeatureLevel, "new_feature_level": newFeatureLevel}, "updating feature level")
+		}
+		if updatedFeatureLevel != nil && (user.FeatureLevel == nil || *updatedFeatureLevel != *user.FeatureLevel) {
+			// handle the case where the value needs to be reset, when the new value is "" (empty string) or "released"
+			if *updatedFeatureLevel == "" || *updatedFeatureLevel == "released" {
+				log.Debug(ctx, map[string]interface{}{"user_id": user.ID}, "resetting feature level")
+				user.FeatureLevel = nil
+			} else {
+				// if the level is 'internal', we need to check against the email address to verify that the user is a Red Hat employee
+				if *updatedFeatureLevel == "internal" && !strings.HasSuffix(user.Email, "@redhat.com") {
```
alexeykazakov: 1. We also need to check if email is verified. See https://github.com/fabric8-services/fabric8-auth/pull/227/files#diff-c5cd9222ed04372b0880450d7dc3e71bR35 (that PR is not merged yet).

2. I really don't like @redhat.com to be hardcoded here... Let's create a configuration for internal email suffix and set it to "@redhat.com" by default so we don't have to configure it for OSIO. But at least it will be possible to use something else in a custom configuration.

controller/users.go
```diff
@@ -580,6 +580,33 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
...
+					log.Error(ctx, map[string]interface{}{"user_id": user.ID}, "user is not a Red Hat employee")
+					return errors.NewBadParameterError("feature_level", updatedFeatureLevel)
```
alexeykazakov: Or rather 401 or 403?

design/account.go
```diff
@@ -208,6 +208,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("company", d.String, "The company")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
 	a.Attribute("cluster", d.String, "The OpenShift API URL of the cluster where the user is provisioned to")
+	a.Attribute("feature-level", d.String, "The level of features that the user wants to use (for unreleased features)")
```
alexeykazakov: featureLevel for consistency sake?

design/account.go
```diff
@@ -222,6 +223,7 @@ var updateUserDataAttributes = a.Type("UpdateIdentityDataAttributes", func() {
 	a.Attribute("bio", d.String, "The bio")
 	a.Attribute("url", d.String, "The url")
 	a.Attribute("company", d.String, "The company")
+	a.Attribute("feature-level", d.String, "The level of features that the user wants to use (for unreleased features)")
```
alexeykazakov: featureLevel?

controller/users.go
xcoulon: yeah, let me reverse it, actually.

controller/users.go
xcoulon: yes, I intend to use the `EmailVerified` field once it's in place. Let me put a `TODO` here for now. And yes, let me use a config variable to externalize the email address domain, that will be cleaner, indeed.

controller/users.go
xcoulon: Right: `400` means that the request syntax is invalid and `401` means that the request did not contain the required credentials, which are not the cases here. 

`403` means that server understood the request but refuses to process it. That sounds like a better status, indeed.

Let me fix that

design/account.go
xcoulon: sure, let me fix that, too.

design/account.go
xcoulon: yes ;)

controller/users.go
alexeykazakov: Something wrong with your VSCode then ;) I don't have this problem in mine. But that's a minor thing. We can keep it.

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-238-3`

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-238-4`

controller/users_blackbox_test.go
```diff
@@ -56,601 +55,769 @@ func (s *TestUsersSuite) SetupSuite() {
...
+			updateUsersPayload := newUpdateUsersPayload(
+				WithUpdatedEmail(newEmail),
+				WithUpdatedFullName(newFullName),
+				WithUpdatedBio(newBio),
+				WithUpdatedImageURL(newImageURL),
+				WithUpdatedURL(newProfileURL))
+			test.UpdateUsersOK(t, secureService.Context, secureService, secureController, updateUsersPayload)
+		})
+
+		t.Run("patch content info", func(t *testing.T) {
```
sbose78: typo, use `context` please

controller/users.go
```diff
@@ -596,6 +597,35 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
 			}
 		}
 
+		updatedFeatureLevel := ctx.Payload.Data.Attributes.FeatureLevel
```
sbose78: There's a lot of feature-toggle specific logic here - wondering if we should move it in a separate function to avoid the already huge controller ( blame me ) from grown bigger :)

design/account.go
```diff
@@ -222,6 +222,7 @@ var userDataAttributes = a.Type("UserDataAttributes", func() {
 	a.Attribute("company", d.String, "The company")
 	a.Attribute("providerType", d.String, "The IDP provided this identity")
 	a.Attribute("cluster", d.String, "The OpenShift API URL of the cluster where the user is provisioned to")
+	a.Attribute("featureLevel", d.String, "The level of features that the user wants to use (for unreleased features)")
```
sbose78: any reason for not using an `enum` here? Maybe to leave things flexible?

sbose78: to be able to unset it maybe :)

xcoulon: yes, I think we should remain flexible here, indeed.

controller/users_blackbox_test.go
xcoulon: fixed, thanks for spotting this typo!

controller/users.go
xcoulon: ok, I moved it in a separate, dedicated method of the controller: https://github.com/fabric8-services/fabric8-auth/pull/238/commits/fbeb03643b66158c8e8ace5d14d4dfb065c58c5a

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-238-5`

design/account.go
alexeykazakov: Yeah... Enum was my first thought here too. But let's keep it as string because Auth should not really care what feature levels we use.

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-238-6`

xcoulon:
thanks for the review, @sbose78 and @alexeykazakov!

PR #239, opened 2017-12-29
Changed state param from UUID to string

### LINKED ISSUE #234: api/authorize should not require state param to be UUID
Currently api/authorize requires state param to be UUID. We should not limit it to UUID. Please check the OAuth2 spec to see if there is any requirements to the state or it can be any string.

This is a follow up issue on https://github.com/fabric8-services/fabric8-auth/pull/210#issuecomment-352968458

kishansagathiya:
Fixes #234
/api/authorize still returns 307 Temporary Redirect on success. I think we are yet to make a call on whether to change it to 302 or continue with 307.


auth/state.go
```diff
@@ -18,7 +19,7 @@ const (
 // OauthStateReference represents a oauth state reference
 type OauthStateReference struct {
 	gormsupport.Lifecycle
-	ID       uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key"`
+	ID       string `sql:"type:uuid default CAST(uuid_generate_v4() AS varchar(50))" gorm:"primary_key"`
```
kishansagathiya: Though this does not create any problem, but I just need to make sure that this is what we want or this is the right way.

kishansagathiya:
@alexeykazakov UUID is unique, but so is the state. Why do we need both?

migration/sql-files/002-oauth-states.sql
```diff
@@ -3,6 +3,6 @@ CREATE TABLE oauth_state_references (
     created_at timestamp with time zone,
     updated_at timestamp with time zone,
     deleted_at timestamp with time zone,
-    id uuid primary key DEFAULT uuid_generate_v4() NOT NULL,
+    id varchar primary key DEFAULT CAST(uuid_generate_v4() AS varchar(50)) NOT NULL,
```
alexeykazakov: Don't change existing .sql migration files. Each .sql file represents particular DB version. Add a new .sql to migrate DB to a new version.

alexeykazakov:
@kishansagathiya If we keep ID as UUID we don't have to migrate IDs of the existing records in DB. And it will give us better flexibility if we separate ID and State in case we change the format of State later again in the future.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-239-2`

migration/sql-files/012-add-state-to-auth-state-reference.sql
```diff
@@ -0,0 +1,2 @@
+-- Alter Oauth state reference table to add state used in oauth workflow
+ALTER TABLE oauth_state_references ADD COLUMN state TEXT unique;
```
sbryzak: To enforce a not null, non-zero-length string, I believe this is the correct DDL:

ALTER TABLE oauth_state_references ADD COLUMN state TEXT unique NOT NULL CHECK (oauth_state_references <> '')

alexeykazakov: I will let @sbryzak to help you with correct SQL statement here.
We need to make sure:
1. State can be not empty unique string (in case it's not optimal to check if the string is empty on the DB level we could check it in our Create() method in the repo.
2. We have an unique index for the state column which is used when we look up the state.

token/oauth/oauth2.go
```diff
@@ -122,21 +122,14 @@ func SaveReferrer(ctx context.Context, db application.DB, state uuid.UUID, refer
...
-		ref, err := appl.OauthStates().Load(ctx, stateID)
+
+	err := application.Transactional(db, func(appl application.Application) error {
+		ref, err := appl.OauthStates().Load(ctx, state)
 		if err != nil {
 			return err
 		}
 		referrer = ref.Referrer
-		err = appl.OauthStates().Delete(ctx, stateID)
+		err = appl.OauthStates().Delete(ctx, state)
```
alexeykazakov: You don't have to change Delete(stateID) to Delete(state). You just loaded the reference and you have its ID (ref.ID). Then you won't need to load the reference in the Delete() again.

migration/sql-files/012-add-state-to-auth-state-reference.sql
sbryzak: After testing against my local postgres (v9.6.3) db, I can confirm the above DDL works.  Creating a unique column causes the db to create a corresponding index to enforce the unique constraint:

```
postgres=# ALTER TABLE oauth_state_references ADD COLUMN state TEXT unique NOT NULL CHECK (state <> '');
ALTER TABLE
postgres=# \d oauth_state_references
                       Table "public.oauth_state_references"
   Column   |           Type           | Collation | Nullable |      Default
------------+--------------------------+-----------+----------+--------------------
 created_at | timestamp with time zone |           |          | 
 updated_at | timestamp with time zone |           |          | 
 deleted_at | timestamp with time zone |           |          | 
 id         | uuid                     |           | not null | uuid_generate_v4()
 referrer   | text                     |           | not null | 
 state      | text                     |           | not null | 
Indexes:
    "oauth_state_references_pkey" PRIMARY KEY, btree (id)
    "oauth_state_references_state_key" UNIQUE CONSTRAINT, btree (state)
Check constraints:
    "oauth_state_references_state_check" CHECK (state <> ''::text)
postgres=#

**Comment by fabric8cd** [comment]
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-239-3`

File: migration/sql-files/013-add-state-to-auth-state-reference.sql [hunk 5]
```diff
@@ -1,2 +1,4 @@
 -- Alter Oauth state reference table to add state used in oauth workflow
-ALTER TABLE oauth_state_references ADD COLUMN state TEXT unique;
\ No newline at end of file
+ALTER TABLE oauth_state_references ADD COLUMN state TEXT unique;
+UPDATE oauth_state_references SET state = ID;
+ALTER TABLE oauth_state_references ALTER state SET NOT NULL;
```
alexeykazakov: Use `ALTER TABLE oauth_state_references ALTER state SET NOT NULL CHECK (state <> '')` to make sure state is not an empty string.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-239-4`

alexeykazakov:
@kishansagathiya any updates on this? We can't really switch Che to Auth without fixing this issue.

kishansagathiya:
@alexeykazakov This PR has already been updated with https://github.com/fabric8-services/fabric8-auth/pull/239/commits/11c3bca9c8c29a3ca700ba1e85343321735199d5 . Though, since it's been few days there are merge conflicts due to latest commits. let me resolve that.

kishansagathiya:
@alexeykazakov @sbose78 Conflicts resolved

token/oauth/oauth2.go
sbose78: @kishansagathiya can you take care of the above review comment?

kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/239/files#diff-12970846c32607659122a58e6cdb08c8R132
done here

kishansagathiya: @sbose78 I think you were right about this.

kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/239/commits/56e8c2a968adcf3ed11bbbe754cf8c5b9dddc5f0

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-239-7`

sbose78:
[test]

auth/state.go
```diff
@@ -61,27 +66,34 @@ type GormOauthStateReferenceRepository struct {
...
 		log.Error(ctx, map[string]interface{}{
-			"oauth_state_reference_id": ID.String(),
+			"oauth_state_reference_state": state,
 		}, "unable to delete the oauth state reference")
 		return errors.NewInternalError(ctx, err)
 	}
 	if tx.RowsAffected == 0 {
 		log.Error(ctx, map[string]interface{}{
-			"oauth state reference": ID.String(),
+			"oauth state reference": state,
```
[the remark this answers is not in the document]
sbose78: @kishansagathiya could you take care of this review comment ?
for the key in the error log message - Use smaller case letters with words joined by "_" , example, "state_id" instead of "State ID",

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-239-8`

PR #244, opened 2018-01-05
Fix health prob

alexeykazakov:
Readiness prob is broken after merging https://github.com/fabric8-services/fabric8-auth/pull/227
We need to add the tenant URL in Config Map first then we can check if it's empty in our readiness prob.

Removing it from the prob until we fix our Config Map.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-244-1`

PR #245, opened 2018-01-05
update template for notification service url

sbose78:
The corresponding configuration needs to be added to the service. 
Note: notification service url is not accessible outside openshift.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-245-1`

PR #251, opened 2018-01-09
Log configuration error in status probe

sbose78:
When /api/status is hit, if there is an error in configuration that causes a 503 - log the error as well.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-251-1`

PR #254, opened 2018-01-10
send the identity id instead of user id

sbose78:
because the notification service uses the `GET /api/users/<identity-id>` to look up user details.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-254-1`

PR #255, opened 2018-01-10
Add verify email url to openshift template

alexeykazakov:
We need this in our template so we can configure prod.
Do not merge until we have this CM in preview and prod!
  
fixes https://github.com/fabric8-services/fabric8-auth/issues/257


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-255-1`

sbose78:
tested in prod. Works.

PR #261, opened 2018-01-11
update keycloak on email verification

sbose78:
Upon email verification, I'm using the same API we used to create/update user , to update the user.
Not using the user profile Update API because that needs a user token, and in the verification controller, the user token is absent.

### DIFF
--- a/controller/users.go
+++ b/controller/users.go
@@ -846,15 +846,15 @@ func (c *UsersController) List(ctx *app.ListUsersContext) error {
 func (c *UsersController) VerifyEmail(ctx *app.VerifyEmailUsersContext) error {
 	verifiedCode, err := c.EmailVerificationService.VerifyCode(ctx, ctx.Code)
 	var errResponse string
-	isVerified := "true"
+	isVerified := true
 	if err != nil {
 		errResponse = err.Error()
-		isVerified = "false"
+		isVerified = false
 	} else if verifiedCode == nil {
 		errResponse = "unable to verify code"
-		isVerified = "false"
+		isVerified = false
 	}
-	redirectURL, err := rest.AddParam(c.config.GetEmailVerifiedRedirectURL(), "verified", string(isVerified))
+	redirectURL, err := rest.AddParam(c.config.GetEmailVerifiedRedirectURL(), "verified", fmt.Sprint(isVerified))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-261-1`

controller/users.go
```diff
@@ -864,6 +864,54 @@ func (c *UsersController) VerifyEmail(ctx *app.VerifyEmailUsersContext) error {
...
+	if isVerified {
+		verfiedUser := verifiedCode.User
+		tokenEndpoint, err := c.config.GetKeycloakEndpointToken(ctx.RequestData)
+		if err != nil {
+			return errors.NewInternalError(ctx, err)
+		}
+		protectedAccessToken, err := auth.GetProtectedAPIToken(ctx, tokenEndpoint, c.config.GetKeycloakClientID(), c.config.GetKeycloakSecret())
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"keycloak_client_id": c.config.GetKeycloakClientID(),
```
alexeykazakov: `"error": err,`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-261-2`

controller/users.go
```diff
@@ -864,6 +864,54 @@ func (c *UsersController) VerifyEmail(ctx *app.VerifyEmailUsersContext) error {
...
+
+		if protectedAccessToken != "" {
+			// try hitting the admin user endpoint only if getting a PAT
+			// was successful.
+
+			usersEndpoint, err := c.config.GetKeycloakEndpointUsers(ctx.RequestData)
+
+			identity, err := loadKeyCloakIdentity(c.db, verfiedUser)
+			if err != nil {
+				return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
```
alexeykazakov: Log the error but don't return the error. Maybe would be easier to control errors if you move this code which updates KC in a separate method?

sbose78: I logged the keycloak errors without returning it - but this error is a db-related error and I thought it would be serious enough to be reported. What do you think? :)

sbose78:
We aren't calling keycloak in the controller tests, all keycloak go services are mocked up.

controller/users.go
alexeykazakov: Oh, it’s about loading identity with KC type from our DB. Not about loading it from KC :)
My fault. Let’s return this error then.

sbose78:
How to test : 


```
curl --request PATCH \
  --header 'authorization: Bearer eyJhbGciOh93vQ' \
  --data '{"data": {"attributes": { "email":"sbose0708@gmail.com" }, "type": "identities"}}'
```

This should return a usual 200 OK

Then, dig up the most recent code from the database 
`select code from verification_codes;` 

**Construct the verification url**

http://localhost:8089/api/users/verifyemail?code=e313d6d7-12b9-499d-b1e4-b958aa7a4fa4

This should redirect to: 
https://prod-preview.openshift.io/_home

Check in keycloak for the email verification status.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-261-3`

sbose78:
Thanks for the approval @alexeykazakov , I've added logging https://github.com/fabric8-services/fabric8-auth/pull/261/commits/15a322faf08edb7f8e2a1cd3d6a860045e88e4f1

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-261-4`

Issue #267, opened 2018-01-17
Add username for cluster tokens in oso-clusters.conf 


alexeykazakov:
We need to add a hardcoded username for cluster tokens in https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/conf-files/oso-clusters.conf so we can return this username as part of /token?for=... result payload for cluster tokens.

PR #269, opened 2018-01-17
Add username for cluster tokens

alexeykazakov:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/267

We need to update the cluster-config.json in prod and preview too. Otherwise it won't deploy.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-269-1`

PR #272, opened 2018-01-17
Added refresh api in /api/token endpoint

kishansagathiya:
Sample Requests:

Authorize
```
http://localhost:8089/api/authorize?response_type=code&client_id=740650a2-9c44-4db5-b067-a3d1b2cd2d01&scope=user:email&state=cd24b3c3-98aa-4ac1-9d01-eb3ece01419b&redirect_uri=http://localhost:8089/api/status
```
Exchange code for token
```

curl -v -d '{"grant_type":"authorization_code","client_id": "740650a2-9c44-4db5-b067-a3d1b2cd2d01", "code":"$AUTHORIZATION_CODE", "redirect_uri":"http://localhost:8089/api/status"}' -H "Content-Type: application/json" -X POST http://localhost:8089/api/token -D-
```
Use refresh token generate a new token
```
curl -v -d '{"grant_type":"refresh_token","client_id": "740650a2-9c44-4db5-b067-a3d1b2cd2d01", "refresh_token":"$REFRESH_TOKEN", "scope": "openid"}' -H "Content-Type: application/json" -X POST http://localhost:8089/api/token -D-
```

Fixes https://github.com/fabric8-services/fabric8-auth/issues/270

Issue #270, opened 2018-01-17
Support token refresh in /api/token


kishansagathiya:
The `/api/token/refresh` endpoint is not compliant with the OIDC spoecification: token refresh should be achived with the same endpoint as token retrieval
(/api/token), but with different parameters (especially the `refresh_token` grant type), as described in the specification here:
http://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken

We will have to support both endpoints. So create refresh api at `/api/token` and reverse proxy `/api/token/refresh` to `/api/token`

alexeykazakov:
To be clear. We can't remove the existing endpoint. It's still used by UI. We need to keep both until we update UI. Then we can remove the existing one.

kishansagathiya:
`/refresh` api returns `AuthToken`, while `/token` api returns OAuthToken. I think `/refresh` should also return `AuthToken`,  http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse

sbose78:
@kishansagathiya let's leave /api/token/refresh as is, for now - since it's a critical piece of code called by the UI. 

In the `POST /api/token` , return `OAuthToken` as defined in the spec, and as already being done in the `Exchange` controller.

PR #273, opened 2018-01-18
Add expires_in to token response. Some cleanup

alexeykazakov:
- Adds `expires_in` to token response. `expires_in` seems to be a standard name for that attribute in OAuth2 and OpenID Connect but some dialects uses expiry. So, let's use both to be safe.
- Some clean up

This is a followup on https://github.com/fabric8-services/fabric8-auth/pull/272 and https://github.com/fabric8-services/fabric8-auth/issues/270

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -346,36 +346,37 @@ func (c *TokenController) Delete(ctx *app.DeleteTokenContext) error {
 	return ctx.OK([]byte{})
 }
 
-// Exchange provides OAuth2 token exchange. Currently only grant_type="client_credentials" is supported
-// allowing clients to authenticate using a service account ID and secret value.
+// Exchange provides OAuth2 and OpenID Connect token exchange.
+// Currently only grant_type="client_credentials", "authorization_code", and "refresh_token" are supported.
+//
+// grant_type="client_credentials" allows clients to authenticate using a service account ID and secret value.
 // A service account token is returned as the result of successful exchange.
-// May be expanded in the future to support other exchange types.
+//
+// grant_type="authorization_code" is part of OAuth2 authorization flow.
+//
+// grant_type="refresh_token" covers OpenID Connect token refresh flow.
 func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
-
-	const clientCredentials = "client_credentials"
-	const authorizationCode = "authorization_code"
-	const refreshToken = "refresh_token"
-
 	payload := ctx.Payload
 	if payload == nil {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("payload", "nil").Expected("not empty payload"))
 	}
 
-	log.Info(ctx, map[string]interface{}{
+	log.Debug(ctx, map[string]interface{}{
 		"client_id":  payload.ClientID,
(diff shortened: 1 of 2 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-273-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-273-2`

design/token.go
```diff
@@ -218,12 +218,14 @@ var OauthToken = a.MediaType("application/vnd.oauthtoken+json", func() {
 	a.Description("Oauth 2.0 token payload")
 	a.Attributes(func() {
 		a.Attribute("access_token", d.String, "Access token")
-		a.Attribute("expiry", d.String, "Expiry")
+		a.Attribute("expires_in", d.String, "Access token expires in seconds")
+		a.Attribute("expiry", d.String, "The same as expires_in.")
```
kishansagathiya: Can't we just get rid of `expiry` ? `grant_type=refresh_token` and `grant_type=authorization_code` are not used by any other services so far and in `grant_type=client_credentials` we are not passing `expires_in` or `expiry` (why?).

controller/token.go
```diff
@@ -346,36 +346,37 @@ func (c *TokenController) Delete(ctx *app.DeleteTokenContext) error {
 	return ctx.OK([]byte{})
 }
 
-// Exchange provides OAuth2 token exchange. Currently only grant_type="client_credentials" is supported
-// allowing clients to authenticate using a service account ID and secret value.
+// Exchange provides OAuth2 and OpenID Connect token exchange.
+// Currently only grant_type="client_credentials", "authorization_code", and "refresh_token" are supported.
+//
+// grant_type="client_credentials" allows clients to authenticate using a service account ID and secret value.
 // A service account token is returned as the result of successful exchange.
```
kishansagathiya: We should also remove this line, since it doesn't return just service account token any more

design/token.go
sbose78: Fair point. 
What does the spec say? 

grant_type=client_credentials doesn't have an expiry because it is not a jwt, plus if needed we will simply change the credentials everywhere.

sbose78: But in future, this will be consumed in other places too ( in the platform UI for example ), so it'll be nice to have Support for all dialects.

kishansagathiya: As per specs `expires_in` in not required, but recommended https://tools.ietf.org/html/rfc6749#section-5.1

kishansagathiya: OAuth2.0 and OpenID Connect both uses `expires_in`. `expiry` might have come from an unreliable reference or from docs of some external provider who doesn't follow specs properly, in which case we don't need to support `expiry` any more. Also the word `expiry` mislead you for considering it as a time-stamp.

controller/token.go
alexeykazakov: This part of the comment is for client_credentials only.

kishansagathiya: Oh yes, my bad. No problem in keeping it then

PR #276, opened 2018-01-19
Remove expiry

alexeykazakov:
OAuth2 and OpenID Connect specifies expire_in. Not expiry. Some dialects still use expiry (for example go oauth2 token uses expiry) but we do not have any clients yet which use our authorize and token/refresh endpoints. So, it's safe now to support the OAuth2 and OpenID Connect expire_in attribute only and drop expiry. Especially because we misused expiry attribute. Expiry is supposed to be an int which represents a Unix time stamp. Not a string.

### DIFF
--- a/controller/token_blackbox_test.go
+++ b/controller/token_blackbox_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"golang.org/x/oauth2"
+	"strconv"
 )
 
 type TestTokenREST struct {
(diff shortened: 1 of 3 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-276-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-276-2`

controller/token_blackbox_test.go
```diff
@@ -221,6 +222,9 @@ func (rest *TestTokenREST) checkAuthorizationCode(service *goa.Service, controll
 	require.Equal(rest.T(), rest.sampleAccessToken, *token.AccessToken)
 	require.NotNil(rest.T(), token.RefreshToken)
 	require.Equal(rest.T(), rest.sampleRefreshToken, *token.RefreshToken)
+	expiresIn, err := strconv.Atoi(*token.ExpiresIn)
+	require.Nil(rest.T(), err)
+	require.True(rest.T(), expiresIn > 60*59*24*30 && expiresIn < 60*61*24*30) // The expires_in should be withing a minute range of 30 days.
```
sbose78: nice :)

PR #294, opened 2018-01-24
Build docs from sources

alexeykazakov:
The doc sources were changed but the html was not re-built. Fixing this.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-294-1`

PR #306, opened 2018-01-27
Validate all incoming tokens

alexeykazakov:
Regardless of security declaration in goa design definition we should validate all incoming tokens.
For example if we do not require Authorization header in /api/users/:ID we still check the token in case the request is coming from a Service Account. We should validate such tokens anyway.

This PR modifies our logging middleware. Now we not only add all the tokens in the context but also validate them. If there is Authorization header with a bearer token then we will validate it. If token is not valid then 401 will be returned.

### DIFF
--- a/goamiddleware/jwt_token_context.go
+++ b/goamiddleware/jwt_token_context.go
@@ -2,29 +2,22 @@ package goamiddleware
 
 import (
 	"context"
-	"crypto/ecdsa"
-	"crypto/rsa"
 	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/fabric8-services/fabric8-auth/token"
 
-	"github.com/dgrijalva/jwt-go"
 	"github.com/goadesign/goa"
-	goajwt "github.com/goadesign/goa/middleware/security/jwt"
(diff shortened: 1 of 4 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-306-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-306-2`

sbose78:
And also tried out incorrect token!
`{"errors":[{"code":"unauthorized_error","detail":"There is no public key with such ID: zD-57oBFIMUZsAWqUnIsVu_x71VIjd1irGkGUOiTsL8","status":"401","title":"Unauthorized error"}]}`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-306-3`

aslakknutsen:
> And also tried out incorrect token!

That was kinda the important part.. ;)

sbose78:
I was more worried about breaking things, initially. :D

PR #307, opened 2018-01-27
Explicitly read response body when closing it to avoid FD leaks

alexeykazakov:
It seems that we already read the body before we close it but maybe we still miss it somewhere or there is a panic or something else. So just in case let's explicitly read response body when closing it in defer to make sure we are not leaking FDs.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-307-1`

PR #309, opened 2018-01-29
Add trailing slash to user.cluster, APIURL, ConsoleURL, and MetricsURL

alexeykazakov:
Regardless of actual URLs we store in user's profile and cluster configuration for OSO API, Console, and Metrics URL we should return these URLs with a trailing slash (for example "https://api.*.openshift.com/")


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-309-1`

PR #313, opened 2018-01-30
Create OpenID compliant `/userinfo` Endpoint

### LINKED ISSUE #285: Create OpenID compliant `/userinfo` Endpoint
1. Create new `/userinfo` endpoint should be like

Request 
```
  GET /userinfo HTTP/1.1
  Host: server.example.com
  Authorization: Bearer SlAV32hkKG

```
Response
```
 HTTP/1.1 200 OK
  Content-Type: application/json

  {
   "sub": "248289761001",
   "given_name": "Jane",
   "family_name": "Doe",
   "preferred_username": "j.doe",
   "email": "janedoe@example.com",
  }
```
2.  Add `/user` to openid-configuration

kishansagathiya:
- [x]   Create new `/userinfo` endpoint should be like

Request 
```
  GET /userinfo HTTP/1.1
  Host: server.example.com
  Authorization: Bearer SlAV32hkKG

```
Response
```
 HTTP/1.1 200 OK
  Content-Type: application/json

  {
   "sub": "248289761001",
   "given_name": "Jane",
   "family_name": "Doe",
   "preferred_username": "j.doe",
   "email": "janedoe@example.com",
  }
```
- [x]  Add `/userinfo` to openid-configuration
- [ ]  Tests

Fixes #285


controller/userinfo.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+
+		sub := identity.ID.String()
+		userInfo := &app.UserInfo{
+			Sub:           &sub,
+			GivenName:     &user.FullName,
+			PreferredName: &identity.Username,
+			FamilyName:    &user.FullName,
+			Email:         &user.Email,
```
kishansagathiya: Let me know if this seems inappropriate. FamilyName is inappropriate, but I don't think we have any appropriate properties

design/account.go
```diff
@@ -93,6 +93,46 @@ var userList = JSONList(
...
+
+	a.Action("show", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.GET(""),
+		)
+		a.Description("Get the authenticated user")
+		//a.UseTrait("conditional")
+		a.Response(d.OK, userInfo)
+		//a.Response(d.NotModified)
```
kishansagathiya: Commented to avoid errors for now. Will uncomment this in next commit with support of `a.Response(d.NotModified)`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-1`

controller/userinfo.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Show runs the show action.
+func (c *UserinfoController) Show(ctx *app.ShowUserinfoContext) error {
+	id, err := c.tokenManager.Locate(ctx)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "Bad Token")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("Bad Token"))
```
kishansagathiya: In GET `/user` this is bad request, but I think if token is invalid Unauthorized would me more appropriate. Let me know what do you think. Also I have changed all generic errors (pkg/errors) to autherrors(fabric8-auth/errors)

controller/userinfo.go
```diff
@@ -0,0 +1,71 @@
...
+				"identity_id": id,
+			}, "Auth token containers id %s of unknown Identity", id)
+			return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(fmt.Sprintf("Auth token contains id %s of unknown Identity\n", id)))
+		}
+		var user *account.User
+		userID := identity.UserID
+		if userID.Valid {
+			user, err = appl.Users().Load(ctx.Context, userID.UUID)
+			if err != nil {
+				return jsonapi.JSONErrorResponse(ctx, errors.NewInternalErrorFromString(ctx, fmt.Sprintf("Can't load user with id %s", userID.UUID)))
```
kishansagathiya: GET `/user` seems to return NotFound error((https://github.com/fabric8-services/fabric8-auth/blob/master/account/user.go#L99) ) here. But API design of GET `/user` doesn't mention NotFound as a valid response

controller/userinfo.go
sbose78: `FamilyName` is fine: https://github.com/fabric8-services/fabric8-auth/blob/master/token/token.go#L62  Does the spec say anything? Keycloak uses `familyName`

controller/userinfo.go
sbose78: Let's stick to `NotFound` here too ?

controller/userinfo.go
sbose78: umm. didn't get you, what do you mean by 'we dont have any appropriate properties?'

kishansagathiya:
@sbose78 I think I was supposed to extract claims from the token. I am doing the whole thing wrong way. Let me fix this in next commit. Thanks for comments.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-2`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-4`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-5`

sbose78:
@kishansagathiya 
> I think I was supposed to extract claims from the token

Is this a requirement?  If not, why are we returning user details differently here? I'm a bit confused as to why you did that :)

Also,
1. What if the user is not present in the db ?
2. Should we be returning just those 5 or 6 attributes of the user and not the rest ?

kishansagathiya:
@sbose78 The userinfo that I have returned are default claims and from the specs this is how we are supposed to do it. If the user is not present in the database the token wouldn't have been generated in the first place.
The attributes that we are returning are required by che team, but there is not harm in returning more.

alexeykazakov:
@kishansagathiya as far as I can see the only claim/info which is required by OpenID Connect is "sub". It's our Identity ID.
So, let's add only sub + info which is expected by Che to the response for now.
If the following info is enough for Che:
```
  {
   "sub": "248289761001",
   "given_name": "Jane",
   "family_name": "Doe",
   "preferred_username": "j.doe",
   "email": "janedoe@example.com",
  }
```
then let's return these claims only. All these claims are available in our DB.
Here are some reasons why we should load this info from DB:
- Token could contain an outdated information. If the client only need the info from the token then why it calls the user info endpoint? Why not just parse the token and get the required info from it?
- We do generate tokens for identities which do not exist in our DB. See api_client param in our login endpoint.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-7`

kishansagathiya:
@alexeykazakov Let me know if this approach seem appropriate. if so i will finish remaining tests.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-8`

controller/userinfo.go
alexeykazakov: Yes, I think 401 is more appropriate here. Just use lower case and add the original error message:
jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("bad token: " + err.String()))

controller/userinfo.go
alexeykazakov: It's rather 401 if we can't found the corresponding user I think..

design/account.go
```diff
@@ -93,6 +93,44 @@ var userList = JSONList(
 	pagingLinks,
 	userListMeta)
 
+// userInfo represents an userInfo object
+var userInfo = a.MediaType("application/vnd.userInfo+json", func() {
+	a.UseTrait("jsonapi-media-type")
+	a.TypeName("UserInfo")
+	a.Description("User Information")
+	a.Attributes(func() {
+		a.Attribute("sub", d.String)
```
alexeykazakov: Add descriptions.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-10`

account/userprofile/service.go
```diff
@@ -0,0 +1,72 @@
+package userprofile
```
sbose78: just call it userinfo? because we already have keycloak user profile, would start getting confusing.

login/service.go
```diff
@@ -1091,3 +1092,42 @@ func TokenToJson(ctx context.Context, outhToken *oauth2.Token) (string, error) {
 
 	return string(b), nil
 }
+
+// UserInfo gets user infomation given a context containing access_token
+func (keycloak *KeycloakOAuthProvider) UserInfo(ctx context.Context) (*account.User, *account.Identity, error) {
```
sbose78: not needed anymore.

account/userprofile/service.go
kishansagathiya: or just user?

kishansagathiya:
@sbose78  yet to change the package name from `userprofile` to something else. Let me know if `user` works for you. I will push again after that

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-11`

account/userprofile/service.go
sbose78: might be confusing because we have a user.go already :)

kishansagathiya: userprofile package is used by both user.go and userinfo.go

sbose78: makes sense. Let's go with `user` as the package name. And please address review comments using the commit IDs wherever possible. :)

kishansagathiya: sure

login/service.go
kishansagathiya: https://github.com/fabric8-services/fabric8-auth/pull/313/commits/7c46e498276b11657b352032576054621093f36a

kishansagathiya:
[test]

alexeykazakov:
[test]

sbose78:
@kishansagathiya sorry for not noticing this before, but could you please add a test for `account/userinfo/service_blackbox_test.go` ? I know we already test the controller, but since we have a separate service, it is important that is an independent test for the service as well.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-13`

controller/userinfo.go
```diff
@@ -0,0 +1,51 @@
...
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+
+	fullName := strings.Split(user.FullName, " ")
+	sub := identity.ID.String()
+	userInfo := &app.UserInfo{
+		Sub:               &sub,
+		GivenName:         &fullName[0],
+		PreferredUsername: &identity.Username,
+		FamilyName:        &fullName[1],
```
alexeykazakov: It will panic here if there is no " " in the full name. Use account.SplitFullName(user.FullName) to get first and last names.

account/userinfo/service.go
```diff
@@ -0,0 +1,72 @@
...
+	return &AccountProvider{
+		Identities:   identities,
+		Users:        users,
+		TokenManager: tokenManager,
+		DB:           db,
+	}
+}
+
+type AccountProvider struct {
+	Identities     account.IdentityRepository
```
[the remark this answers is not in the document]
kishansagathiya: UserInfoProvider? trying to be consistent with the packagename

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-14`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-15`

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-313-16`

PR #314, opened 2018-01-31
ISSUE-301 migration script for new authorization features

### LINKED ISSUE #301: Database schema changes for Authorization
The following schema changes are to be made for the Authorization API:

1. Rename IDENTITY_MANAGEMENT to MEMBERSHIP

Since this table hasn't been created yet, it is a simple matter to just create the new table:

```
CREATE TABLE MEMBERSHIP (
    member_of uuid NOT NULL references identities(id),
    member_id uuid NOT NULL references identities(id),
    PRIMARY KEY (member_of, member_id)
);
```

2. Create a new RPT_TOKEN table (this table will track RPT tokens):

```
CREATE TABLE RPT_TOKEN (
    token_id uuid primary key,
    expiry_time timestamp with time zone,
    identity_id uuid NOT NULL references identities(id),
    status integer NOT NULL
);
```

3. Create a new TOKEN_RESOURCE table (tracks which resources are contained in a token):

```
CREATE TABLE TOKEN_RESOURCE (
    token_id uuid NOT NULL references rpt_token(token_id),
    resource_id uuid NOT NULL references resource(resource_id),
    last_accessed timestamp with time zone
);
```

4. Alter the data type of IDENTITY_ROLE.IDENTITY_ROLE_ID to uuid:

```
alter table identity_role drop column identity_role_id;
alter table identity_role add column identity_role_id uuid primary key;
```

5. Add an IDENTITY_RESOURCE_ID column to identities:

`alter table identities add column identity_resource_id uuid references resource (resource_id);`

6. Remove the OWNER_ID column from the RESOURCE table:

`alter table RESOURCE drop column OWNER_ID`

sbryzak:
Fixes #301 

Database migration scripts to prepare schema for new authorization features.

### DIFF
--- a/migration/sql-files/019-authorization-part-2.sql
+++ b/migration/sql-files/019-authorization-part-2.sql
@@ -1,20 +1,22 @@
 ALTER TABLE resource DROP CONSTRAINT resource_parent_resource_id_fkey;
 ALTER TABLE identity_role DROP CONSTRAINT identity_role_resource_id_fkey;
 
-ALTER TABLE resource ALTER COLUMN resource_id TYPE varchar;
-ALTER TABLE resource ALTER COLUMN parent_resource_id TYPE varchar;
-ALTER TABLE resource ALTER COLUMN name TYPE varchar;
+ALTER TABLE resource ALTER COLUMN resource_id TYPE varchar(256);
+ALTER TABLE resource ALTER COLUMN parent_resource_id TYPE varchar(256);
+ALTER TABLE resource ALTER COLUMN name TYPE varchar(256);
 ALTER TABLE resource DROP COLUMN owner_id;
 
 ALTER TABLE resource ADD CONSTRAINT resource_parent_resource_id_fkey FOREIGN KEY (parent_resource_id) REFERENCES resource (resource_id);
 
-ALTER TABLE identity_role ALTER COLUMN resource_id TYPE varchar;
+ALTER TABLE resource ADD COLUMN creator_id uuid REFERENCES Identities (id);
+
+ALTER TABLE identity_role ALTER COLUMN resource_id TYPE varchar(256);
 ALTER TABLE identity_role ADD CONSTRAINT identity_role_resource_id_fkey FOREIGN KEY (resource_id) REFERENCES resource (resource_id);
 
-ALTER TABLE identities ADD COLUMN identity_resource_id varchar;
+ALTER TABLE identities ADD COLUMN identity_resource_id varchar(256);
 ALTER TABLE identities ADD CONSTRAINT identities_identity_resource_id_fkey FOREIGN KEY (identity_resource_id) REFERENCES resource (resource_id);
 
-ALTER TABLE role ALTER COLUMN name TYPE varchar;
+ALTER TABLE role ALTER COLUMN name TYPE varchar(256);
 
 CREATE TABLE membership (
   member_of uuid NOT NULL references identities (id),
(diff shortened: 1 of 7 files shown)


fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-314-1`

alexeykazakov:
There are test failures.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-314-2`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-314-3`

migration/sql-files/019-authorization-part-2.sql
```diff
@@ -0,0 +1,36 @@
+ALTER TABLE resource DROP CONSTRAINT resource_parent_resource_id_fkey;
+ALTER TABLE identity_role DROP CONSTRAINT identity_role_resource_id_fkey;
+
+ALTER TABLE resource ALTER COLUMN resource_id TYPE varchar;
+ALTER TABLE resource ALTER COLUMN parent_resource_id TYPE varchar;
```
sbose78: Any reason why it is varchar and not uuid ?

migration/sql-files/019-authorization-part-2.sql
```diff
@@ -0,0 +1,36 @@
...
+ALTER TABLE identity_role ADD CONSTRAINT identity_role_resource_id_fkey FOREIGN KEY (resource_id) REFERENCES resource (resource_id);
+
+ALTER TABLE identities ADD COLUMN identity_resource_id varchar;
+ALTER TABLE identities ADD CONSTRAINT identities_identity_resource_id_fkey FOREIGN KEY (identity_resource_id) REFERENCES resource (resource_id);
+
+ALTER TABLE role ALTER COLUMN name TYPE varchar;
+
+CREATE TABLE membership (
+  member_of uuid NOT NULL references identities (id),
+  member_id uuid NOT NULL references identities (id),
```
sbose78: Is this the auto-generated ID for this table? In that case, why didn't we name it `id` instead of `member_id` ( just like `id` in the user table. :)  
No strong opinions, just curious.

sbose78: This is the identity_id I think, right?

migration/sql-files/019-authorization-part-2.sql
sbryzak: Yes, the resource_id value can actually be set by the resource server when the resource is registered.  In the case of WIT, they will use the natural key value of the resource that is being created.  It will most likely be a uuid, but it doesn't have to be.  In any case the column is indexed and has a unique constraint, so there are no issues with performance or duplicate key values.

migration/sql-files/019-authorization-part-2.sql
sbryzak: There is no auto-generated ID for the membership table, it is simply a many-to-many relationship between Identities.  The member_id is the identifier of the member identity, and the member_of is the identifier of the group/team/organization that it is a member of.

migration/sql-files/019-authorization-part-2.sql
sbose78: Thanks - makes sense!

alexeykazakov: I'm not sure by maybe we should limit the size of varchars? To make sure resource services are not setting some crazy long IDs.

sbryzak: +1, what do you think is a reasonable size?

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-314-4`

Issue #292, opened 2018-01-24
Implement List Organizations endpoint


sbryzak:
This endpoint lists the organizations to which a user is an owner, a member, or to which they have been designated any other role.

See [this section](https://docs.google.com/document/d/1qmpH8Yex0yKmcTauPqVoGmquRiEbab1E8fFf90tEw3A/edit#heading=h.lqkgi6kanilk) of the detailed design documentation for more information.

```
GET /api/organization HTTP/1.1
Content-Type: application/json
Host: as.example.com
Authorization: Bearer 204c69636b6c69
```

```
HTTP/1.1 200 OK
...
[{
   "organization_id":"4444-3333-5555-6666-HHHH-BBBB-AAAA",
   "name":"Red Hat",
   "owner":"false",
   "roles":[
      "admin"
   ]
}]
```

PR #317, opened 2018-02-02
ISSUE-290, ISSUE-292 initial code for creating and listing organizations

### LINKED ISSUE #290: Implement Create Organization endpoint
This endpoint is used to create a new organization.

See [this section](https://docs.google.com/document/d/1qmpH8Yex0yKmcTauPqVoGmquRiEbab1E8fFf90tEw3A/edit#heading=h.myfuzmxxrg1y) of the detailed design document for more information.

```
POST /api/organization HTTP/1.1
Content-Type: application/json
Host: as.example.com
Authorization: Bearer 204c69636b6c69
...

{  
   "name":"Acme"
}
```

```
HTTP/1.1 200 OK
Content-Type: application/json
...
{  
   "organization_id":"4444-3333-5555-6666-HHHH-BBBB-AAAA"
}
```

This endpoint should execute two insert statements:

```
INSERT INTO RESOURCE (
		resource_id,
		parent_resource_id,
		owner_id,
		resource_type_id,
		name)
	VALUES (
		(new UUID value),
		null,
		(identity id of the user creating the organization),
		(select resource_type_id from resource_type where name = ‘organization’),
		‘Red Hat’
);
```

and

```
	INSERT INTO IDENTITY (
		identity_id,
		identity_resource_id)
	VALUES (
		(new UUID value),
		(the resource_id for the row we just created in the resource table)
	);
```

sbryzak:
fixes #290 and #292 

This PR implements both the create organization and list organization endpoints.

Create organization:

```
POST /api/organizations
Authorization: Bearer <token>

{"name":"Organization Name"}
```

Response:

`{"organization-id": "ec54103e-962c-4f53-a530-3c5408623c31"}`

List organizations:

```
GET /api/organizations
Authorization: Bearer <token>
```

Response:

```
{"data": [
  {"id":"ec54103e-962c-4f53-a530-3c5408623c31",
   "name":"Organization Name",
   "member":"false",
   "roles":["owner"]}
]}
```


fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-3`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-4`

sbryzak:
The list organizations endpoint was tricky to implement, but I think I've got the queries right now.  Still some work to do for it, also will need a metric ton of tests to cover all the scenarios (of which there are many).

sbose78:
Haha I agree, we are going to have some complex queries here :)

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-7`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-8`

sbryzak:
Ready for review now, everything should be good to go.  We'll need to add more integration testing once the other endpoints are implemented for managing teams, groups, invites, etc.  Lots of cross-testing between functions will be required.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-10`

account/identity.go
```diff
@@ -21,6 +22,11 @@ import (
 const (
 	// KeycloakIDP is the name of the main Keycloak Identity Provider
 	KeycloakIDP string = "kc"
+
+	IDENTITY_RESOURCE_TYPE_ORGANIZATION = "identity/organization"
+	IDENTITY_RESOURCE_TYPE_TEAM         = "identity/team"
+	IDENTITY_RESOURCE_TYPE_GROUP        = "identity/group"
+	IDENTITY_RESOURCE_TYPE_USER         = "identity/user"
```
sbose78: Let's use camelcase as a Go convention?
https://stackoverflow.com/a/22688926/823386 unless we have a reason not to :)

account/identity.go
```diff
@@ -57,6 +63,14 @@ func (u NullUUID) Value() (driver.Value, error) {
 	return u.UUID.Value()
 }
 
+// This struct is used to return the Organizations for which an Identity is associated
+type IdentityOrganization struct {
```
sbose78: Wondering if we should move out the authorization/organization related methods/structs to a different go file in the same package ? Saying this, so that we can keep `account/identity.go` clean and simple, and move out the complex derivatives to smaller/logical files into something like `account/organization.go`

controller/organization.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/fabric8-services/fabric8-auth/jsonapi"
+	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/fabric8-services/fabric8-auth/login"
+	"github.com/fabric8-services/fabric8-auth/token"
+	"github.com/goadesign/goa"
+	"github.com/satori/go.uuid"
+	"strings"
+)
+
+const ORGANIZATION_OWNER_ROLE = "owner"
```
sbose78: Could we use camel case?

controller/organization.go
```diff
@@ -0,0 +1,157 @@
...
+		// Assign the owner role for the new organization to the current user
+		identityRole := &role.IdentityRole{
+			IdentityID: userIdentity.ID,
+			ResourceID: res.ResourceID,
+			RoleID:     ownerRole.RoleID,
+		}
+
+		err = appl.IdentityRoleRepository().Create(ctx, identityRole)
+		if err != nil {
+			return jsonapi.JSONErrorResponse(ctx, goa.ErrInternal(err.Error()))
```
sbose78: We typically use `errors.NewInternalError(..)` or `errors.NewInternalErrorFromString(..)` from our `errors` packages.

controller/organization.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+
+	results := []*app.OrganizationData{}
+
+	for _, org := range orgs {
+		orgData := &app.OrganizationData{
```
sbose78: Could you move this logic out to a `func toAppOrganization(..)` like the way we do in other places https://github.com/fabric8-services/fabric8-auth/blob/master/controller/collaborators.go#L108

controller/organization_blackbox_test.go
```diff
@@ -0,0 +1,135 @@
...
+	require.Equal(rest.T(), orgName, org.Name)
+	require.Equal(rest.T(), 1, len(org.Roles))
+	require.Equal(rest.T(), ORGANIZATION_OWNER_ROLE, org.Roles[0])
+}
+
+/*
+* This test will attempt to create a new organization
+ */
+func (rest *TestOrganizationREST) TestCreateOrganizationUnauthorizedFail() {
+	/*sa := account.Identity{
```
sbose78: commented out? :)

design/organization.go
```diff
@@ -0,0 +1,77 @@
...
+		a.Attribute("organization_id")
+	})
+})
+
+var organizationArray = a.MediaType("application/vnd.organization-array+json", func() {
+	a.UseTrait("jsonapi-media-type")
+	a.TypeName("OrganizationArray")
+	a.Description("Organization Array")
+	a.Attributes(func() {
+		a.Attribute("organizations", a.ArrayOf(organizationData))
```
sbose78: Let's call this attribute `data` to keep some consistency with other APIs. In the future, we might have to consider pagination. In the jsonapi spec, we would usually have a `type = organizations` if at all. Let's skip the latter and use `data` now?

migration/migration.go
```diff
@@ -170,6 +170,9 @@ func GetMigrations(configuration MigrationConfiguration) Migrations {
 	// Version 20
 	m = append(m, steps{ExecuteSQLFile("020-add-response-mode-to-auth-state-reference.sql")})
 
+	// verion 21
+	m = append(m, steps{ExecuteSQLFile("021-organization-dml.sql")})
```
sbose78: There are a lot of migrations happening here - could we please add some tests to it?

Example: https://github.com/fabric8-services/fabric8-auth/pull/322/files#diff-8300fcfb905de8543555d4fff40a7c2f
The above is an overlap with some of your `identity_role` changes, and also tests whether the database now allows creation of `identity_role` entities with a uuid as the ID / primary key field.

sbose78:
Could you also add a 'how to test the API' section to the description, please ?

account/identity.go
sbryzak: Interesting, I didn't realise that Go broke established norms for constant naming conventions. ;) I've updated to use camel case.

account/identity.go
sbryzak: I thought about this, however since this struct is used nowhere else except for a single method in the identity.go file, it didn't make much sense to me to move it.

controller/organization.go
sbryzak: No problem, I'll update it.

controller/organization.go
sbryzak: I've gone through and updated these - there is one instance where I use the following:

		return jsonapi.JSONErrorResponse(ctx, goa.ErrBadRequest("Organization name cannot be empty"))

The errors.NewBadParameterError() method doesn't seem to be capable of accepting a msg parameter, so is there another way of returning this error with a message or is the above code the only way to do this?

controller/organization.go
sbryzak: I've moved this out to a separate function as suggested

controller/organization_blackbox_test.go
sbryzak: I've removed this test for now

design/organization.go
sbryzak: Done.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-12`

account/identity.go
sbose78: I was thinking about moving this and the `ListOrganizations(...)` method, what do you think?

controller/organization.go
sbose78: `errors.NewBadParameterError("organization","empty")` not the cleanest way, but this could be done. I don't have a very strong opinion on this - so you could do either ways :)

controller/organization.go
```diff
@@ -34,7 +34,7 @@ func NewOrganizationController(service *goa.Service, db application.DB) *Organiz
 func (c *OrganizationController) Create(ctx *app.CreateOrganizationContext) error {
 	currentUser, err := login.ContextIdentity(ctx)
 	if err != nil {
-		return jsonapi.JSONErrorResponse(ctx, goa.ErrUnauthorized(err.Error()))
+		return errors.NewUnauthorizedError(err.Error())
```
sbose78: What I implied is `jsonapi.JSONErrorResponse(ctx,errors.NewUnauthorizedError(err.Error()))`

account/identity.go
sbryzak: How about we create a new file called identity_management.go and put all IdM related functions there?  Would the ListOrganizations() etc methods still be provided by IdentityRepository, or are you thinking they should be moved to a new Repository class (IdentityManagementRepository?) instead?

migration/migration.go
sbryzak: I've added some tests for the db changes.

account/identity.go
sbose78: I'm not in favor of making a new repository class :) We could just move the one struct and the one repository method into the different file for slightly better code organization - not really considering moving that to a different `interface implementation `   at the moment.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-13`

account/identity.go
sbryzak: No problem, I'll just move the struct and ListOrganizations() method to a new file, but keep them in the existing IdentityRepository.

account/identity.go
```diff
@@ -21,6 +22,11 @@ import (
 const (
 	// KeycloakIDP is the name of the main Keycloak Identity Provider
 	KeycloakIDP string = "kc"
+
+	IdentityResourceTypeOrganization = "identity/organization"
+	IdentityResourceTypeTeam         = "identity/team"
+	IdentityResourceTypeGroup        = "identity/group"
+	IdentityResourceTypeUser         = "identity/user"
```
sbose78: Wondering if for consistency's sake we should name these `openshift.io/resource......` or is it because we want to name these differently ?

controller/organization.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Create runs the create action.
+func (c *OrganizationController) Create(ctx *app.CreateOrganizationContext) error {
+	currentUser, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+
+	if len(strings.TrimSpace(*ctx.Payload.Name)) == 0 {
```
sbose78: might want to check if `ctx.Payload.Name` != nil as well?

controller/organization.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	var organizationId uuid.UUID
+
+	err = application.Transactional(c.db, func(appl application.Application) error {
+
+		// Lookup the identity for the current user
+		userIdentity, err := appl.Identities().Load(ctx, *currentUser)
+		if err != nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalErrorFromString(ctx, "Error looking up current user"))
```
sbose78: `appl.Identities().Load(ctx, *currentUser)` also returns an error if the user doesn't exist in the db, and in that case it is should be reported by the API as an unauthorized error  
```
return errors.NewUnauthorizedError(fmt.Sprintf("auth token contains id %s of unknown Identity\n", *id))
```

controller/organization.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		// Lookup the organization resource type
+		resourceType, err := appl.ResourceTypeRepository().Lookup(ctx, account.IdentityResourceTypeOrganization)
+		if err != nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalErrorFromString(ctx, "Error looking up resource type 'identity/organization'"))
```
sbose78: Let's not cast the error to an internal error because `Lookup` might return an `error` even if the record was not found in the database 
https://github.com/fabric8-services/fabric8-auth/pull/317/files#diff-76f758602c8a7d4581a1e3e77b1bda3fR217
in that case, the error should get reported as NotFound and not InternalError
So, I suggest we keep it as 
```
```

Sorry, didn't notice this earlier.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-14`

controller/organization.go
sbryzak: Good idea, done.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-15`

controller/organization.go
sbryzak: Thanks, I've updated this.

controller/organization.go
sbryzak: Done.

sbryzak:
I've updated the description with details of the new endpoints, including sample request/response payloads for testing.

account/identity.go
sbryzak: These have been moved now to `account/identity_management.go`

account/identity.go
sbryzak: This naming was intentional, I wanted to differentiate these resource types from the openshift.io resource types, as they are only handled internally by the auth service.

controller/organization.go
sbryzak: I fixed all this, my mistake.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-16`

authorization/role/identity_role.go
```diff
@@ -24,11 +24,14 @@ type IdentityRole struct {
 	// This is the primary key value
 	IdentityRoleID uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key" gorm:"column:identity_role_id"`
 	// The identity to which the role is assigned
-	Identity account.Identity
+	IdentityID uuid.UUID
```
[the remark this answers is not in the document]
sbryzak: This test has been fixed (added the missing Create() call).

account/identity_management.go
```diff
@@ -0,0 +1,180 @@
...
+// This struct is used to return the Organizations for which an Identity is associated
+type IdentityOrganization struct {
+	OrganizationID uuid.UUID
+	Name           string
+	Member         bool
+	Roles          []string
+}
+
+// Returns an array of all organizations in which the specified user is a member or is assigned a role
+func (m *GormIdentityRepository) ListOrganizations(ctx context.Context, identityID uuid.UUID) ([]IdentityOrganization, error) {
```
sbose78: I think there's a test missing for this function.

migration/sql-files/021-organizations-list-create.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE identity_role drop column identity_role_id;
+ALTER TABLE identity_role ADD COLUMN identity_role_id uuid DEFAULT uuid_generate_v4();
+ALTER TABLE identity_role ADD CONSTRAINT pk_identity_role PRIMARY KEY (identity_role_id);
+CREATE UNIQUE INDEX uq_identity_role_identity_role_resource ON identity_role (identity_id, resource_id, role_id);
+
+CREATE UNIQUE INDEX uq_role_resource_type_name ON role (resource_type_id, name);
+
+INSERT INTO resource_type (name) VALUES ('identity/organization');
+
+INSERT INTO role (role_id, resource_type_id, name) select uuid_generate_v4(), resource_type_id, 'owner' from resource_type where name = 'identity/organization';
```
sbose78: Can we add tests in some form to test the `identity/organization` insertion in the relevant table(s) ?

```
INSERT INTO resource_type (name) VALUES ('identity/organization');
INSERT INTO role (role_id, resource_type_id, name) select uuid_generate_v4(), resource_type_id, 'owner' from resource_type where name = 'identity/organization';
```

sbryzak: We have these covered already - see:

[](https://github.com/fabric8-services/fabric8-auth/pull/317/files#diff-3c2879c729984887477d67ffd0960ffaR19)
[](https://github.com/fabric8-services/fabric8-auth/pull/317/files#diff-3c2879c729984887477d67ffd0960ffaR35)

Also a similar test exists for roles.

account/identity_management.go
sbryzak: It's tested as part of `controller/organization_blackbox_test.go` - since it's non-trivial to actually create an organization without using the controller anyway, it doesn't make sense to try and test this function separately.

sbose78: The tests `controller/organization_blackbox_test.go` is to test the controller irrespective of whether we are using the actual or mocked 'dao' /repository layer.

However the tests here at the repository / dao layer is to ensure that this `ListOrganizations` Golang API method works as per the contract - example, the tests in https://github.com/fabric8-services/fabric8-auth/blob/master/account/identity_blackbox_test.go

migration/sql-files/021-organizations-list-create.sql
sbose78: The migration tests are run in a database called `test` , which is not the same database where our other tests run. So effectively the above lines don't really cover that :) 
This `test` database is one which is created only to test out migration SQLs.
So in the `testMigration21` you could have something like this to test out whether the migration SQLs actually produce the intended outcome.
```
        rows, err := sqlDB.Query("SELECT name FROM resource_type")
	if err != nil {
		t.Fatal(err)
	}
	defer rows.Close()
	for rows.Next() {
                var resourceTypeName string
		err = rows.Scan(&resourceTypeName)
               // assert resourceTypeName ....
               // .....
	}
```

migration/sql-test-files/021-test-organizations.sql
```diff
@@ -0,0 +1,11 @@
+insert into resource_type (resource_type_id, name) values ('c3605c89-d2ae-4d3f-b4b6-dc0078531c9e', 'resource/migration-test');
+insert into resource (resource_id, resource_type_id, name) values('9c17b0a3-a56e-4e44-ad93-756ec85e94ac', 'c3605c89-d2ae-4d3f-b4b6-dc0078531c9e', 'migration-test-resource');
+insert into role (role_id, resource_type_id, name) values ('11b3a386-70ef-4ef5-bc5e-e897cb2ca859', 'c3605c89-d2ae-4d3f-b4b6-dc0078531c9e', 'migration-test-role');
+insert into identities (id, username, registration_completed) values ('7bb8876f-7d93-46ad-bbd3-733b77b76c55', 'migration-test-user', true);
+insert into identity_role (identity_role_id, resource_id, identity_id, role_id) values ('977182c3-71b9-4954-bd58-834219d6441b', '9c17b0a3-a56e-4e44-ad93-756ec85e94ac', '7bb8876f-7d93-46ad-bbd3-733b77b76c55', '11b3a386-70ef-4ef5-bc5e-e897cb2ca859');
+
+delete from identity_role where identity_role_id = '977182c3-71b9-4954-bd58-834219d6441b';
```
sbose78: its fine to have the cleanup SQLs, but we don't really need it.
because these SQLs are run against the `test` database :)

sbryzak: Without the cleanup, our tests fail because we actually count the total number of roles/resource types in the table and assert that the counts are as expected.

account/identity_management.go
sbryzak: So how would you suggest we create the organization?  Do we use the controller's Create() method? (I didn't know if we have access to the controllers from these tests?)

migration/sql-files/021-organizations-list-create.sql
sbryzak: No problem, I can add a check here for new types.

account/identity_management.go
sbose78: You could use repository methods or SQLs to do so - but usually we use the repository methods since they wrap the SQLs.
The way I did here https://github.com/fabric8-services/fabric8-auth/pull/322/files#diff-4bd1c185cd704de02aaa93a028975a89R50

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-17`

account/identity_management.go
sbryzak: Yes but as I said it's non-trivial to actually create an organization - you need to create a resource object, an identity object, assign a role to the user creating it, and so on.. just look at all the code in the organization controller's Create() method.  Are you suggesting I replicate all of this complex functionality in the repository test to create an organization simply so that I have data to test the ListOrganizations() method?  In this case, I would have thought that the integration tests would exercise this code sufficiently, especially since there are numerous other tests still to come once we implement organization/team/group memberships.

sbose78: > Are you suggesting I replicate all of this complex functionality in the repository test to create an organization simply so that I have data to test the ListOrganizations() method? 

Yes, because the controller tests are for validating the controller contract. The repository tests are for validating the repository contract in a specific package. The tests are critical here because the code in this package/repository is going to consumed by others packages.

sbryzak: I don't like the idea of duplicating so much code, how about I look at some way of moving the create organization logic to the repository codebase instead of having it in the controller?  The only question I have is whether it is ok to have a repository method access other repositories/tables.  To me this seems allowable, as the ListOrganizations() method is already querying multiple tables anyway.  What do you think?

sbose78: >  how about I look at some way of moving the create organization logic to the repository codebase instead of having it in the controller? 

That's a great idea irrespective of our conversation around tests. We should abstract out `CreateOrganization(..)` to a repository method !

In fact, ( contradictory to a few comments back ) I'm even wondering if we should go with an `OrganizationRepository` which has the two methods `Create(..)` , `List(..)` ? Looks a lot cleaner to me , that way :)

I am thinking along the lines of having `account/organization.go` .. which would have the  above `OrganizationRepository` definition & methods.

sbose78: Or you could define an `OrganizationService` in `account/organization.go` with `List` and `Create` options.

sbryzak: I like the idea of creating a new repository, however how about we call it IdentityManagementRepository and have it include the team/group and membership operations provided there also, as these are all very closely related?

sbryzak: Making it a service object is appealing also, as it could simply interact with the existing repositories / directly with the db via queries.

sbryzak: The issue I see with having different files for each thing, is that there are an awful lot of cross-cutting concerns here.  Assigning roles, managing memberships, etc all apply to organizations, teams and groups, not to mention querying permissions applies to all of them also.  If we don't want to group them together just because we don't like "IdentityManagement" as a name, is there an alternative name that we could use instead?

sbryzak: Ok, I've thought about this some more.  Perhaps we could have a separate service file for each type of user collection, i.e. OrganizationService, TeamService, GroupService, and then in addition to this have services for each of the cross-cutting concerns, e.g. RoleService, MembershipService, InvitationService, etc.  I think the number of new files we would be creating is reasonable this way.  Also, do you think we would access these services in the same way that we access the repositories? i.e. via the application object?  We could easily provide application.GetOrganizationService(), application.GetTeamService() and so on, which would set up the service objects with the database connection.

One last thought - placing these in the account package doesn't seem intuitive to me, as these features aren't really account-related.  It is a pity that the "account" package isn't called "identity" instead, then it would make more sense to put them there.  Do you have any suggestions as to the best place to locate these services?

sbose78: Fair concerns @sbryzak . 

I was thinking along the lines of how we would design the structs/interfaces irrespective of splitting files.

```
type OrganizationService interface {
         List(...)
         Create(...) // calls of a bunch of repository methods from roles/resources/identities         
         AssignRole(..) // calls `RoleRepository` methods
         GetPermissions(..)
}
```

```
type TeamService interface {
         List(...)  
         Create(...) // calls of a bunch of repository methods from roles/resources/identities
         AssignRole(..) // calls `RoleRepository` methods
         GetPermissions(..)
}
```


```
type GroupService interface {
         List(...)  
         Create(...) // calls of a bunch of repository methods from roles/resources/identities
         AssignRole(..) // calls `RoleRepository` methods
         GetPermissions(..)
}
```

I haven't thought it out entirely but the above design is something we need to come up with - we could put them in the same/different file/package based on the dependencies :)

sbose78: Looks like we commented at the same time :) I agree with your *entire* comment and the details could best be picked/improvised while implementing.

> One last thought - placing these in the account package doesn't seem intuitive to me, as these features aren't really account-related. It is a pity that the "account" package isn't called "identity" instead, then it would make more sense to put them there. Do you have any suggestions as to the best place to locate these services?

Was wondering so too.. let me think.

sbose78: Maybe under sub packages in `authorization` ? :)

sbryzak: Ah, how about the `authorization` package itself?

sbose78: Yup, should be fine :)

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-20`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-21`

main.go
```diff
@@ -242,6 +243,15 @@ func main() {
 	clustersCtrl := controller.NewClustersController(service, config)
 	app.MountClustersController(service, clustersCtrl)
 
+	// Mount "resources" controller
+	resourcesCtrl := controller.NewResourceController(service, appDB)
+	app.MountResourceController(service, resourcesCtrl)
+
+	// Mount "organizations" controller
+	organizationService := model.NewOrganizationModelService(db, appDB)
+	organizationsCtrl := controller.NewOrganizationController(service, appDB, organizationService)
```
sbose78: Slight change here
```
organizationModelService := model.NewOrganizationModelService(...)
organizationService := model.NewOrganizationService(organizationModelService)
organizationCtrl := model.NewOrganizationController(service,appDB,organizationService)
```

Isn't this what we discussed :) ? had kinda outlined so in https://gist.github.com/sbose78/4bb7d589014ee896958956981c84f1a7
@alexeykazakov @sbryzak

sbryzak:
@sbose78 Hopefully I have the design correct now for the new model service layer.  In making these latest changes another import cycle got introduced, so I had to work around it by creating a "common" package for the structs/consts used across the API.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-22`

controller/organization.go
```diff
@@ -0,0 +1,112 @@
...
+	currentUser, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+
+	var orgs []common.IdentityOrganization
+
+	err = application.Transactional(c.db, func(appl application.Application) error {
+
+		orgs, err = c.orgService.ListOrganizations(ctx, *currentUser)
```
sbose78: A bit uncomfortable seeing that we using a pointer before checking if it's not nil :)

sbryzak: Good point, I've added a check that it's not nil.  I also removed the Transactional() block as it's done inside the service.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-23`

authorization/model/organization_model_service.go
```diff
@@ -0,0 +1,264 @@
...
+		ResourceID: res.ResourceID,
+		RoleID:     ownerRole.RoleID,
+	}
+
+	err = s.repo.IdentityRoleRepository().Create(ctx, identityRole)
+	if err != nil {
+		return nil, errors.NewInternalError(ctx, err)
+	}
+
+	if err != nil {
```
sbose78: repetitive,  not needed :)

sbryzak: Fixed.

authorization/model/organization_model_service_blackbox_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+type organizationServiceBlackBoxTest struct {
+	gormtestsupport.DBTestSuite
+	repo             resource.ResourceRepository
+	identityRepo     account.IdentityRepository
+	identityRoleRepo role.IdentityRoleRepository
+	resourceRepo     resource.ResourceRepository
+	resourceTypeRepo resource.ResourceTypeRepository
+	roleRepo         role.RoleRepository
+	orgService       authorization.OrganizationService
```
sbose78: The name of the file is `organization_model_service_blackbox_test.go` but yours tests test the `OrganizationService` contract.

`organization_model_service_blackbox_test.go` should have the `OrganizationModelService` contracts tested.
`organization_service_blackbox_test.go` should have the `OrganizationService` contracts tested.

sbose78:
The design and code organization looks good, my final comment is around having both `OrganizationModel Service` and `OrganizationService` tests . They would be fairly same for now - but we still need to have them because of the very reasons we have the 2 interfaces :) 

Apart from the above, looks great to me!

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-24`

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-317-25`

authorization/model/organization_model_service_blackbox_test.go
sbryzak: I've renamed and reorganized these tests, there is some duplication between the service and the model service tests, however I suppose that it doesn't hurt anything.

Issue #316, opened 2018-02-02
OSiO SA Token: Che Host


sbose78:
This is part of the work to remove 'edit' privileges from the Che namespace in OpenShift Online

aslakknutsen:
Create a Che Host OSiO SA token.. :)

alexeykazakov:
Done.

PR #318, opened 2018-02-02
SA for Che

alexeykazakov:
This PR is just about adding rh-che SA to our dev-mode configuration.

Related to https://github.com/fabric8-services/fabric8-auth/issues/316


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-318-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-318-2`

sbose78:
Have the environment-side changes been made ?

alexeykazakov:
>Have the environment-side changes been made ?

Not yet. See houskeeping.
I added a check for expected service accounts in configuration. So, the deployment will fail if the configuration is missing any of expected service accounts.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-318-3`

configuration/configuration.go
```diff
@@ -260,6 +261,38 @@ func NewConfigurationData(mainConfigFile string, serviceAccountConfigFile string
...
+		"fabric8-jenkins-idler": true,
+		"fabric8-oso-proxy":     true,
+		"online-registration":   true,
+		"fabric8-notification":  true,
+		"rh-che":                true,
+	}
+	for _, sa := range c.sa {
+		if sa.Name == "" {
+			msg := "service account name is empty in service account config"
+			c.appendDefaultConfigErrorMessage(&msg)
```
sbose78: Can't we maintain an array ( named `missingServiceAccounts`  initialized as empty ) where we keep pushing the names which are missing and in the end return an error, if `len(missingServiceAccounts) != 0` .. instead of maintaining `notFoundServiceAccountNames` ?

alexeykazakov: We don't need an array or map to check the number of SAs :) The point is to check if all expected SAs are present in configuration.

sbose78: Understood, this is the only place ( others are example files ) where you can maintain actual service account names, and not call it an example :)

PR #322, opened 2018-02-06
Get assigned roles for a resource 

### LINKED ISSUE #289: Implement List Assigned Roles by Resource endpoint
This endpoint lists the roles that have been assigned for a specified resource.

See [this section](https://docs.google.com/document/d/1qmpH8Yex0yKmcTauPqVoGmquRiEbab1E8fFf90tEw3A/edit#heading=h.eal3oploym7i) of the detailed design document for more information.

```
GET /api/resource/KX3A-39WE/roles HTTP/1.1
Authorization: Bearer MHg3OUZEQkZBMjcx
```

```
HTTP/1.1 200 OK
...
[{
   "identifier":"9XH4-BN3Y"
   "role_id":"JA7X-PV9Z",
   "role_name":"collaborator",
   "assignee_id":"1111-2222-3333-4444-AAAA-BBBB",
   "assignee_name":"john_smith",
   "assignee_type":"user",
   "inherited":"false"
},
{
   "identifier":"9XH4-BN7J"
   "role_id":"P9A5-MX3H",
   "role_name":"administrator",
   "assignee_id":"9999-8888-7777-6666-ZZZZ-XXXX",
   "assignee_name":"operations_team",
   "assignee_type":"team",
   "inherited":"true",
   "inherited_from":"P3Y8-17WB"
}]
```

sbose78:
This API lists all the assigned roles for a specific resource. The URL pattern for the same is
`/api/resources/:id/roles/assigned`

- [x] Added `/api/resources/:id/roles/assigned` and tests.
- [x] Added a `DataConflictError` type and mapped it to 409 , with tests.
- [x] Updated repository for `role` and added tests for `DataConflictError` in roles.
- [x] renamed `model` to `models` 
- [x] add utility methods in `test/authorization.go`

Since I worked on this "List" API before the "Add" API, we'll need to create the data manually:

```
INSERT INTO "resource_type" ("resource_type_id","name") VALUES('8e8bc9e2-6a24-47f8-a98f-6c1043143390','openshift.io/some/resourcetype');
INSERT INTO resource("resource_id","resource_type_id","name") VALUES ('4c169372-3b8a-4014-8017-49867b8681a7','8e8bc9e2-6a24-47f8-a98f-6c1043143390','SpaceR');
INSERT INTO role("role_id","resource_type_id","name") VALUES ('6008c541-15da-4662-9641-448ded5422da','8e8bc9e2-6a24-47f8-a98f-6c1043143390','collab');
INSERT INTO identities("id","username","registration_completed","provider_type") VALUES ('eca0ff77-4a85-408a-ad8e-729f8bdb96d9','f2ede785-9836-4235-93a1-8f037a284f12','false','kc');
INSERT INTO identity_role("identity_role_id","identity_id","resource_id","role_id") VALUES ('1a2a6b3a-f636-4246-a718-b662292897ba','eca0ff77-4a85-408a-ad8e-729f8bdb96d9','4c169372-3b8a-4014-8017-49867b8681a7','6008c541-15da-4662-9641-448ded5422da');

```

```
GET http://localhost:8089/api/resources/4c169372-3b8a-4014-8017-49867b8681a7/roles/assigned
Authorization: Bearer <token>
```

The output should be something like:
```
{
  data: [
    {
      assignee_id: "cb567f71-b09c-4784-a692-9270a6679af8",
      assignee_type: "user",
      identifier: "eb5145eb-474c-4776-8dee-b22a542f77e4",
      inherited: false,
      role_name: "collab"
    }
  ]
}

```

Fixes #289


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-2`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-3`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-4`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-5`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-6`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-7`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-8`

sbose78:
@alexeykazakov  @sbryzak  let's not review it now - let me clean this PR up.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-9`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-10`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-11`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-12`

sbose78:
[test]

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-13`

sbryzak:
I'm wondering now if we should rename the `identifier` return parameter to` identity_role_id` instead, to reduce ambiguity.

alexeykazakov:
> I'm wondering now if we should rename the `identifier` return parameter to `identity_role_id` instead, to reduce ambiguity.

Yeah... `identity_role_id` is better than `identifier` but do we need it at all? Why would we want to include this ID into the response?

sbryzak:
I think it's useful if you want to delete that specified role.  I suppose as an alternative if you want to delete then you could simply pass the resource_id, assignee_id and role name and have the auth service query for the actual role record.  For that matter, we probably don't need role_id either, we could just use the role name.

alexeykazakov:
Yeah.. For me identity_role_id is an implementation detail. Resource server should care about resource (resource_id), user/identity (identity_id) and role (name or role_id?).

alexeykazakov:
Do you think we should move authorization/common/authorization_common.go to authorization/organization/common/authorization_common.go ?
Why not keep just one authorization/common/authorization_common.go?

alexeykazakov:
I see you are reorganizing packages. And moving all model code to separate packages:
/authorization
/authorization/organization/some_model_service.go
/authorization/assignment/some_model_service.go
/services.go
etc.
instead of having the original structure:
/authorization
/authorization/model/all_the_model_services_*.go
/all_services_*.go

Any particular reason for that (i.g. cycle dependencies)? I like the original structure more because it looks simpler. We won't have that much files there, will we? Plus if we have too many fine grained packages it can be actually harder to solve cycle dependencies.

sbryzak:
@alexeykazakov No I don't think the authorization/common types should be moved.  Already this file contains constants not related to organizations, and I am certain it will expand as we fill out the functionality for teams and security groups.  This file is important across layers - controllers, services and tests and I see no value in introducing any further subdivision to its types and constants.

alexeykazakov:
> No I don't think the authorization/common types should be moved.

@sbryzak, that was my point too. So, asking why @sbose78 wants to move it.

sbryzak:
@alexeykazakov I've also thought about roles, I think for simplicity we should just stick with using the role name, and hide our internal `role_id` values from the client.  We can force unique role names per resource type in the database (@sbose78 could you add a unique constraint on the `ROLE` table for `RESOURCE_TYPE_ID` and `NAME` if we don't already have one?)

authorization/assignment/role_assignment_model_service.go
```diff
@@ -0,0 +1,136 @@
...
+				RoleID: roleIDAsUUID,
+				Name:   roleName,
+			},
+		}
+		if parentResourceID != nil {
+			ir.Resource.ParentResourceID = parentResourceID
+		}
+		identityRoles = append(identityRoles, ir)
+	}
+	return identityRoles, nil
```
alexeykazakov: Are you returning all available identity_roles for all identities for this resource or I'm missing something?

controller/resource_roles.go
```diff
@@ -0,0 +1,87 @@
...
+func convertIdentityRoleToAppRoles(ctx context.Context, roles []role.IdentityRole) []*app.IdentityRolesData {
+	var rolesList []*app.IdentityRolesData
+	for _, r := range roles {
+		rolesList = append(rolesList, convertIdentityRoleToAppRole(ctx, r))
+	}
+	return rolesList
+}
+func convertIdentityRoleToAppRole(ctx context.Context, r role.IdentityRole) *app.IdentityRolesData {
+	inherited := false
+	if r.Resource.ParentResourceID != nil {
```
alexeykazakov: Or just `inherited := r.Resource.ParentResourceID != nil`

controller/resource_roles.go
```diff
@@ -0,0 +1,87 @@
...
+		}, "error retrieving list of roles for a specific resource")
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+	roleList := convertIdentityRoleToAppRoles(ctx, roles)
+	return ctx.OK(&app.Identityroles{
+		Data: roleList,
+	})
+}
+
+func convertRoleToAppRoles(ctx context.Context, roles []role.Role) []*app.RolesData {
```
alexeykazakov: It's not used anywhere, right?

sbose78:
> Do you think we should move authorization/common/authorization_common.go to authorization/organization/common/authorization_common.go ?

That's what I've done :)
I didn't want to have a `utils` like package here and eventually end up with `everything that doesn't fit elsewhere`. In this case, `IdentityOrganization` is a very well defined concept and should be inside `/organization`.

> Any particular reason for that (i.g. cycle dependencies)? I like the original structure more because it looks simpler. We won't have that much files there, will we? Plus if we have too many fine grained packages it can be actually harder to solve cycle dependencies.

Re: Re-org files/packages/dirs

Apart from the `controllers` package, in this codebase we didn't really group go files into generic packages.
Example, we didn't group the repository-related `go` files into `/repositories`, we have created them under `/account/` .
( same in `wit` repo, we have `area`,`iterations`  packages )

I wanted to keep it consistent that way and ensure we don't end up with a huge list of go files in the same package.  Plus, this makes it easier to fine code related to a specific concept.

authorization/assignment/role_assignment_model_service.go
sbose78: Yes. Umm. was it just supposed to be for a specific user? I had the github collaborators page in mind where I could see all of us :)

sbose78:
and I think it's easier to solve cyclic dependencies this way :)

I don't have too much of a strong opinion on this though. but I think the directory structure needed a bit of cleanup.

sbose78:
@sbryzak 
>  I've also thought about roles, I think for simplicity we should just stick with using the role name, and hide our internal role_id values from the client. We can force unique role names per resource type in the database (@sbose78 could you add a unique constraint on the ROLE table for RESOURCE_TYPE_ID and NAME if we don't already have one?)

Makes sense, will do this!

sbose78:
> No I don't think the authorization/common types should be moved. Already this file contains constants not related to organizations, 

Thanks @sbryzak , will move it out then.

controller/resource_roles.go
sbose78: fixed bd13f5a

alexeykazakov:
@sbose78 I see what you mean but TBH still prefer better separation by layers. Controller layer, service layer, DB layer, etc. IMO it's easier to isolate and solve decencies. It's OK to split each layer to sub-packages though but I don't think we should do it until we have too many go files.

authorization/assignment/role_assignment_model_service.go
alexeykazakov: From the doc (mentioned in the issue):
"This endpoint is used to list the roles that have been assigned to a **user or team** for a specific resource.  It includes roles that have been assigned explicitly for a resource, in addition to roles that have been inherited from a parent resource."

It should be another endpoint to list the space/area collaborators/members. Not this one if I understand correctly. I don't see much value in the endpoint which list all the identities-roles for the resource. As we agreed the identity-role relation is our implementation detail. We need it to track relations between identities and roles inside Auth.
@sbryzak please clarify if I'm wrong.

sbose78: but the sample response has different assignees .
```
HTTP/1.1 200 OK
...
[{
   "identifier":"9XH4-BN3Y"
   "role_id":"JA7X-PV9Z",
   "role_name":"collaborator",
   "assignee_id":"1111-2222-3333-4444-AAAA-BBBB",
   "assignee_name":"john_smith",
   "assignee_type":"user",
   "inherited":"false"
},
{
   "identifier":"9XH4-BN7J"
   "role_id":"P9A5-MX3H",
   "role_name":"administrator",
   "assignee_id":"9999-8888-7777-6666-ZZZZ-XXXX",
   "assignee_name":"operations_team",
   "assignee_type":"team",
   "inherited":"true",
   "inherited_from":"P3Y8-17WB"
}]
```
A bit confused here - 
`/resource/:resourceID/roles/members` , `/resource/:resourceID/roles/collaborators` seems like very useful endpoints to me similar to https://github.com/fabric8-services/fabric8-auth/settings/collaboration

Are you saying this PR should have an endpoint to retrieve list of roles associated with me , for a specific resource?

sbose78:
> prefer better separation by layers. Controller layer, service layer, DB layer, etc. IMO it's easier to isolate and solve decencies. It's OK to split each layer to sub-packages though but I don't think we should do it until we have too many go files.

Cool, will move all models to `authorization/models`.

authorization/assignment/role_assignment_model_service.go
alexeykazakov: /resource/:resourceID/roles/members , /resource/:resourceID/roles/collaborators seems to be useful to me too, but this endpoint is different.

> Are you saying this PR should have an endpoint to retrieve list of roles associated with me , for a specific resource?

Yes, this is how I understand it. So, the sample response looks wrong to me.
@sbryzak please clarify what you meant here :)

sbose78: > /resource/:resourceID/roles/members , /resource/:resourceID/roles/collaborators seems to be useful to me too, but this endpoint is different.

`/api/resources/:id/roles/assigned` is an endpoint which returns all users who have *any* role assignment irrespective of what the role is - similar to Github's `GET /repos/:owner/:repo/collaborators` https://developer.github.com/v3/repos/collaborators/



Are you implying that ` /resource/:resourceID/roles/members` , /resource/:resourceID/roles/collaborators` are endpoints which return users who are assigned the named role 'members', 'collaborators' respectively ?

alexeykazakov: OK, I read the doc again and I think I got confused the first time.
I think you are right and this endpoint is just about listing all users and teams which have any role in the resource.

But now I have another concern... what if some user have multiple roles (because of inherited resources)? Will we list all his/her roles? Shouldn’t we return the most powerful role only, so, there is only one entry per user/team/group? We could resolve it by analyzing the scopes of overlapping resources.

sbose78: Cool. 

> But now I have another concern... what if some user have multiple roles (because of inherited resources)? Will we list all his/her roles? 

Ah yeah, this API consciously does that I think, because we have the `inherited:true` and `parent_resource_id` in the response

How do we determine if the scope is the most powerful ? Let me read the doc again..

alexeykazakov: We don't know what scope is most powerful because we don't know what the scope mean. It's up to resource service to interpret them. But we could try to resolve the most powerful **role** by analyzing the super/sub set of scopes from these roles... It can be tricky though. @sbryzak WDYT?

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-15`

sbose78:
>  've also thought about roles, I think for simplicity we should just stick with using the role name, and hide our internal role_id values from the client. We can force unique role names per resource type in the database (@sbose78 could you add a unique constraint on the ROLE table for RESOURCE_TYPE_ID and NAME if we don't already have one?

we already have this: 

`CREATE UNIQUE INDEX uq_role_resource_type_name ON role (resource_type_id, name);`

https://github.com/fabric8-services/fabric8-auth/blob/d07d56b843f7408d3c3fe338635d4d49704f866e/migration/sql-files/021-organizations-list-create.sql#L6

sbryzak:
I just caught up with the conversation thread via e-mail (although I can't see the comments here now).  This endpoint is important when you wish to view the access levels for a specific resource.  It shows any roles assigned to users, groups and teams for that resource directly, and also shows the roles that are inherited as a result of the resource being a child resource of another resource that has privileges assigned.  Think sub-areas within a space; there may be specific roles assigned to a user/team/group at the space level, all areas under the same space will automatically inherit the same role assignments, plus any additional ones assigned directly for the area itself.  This is useful for a resource owner for determining who has been granted privileges (either directly or indirectly) to act on that resource.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-16`

sbose78:
Added check for unique constraint in role names and added specific error type : 

https://github.com/fabric8-services/fabric8-
auth/pull/322/commits/a7d6156b958700e4be15f7b04703aab3d96cc2a2#diff-76f758602c8a7d4581a1e3e77b1bda3fR143

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-17`

authorization/assignment/role_assignment_model_service.go
```diff
@@ -0,0 +1,136 @@
...
+		if err = rows.Scan(columnValues...); err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"resource_id": resourceID,
+				"err":         err,
+			}, "error getting rows")
+			return identityRoles, errors.NewInternalError(ctx, err)
+		}
+		identityRoleIDAsUUID, _ := uuid.FromString(identityRoleID)
+		identityIDAsUUID, _ := uuid.FromString(identityID)
+		roleIDAsUUID, _ := uuid.FromString(roleID)
```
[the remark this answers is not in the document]
sbose78: fixed.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-18`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-19`

sbryzak:
@sbose78 I read the Go style guidelines you linked, thanks for that :)  After thinking about it a bit, we could probably rename the authorization/common package to authorization/types instead.  This is a more accurate and less-generalized description of what that package contains.

authorization/role_assignment_service.go
```diff
@@ -0,0 +1,47 @@
...
+	return &RoleAssignmentServiceImpl{modelService: modelService, db: db}
+}
+
+// ListByResource lists assignments made for a specific resource
+func (r *RoleAssignmentServiceImpl) ListByResource(ctx context.Context, resourceID string) ([]role.IdentityRole, error) {
+
+	var roles []role.IdentityRole
+	var err error
+	err = application.Transactional(r.db, func(appl application.Application) error {
+		err = appl.ResourceRepository().CheckExists(ctx, resourceID)
```
alexeykazakov: Why checking if the resource exists? Want to check if the resource exists before executing an expensive list call?

authorization/role_assignment_service.go
```diff
@@ -0,0 +1,47 @@
...
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"resource_id": resourceID,
```
alexeykazakov: Log err.

controller/resource_roles_blackbox_test.go
```diff
@@ -0,0 +1,152 @@
...
+	require.NoError(rest.T(), err)
+	require.NotNil(rest.T(), identityRoleRef2)
+	createdIdentityRoles = append(createdIdentityRoles, *identityRoleRef2)
+
+	svc, ctrl := rest.SecuredControllerWithIdentity(testsupport.TestIdentity)
+	_, returnedIdentityRoles := test.ListAssignedResourceRolesOK(rest.T(), rest.Ctx, svc, ctrl, resourceRef.ResourceID)
+	require.Len(rest.T(), returnedIdentityRoles.Data, 2)
+	require.True(rest.T(), rest.checkExists(*identityRoleRef, returnedIdentityRoles, false))
+	require.True(rest.T(), rest.checkExists(*identityRoleRef2, returnedIdentityRoles, false))
+}
```
alexeykazakov: In any layer (controller, service, model service) create more than expected assigned roles in tests. Check that only expected roles are returned. Just to make sure we do not return not relevant roles. I would probably put most covering tests to the service because the service may be re-used in different controllers. But I'm not sure... Don't want to duplicate all tests in all layers... But it's OK to have some overlap in the tests in different layers.

jsonapi/jsonapi_utility_blackbox_test.go
```diff
@@ -62,6 +62,14 @@ func TestErrorToJSONAPIError(t *testing.T) {
 	require.Equal(t, jsonapi.ErrorCodeUnauthorizedError, *jerr.Code)
 	require.Equal(t, strconv.Itoa(httpStatus), *jerr.Status)
 
+	// test unauthorized error
```
alexeykazakov: // test data conflict error

authorization/role_assignment_service.go
sbose78: Yes, correct.

controller/resource_roles_blackbox_test.go
sbose78: Added tests in controller, and a lot more detailed tests in the service layer https://github.com/fabric8-services/fabric8-auth/pull/322/commits/d70fc33c99735c506f1654d9c199e45615f05a28

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-20`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-21`

authorization/role_assignment_service_blackbox_test.go
```diff
@@ -0,0 +1,240 @@
...
+	require.NoError(s.T(), err)
+
+	var createdIdentityRoles []role.IdentityRole
+
+	// creating an AssignedRole for a parent resource
+	identityRoleRefUnrelated, err := testsupport.CreateTestIdentityRole(s.Ctx, s.DB, *parentResourceRef, *roleRef)
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), identityRoleRefUnrelated)
+
+	// creating an AssignedRole for an inherited resource
```
sbryzak: I suggest you refer to this as child resource, not inherited resource as _inherited_ is not really correct in this context (this applies to all similar references throughout this PR).

authorization/role_assignment_service_blackbox_test.go
```diff
@@ -36,10 +37,172 @@ func (s *roleAssignmentServiceBlackboxTest) TestGetIdentityRoleByResource() {
...
+	identityRoleRef1, err := testsupport.CreateTestIdentityRole(s.Ctx, s.DB, *resourceRef, *roleRef)
+	require.NoError(s.T(), err)
+	require.NotNil(s.T(), identityRoleRef1)
+	createdIdentityRoles = append(createdIdentityRoles, *identityRoleRef1)
+
+	// creating an AssignedRole for an inherited resource
+	identityRoleRef2, err := testsupport.CreateTestIdentityRole(s.Ctx, s.DB, *resourceRef, *roleRef)
```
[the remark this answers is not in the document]
sbryzak: Just to follow up on my previous question, I'm concerned that we allow the assignment of the same role to the same identity for the same resource.  

Line 81:

`identityRoleRef1, err := testsupport.CreateTestIdentityRole(s.Ctx, s.DB, *resourceRef, *roleRef)`

Line 87:

`identityRoleRef2, err := testsupport.CreateTestIdentityRole(s.Ctx, s.DB, *resourceRef, *roleRef)`

It appears they are using the same resource and role references.  This is why I asked previously which identity the role is being assigned to.  @sbose78 could you please clarify what is happening here?

authorization/role_assignment_service_blackbox_test.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (s *roleAssignmentServiceBlackboxTest) compare(createdRole role.IdentityRole, retrievedRole role.IdentityRole, isInherited bool) bool {
+	require.Equal(s.T(), createdRole.IdentityRoleID.String(), retrievedRole.IdentityRoleID.String())
+	require.Equal(s.T(), createdRole.IdentityID.String(), retrievedRole.Identity.ID.String())
+	require.Equal(s.T(), createdRole.Role.Name, retrievedRole.Role.Name)
+
+	if isInherited {
+		require.NotNil(s.T(), createdRole.Resource.ParentResourceID)
+		require.Equal(s.T(), *createdRole.Resource.ParentResourceID, *createdRole.Resource.ParentResourceID)
```
sbryzak: This looks like it is comparing exactly the same value, which will always be true?

authorization/role_assignment_service_blackbox_test.go
```diff
@@ -0,0 +1,240 @@
...
+	} else {
+		require.Nil(s.T(), retrievedRole.Resource.ParentResourceID)
```
sbryzak: A resource can have a parent resource, and yet be assigned a role directly.  What will happen in this case?

sbose78: Here's the test(s) for that https://github.com/fabric8-services/fabric8-auth/commit/d70fc33c99735c506f1654d9c199e45615f05a28#diff-0942c399cdebf10e464e3caf982b4642R44

authorization/models/role_assignment_model_service.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/fabric8-services/fabric8-auth/errors"
+	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/goadesign/goa"
+	"github.com/jinzhu/gorm"
+	uuid "github.com/satori/go.uuid"
+	"time"
+)
+
+// RoleAssignmentModelService defines the service contract for managing role assignments
+type RoleAssignmentModelService interface {
```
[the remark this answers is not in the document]
sbose78: @sbryzak , What other role-related functions do you think we could fit in here? 
I was planning to keep this service really small and 'targetted' to only handle role assignments. :)

authorization/role_assignment_service_blackbox_test.go
```diff
@@ -36,10 +37,172 @@ func (s *roleAssignmentServiceBlackboxTest) TestGetIdentityRoleByResource() {
...
+	err := testsupport.CreateTestIdentityAndUserInDB(s.DB, &resourceOwner)
+	require.NoError(s.T(), err)
+
+	areaResourceType, err := s.Application.ResourceTypeRepository().Lookup(s.Ctx, "openshift.io/resource/area")
+	require.NoError(s.T(), err)
+
+	parentResourceRef, err := testsupport.CreateTestResource(s.Ctx, s.DB, *areaResourceType, "AreaDev")
+	require.NoError(s.T(), err)
+
+	resourceRef, err := testsupport.CreateInheritedTestResource(s.Ctx, s.DB, *areaResourceType, "AreaAuth", *parentResourceRef)
```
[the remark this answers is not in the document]
sbose78: Sure, will refactor.

authorization/role_assignment_service_blackbox_test.go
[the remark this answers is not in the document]
sbose78: Thanks @sbryzak for raising the concern, Inside `testsupport.CreateTestIdentityRole..()` , a new Identity is created everytime the function is called.

This is what happens inside, hence we needn't worry - we are getting a unique identity everytime.

```
assignedIdentity := &account.Identity{
		ID:           uuid.NewV4(),
		Username:     uuid.NewV4().String(),
		ProviderType: account.KeycloakIDP,
	}
```

authorization/role_assignment_service_blackbox_test.go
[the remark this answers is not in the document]
sbose78: Done 135b415

design/role.go
```diff
@@ -0,0 +1,46 @@
...
+		a.Required("data")
+	})
+	a.View("default", func() {
+		a.Attribute("data")
+		a.Required("data")
+	})
+})
+
+var identityRolesData = a.Type("identityRolesData", func() {
+	a.Attribute("identifier", d.String, "Identifier for the owner of the resource")
```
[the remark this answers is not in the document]
sbose78: thanks - 135b4158504440d3d5ecfa9d661a14d351211306

design/role.go
```diff
@@ -0,0 +1,46 @@
...
+	a.Attribute("role_name", d.String, "The name of the role")
+	a.Attribute("assignee_id", d.String, "The ID of the assignee")
+	a.Attribute("assignee_type", d.String, "The type of assignee, example: user,groups,teams")
```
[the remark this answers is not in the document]
sbose78: Thanks! 135b4158504440d3d5ecfa9d661a14d351211306

test/authorization.go
```diff
@@ -0,0 +1,141 @@
...
+		ResourceType:   *resourceType,
+		ResourceTypeID: resourceType.ResourceTypeID,
+		Name:           name,
+	}
+	roleRepository := role.NewRoleRepository(db)
+	err = roleRepository.Create(ctx, &roleRef)
+	return &roleRef, err
+}
+
+func CreateInheritedTestResource(ctx context.Context, db *gorm.DB, resourceType resource.ResourceType, name string, parentResource resource.Resource) (*resource.Resource, error) {
```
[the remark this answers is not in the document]
sbose78: 9f6aab642d63612ae8fec7d10b161d2c56df2021 

Did it slightly different, to make it more explicit . Hope that's fine.

authorization/models/role_assignment_model_service_backbox_test.go
```diff
@@ -0,0 +1,50 @@
...
+func (s *roleAssignmentModelServiceBlackboxTest) TestGetIdentityRoleByResource() {
+	t := s.T()
+	identityRole, err := testsupport.CreateRandomIdentityRole(s.Ctx, s.DB)
+	require.NoError(t, err)
+	require.NotNil(t, identityRole)
+
+	identityRoles, err := s.repo.ListByResource(s.Ctx, identityRole.Resource.ResourceID)
+	require.NoError(t, err)
+	require.Equal(t, true, len(identityRoles) >= 1)
+	require.Equal(t, identityRole.Resource.ResourceID, identityRoles[0].Resource.ResourceID)
```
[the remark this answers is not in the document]
sbose78: Thanks, 4275cd6

authorization/role_assignment_service_blackbox_test.go
sbose78: 7b752e5

authorization/role_assignment_service_blackbox_test.go
sbose78: fixed b5a87de

authorization/role/identity_role.go
```diff
@@ -154,6 +158,109 @@ func (m *GormIdentityRoleRepository) Delete(ctx context.Context, id uuid.UUID) e
 	return nil
 }
 
+// ListAssignedRolesByResource returns all assigned roles associated with a specific resource
+func (m *GormIdentityRoleRepository) ListAssignedRolesByResource(ctx context.Context, resourceID string) ([]IdentityRole, error) {
+	defer goa.MeasureSince([]string{"goa", "db", "identity_role", "list"}, time.Now())
+	var identityRoles []IdentityRole
+
+	m.db = m.db.Debug()
```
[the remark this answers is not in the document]
sbose78: not there anymore, most have been an old commit :)

authorization/assignment/role_assignment_model_service.go
```diff
@@ -0,0 +1,136 @@
+package assignment
```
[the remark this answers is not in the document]
sbose78: will refactor as part of https://github.com/fabric8-services/fabric8-auth/issues/341

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-22`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-23`

authorization/role_assignment_service.go
sbose78: Good point @sbryzak , having this check helps us with the `NotFoundError` communication. What do you think.. is it worth it?

sbose78:
@sbryzak , all comments addressed - please re-review :)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-25`

authorization/models/role_assignment_model_service.go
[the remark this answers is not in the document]
sbryzak: Role revocation for starters, plus other query methods.

authorization/role_assignment_service.go
sbryzak: If you wish to return an error if the resource doesn't exist, then yes I think it's worth it.

authorization/role_assignment_service_blackbox_test.go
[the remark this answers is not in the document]
sbryzak: Ah that makes more sense now, thanks for explaining.

test/authorization.go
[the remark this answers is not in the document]
sbryzak: I'm fine with having two methods, but the naming still seems wrong to me.  You're not creating an inherited resource, you're creating a child resource.  The resource may _inherit_ some privileges, but the resource itself isn't _inherited_ by anything.

sbose78:
@sbryzak , couldn't find the comment box under your comments - not sure why :)

So here's my follow-up 
- renamed to `RoleManagementService` de7bc3b2b24c96fe71bb6ec97d2cbf38907af985
- removed the `CreateInheritedTestResource` : It was a miss on my side since it wasn't in use anyway. 6a2b3d397804722e6fa9b8918796860d872b1030

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-322-27`

PR #324, opened 2018-02-07
Hard delete state references

### LINKED ISSUE #319: Hard delete state references
I don't see any reason to soft delete state references in DB, so, let's hard delete them.

kishansagathiya:
Even though I am raising this PR, I am not quite sure if hard delete state reference is a good idea. It has information like response_mode, referrer and state. Probably we might not need this now, but from an analytics perspective hard deleting is not a good idea.

From top of my head what information i can think of is
1) Number of login requests
2) Which response_mode is being used more and how much more
3) Number of login requests by a particular client 

But I don't know if these kind of information is useful or will be useful in future.

Fixes #319


fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-324-1`

sbose78:
Any scope of adding a test here?

alexeykazakov:
> Probably we might not need this now, but from an analytics perspective hard deleting is not a good idea.

Good points @kishansagathiya

alexeykazakov:
I've put the PR on Hold for now.

sbose78:
Agreed. I'm never a fan of hard deleting anyway . Haha :)

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-324-3`

PR #326, opened 2018-02-09
Fixed inappropriate error handling if delete is unsuccessful

kishansagathiya:


sbose78:
Hi @kishansagathiya , why do think it is inappropriate ?

token/oauth/oauth2.go
```diff
@@ -136,13 +136,17 @@ func LoadReferrerAndResponseMode(ctx context.Context, db application.DB, state s
...
+			return err
+		}
+
+		return nil
 	})
 	if err != nil {
-		log.Error(ctx, map[string]interface{}{
-			"state": state,
-			"err":   err,
-		}, "unable to delete oauth state reference")
```
kishansagathiya: @sbose78 
In this transaction either delete or load can fail, but because of this line regardless of what fails, it always prints "unable to delete oauth state reference"

alexeykazakov: That's OK. Because we failed to delete the reference. The reason for this can be an error during loading. The err field will give you details of the reason.

kishansagathiya: Prime purpose of this transaction is to load the reference and not to delete it. We are deleting because once we have used the reference we don't need it anymore.

fabric8cd:
@kishansagathiya snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-326-2`

kishansagathiya:
> Still don't see much value in so detailed error log message but OK :)

@alexeykazakov  Completely forgot to close this PR, didn't have to merge it :)

PR #331, opened 2018-02-13
WIP More tests for creating and listing orgs

alexeykazakov:
This is a followup on https://github.com/fabric8-services/fabric8-auth/pull/317

- [x] Check org name when testing GORMOrganizationModelService.CreateOrganization()
- [x] Check owner role assigned to the org creator when testing GORMOrganizationModelService.CreateOrganization().
- [x] Create multiple orgs and check that only the relevant ones are returned by GORMOrganizationModelService.ListOrganizations() 
- [x] Add JWT security to org endpoints (**!!!**)
- [x] Log errors in org controller
- [x] Refactor controller tests, so they use user/identity instead of Service Accounts. Plus add tests for 401 responses.

### DIFF
--- a/authorization/model/organization_model_service_blackbox_test.go
+++ b/authorization/model/organization_model_service_blackbox_test.go
@@ -74,6 +74,10 @@ func (s *organizationModelServiceBlackBoxTest) TestCreateOrganization() {
 	require.Nil(s.T(), err, "Could not load the organization's resource")
 
 	require.Equal(s.T(), common.IdentityResourceTypeOrganization, orgResource.ResourceType.Name, "Organization resource type is invalid")
+
+	require.Equal(s.T(), orgResource.Name, "Test Organization ZXYAAA")
+
+	// TODO Check the owner role assigned to the creator
 }
 
 func (s *organizationModelServiceBlackBoxTest) TestListOrganization() {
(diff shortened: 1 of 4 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-331-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-331-3`

alexeykazakov:
There is something wrong with our Org Controller and its test.
Do we want to allow creating and listing orgs to end users using this API directly? If yes, then our tests are wrong because they use service accounts to call these endpoints. There is also missing tests for 401 responses when listing orgs.

If we do not allow creating and/or listing orgs to the end users and these API should be used only by the resources servers (do we really want this?) then our controller handles it wrong.

@sbryzak @sbose78

sbryzak:
@alexeykazakov Good catch, users definitely need to access these endpoints directly, not service accounts, as they query the organizations based on the user in the access token.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-331-5`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-331-6`

PR #344, opened 2018-02-21
Update pkg structure for authorization

### LINKED ISSUE #341: Refactor package structure in authorization package
Recommendation for model, service and repository:
```	
	authorization/assignee/service/assignee.go
	authorization/assignee/model/assignee.go
	authorization/assignee/name.go
        authorization/assignee/convert.go 


	authorization/role/repository/role.go
	authorization/role/model/role.go
	authorization/role/response_mapper.go
```

sbose78:
auth-dir

fixes #341


sbose78:
`make build` is green , with no issues or special workarounds.

authorization/group/type.go
```diff
@@ -0,0 +1,5 @@
+package group
+
+const (
+	IdentityResourceTypeGroup = "identity/group"
```
sbryzak: I was thinking it might be best to have a single authorization/constants type which contains all these, otherwise it's going to get real tedious importing every single one from a different location.

sbose78: I could do a `authorization/constants/constants.py`  and import that everywhere.
Since these are just string constants and will not contain any `struct` definitions ( thereby no business/service/model logic ).

What do you think?

sbryzak: Is there a point to putting it in its own directory when there's just one source file?  I would think it makes more sense keeping it in the authorization package, especially since there won't be much else in that package besides this?

sbose78:
Sure I could do that!

On Feb 21, 2018 19:47, "Shane Bryzak" <notifications@github.com> wrote:

*@sbryzak* commented on this pull request.
------------------------------

In authorization/group/type.go
<https://github.com/fabric8-services/fabric8-auth/pull/344#discussion_r169649366>
:

> @@ -0,0 +1,5 @@
+package group
+
+const (
+	IdentityResourceTypeGroup = "identity/group"

Is there a point to putting it in its own directory when there's just one
source file? I would think it makes more sense keeping it in the
authorization package, especially since there won't be much else in that
package besides this?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/fabric8-services/fabric8-auth/pull/344#discussion_r169649366>,
or mute the thread

authorization/repositories/repositories.go
```diff
@@ -2,15 +2,17 @@ package repositories
```
alexeykazakov: Should be `package repository`. It's not recommended to use plurals in package names in go.

authorization/doc.go
```diff
@@ -0,0 +1 @@
+package authorization
```
alexeykazakov: Add some package description

authorization/group/doc.go
```diff
@@ -0,0 +1 @@
+package group
```
alexeykazakov: Add some package description

authorization/group/type.go
alexeykazakov: If there is some "local" constant then it should be in the corresponding sub package like /authorization/group/group.go
But it makes sense to keep the shared/common constants like *TypeGroup in one place. And I would use /authorization/authorization.go for such constants. Not constants.go.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-344-4`

authorization/doc.go
sbose78: done.

authorization/group/doc.go
sbose78: done.

sbose78:
- fixed tests
- Review comments address
- updated tree image in description

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-344-5`

PR #351, opened 2018-02-27
Support force_pull for /api/token when retrieving cluster tokens

alexeykazakov:
Currently we support force_pull param in /api/token?for=<oso_resource> endpoint when retrieving user OSO tokens.
This PR also adds support of the same force_pull param for cluster tokens too. Not only for user tokens.
When force_pull=true the auth service will load the user profile using the stored dsaas user token. It will return 401 if the token is not valid and rejected by OSO.

Required for https://github.com/openshiftio/openshift.io/issues/2304

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -242,20 +242,10 @@ func (c *TokenController) retrieveToken(ctx context.Context, forResource string,
 		return nil, nil, err
 	}
 
-	osConfig, ok := providerConfig.(*link.OpenShiftIdentityProvider)
+	osConfig, ok := providerConfig.(link.OpenShiftIdentityProviderConfig)
 	if ok && token.IsSpecificServiceAccount(ctx, token.OsoProxy, token.Tenant) {
 		// This is a request from OSO proxy or tenant service to obtain a cluster wide token
-		clusterToken := app.ExternalToken{
-			Scope:          "<unknown>",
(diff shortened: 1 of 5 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-351-1`

controller/token.go
```diff
@@ -302,6 +292,41 @@ func (c *TokenController) retrieveToken(ctx context.Context, forResource string,
...
+			}, "unable to fetch user profile for cluster token")
+			errorResponse := fmt.Sprintf("LINK description=\"%s cluster token is not valid or expired", osConfig.OSOCluster().APIURL)
+			return nil, &errorResponse, errors.NewUnauthorizedError(err.Error())
+		}
+		if osConfig.OSOCluster().ServiceAccountUsername != userProfile.Username {
+			log.Warn(ctx, map[string]interface{}{
+				"for": forResource,
+				"configuration_username": osConfig.OSOCluster().ServiceAccountUsername,
+				"user_profile_username":  userProfile.Username,
+			}, "username from user profile for cluster token does not match username stored in configuration")
```
sbose78: this is going to be a tricky situation, if this ever happens :)

Issue #353, opened 2018-02-27
API for notifying Auth about deprovisioned users


alexeykazakov:
When OSO is deprovisioning a user it should notify OSIO. We should provide API on the OSIO Auth side for this.

Part of https://github.com/openshiftio/openshift.io/issues/2393

alexeykazakov:
For now I think we could just add a field to the user/identity so we have the data. Later we can use it (call tenant service to deprovision users etc).

sbose78:
On the tenant side, the functionality to de-provision needs to be built too, right?
@xcoulon , any idea?

PR #354, opened 2018-02-28
API for notifying Auth about deprovisioned users

alexeykazakov:
- [x] New endpoint which accepts requests from OSO Reg App SA only:
```
PATCH /api/users:id
{
  "data": {
    "type": "identities",
    "attributes": {
      "deprovisioned": true
    }
}
```
Returns `200 OK` if succeed.
- [x] Login returns 401 if user has been deprovisioned
- [x] Token refresh returns 401 if user has been deprovisioned
- [x] `GET /api/user`returns 401 if user has been deprovisioned. Tenant init service is not called.
- [x] `GET /api/userinfo`returns 401 if user has been deprovisioned. Tenant init service is not called.
- [x] `GET /api/users/:id` returns 401 if user has been deprovisioned but only if called by Tenant SA. Still open for other clients, so, UI is not broken.
- [x] Tests

Fixes https://github.com/fabric8-services/fabric8-auth/issues/353


migration/sql-files/022-add-deprovisioned-to-identity.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE identities ADD COLUMN deprovisioned BOOLEAN;
```
sbose78: so postgres should default it to `false`, right?

account/identity.go
```diff
@@ -145,6 +148,22 @@ func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID) (*Ident
 	return &native, errs.WithStack(err)
 }
 
+// LoadWithUser loads an identity and the assosiated User
+// Returns NotFoundError if either identity or user is not found
+func (m *GormIdentityRepository) LoadWithUser(ctx context.Context, id uuid.UUID) (*Identity, error) {
```
sbose78: instead of doing this outside with `Query` everytime, this makes sense.

controller/users.go
```diff
@@ -690,6 +690,47 @@ func (c *UsersController) Update(ctx *app.UpdateUsersContext) error {
...
+	})
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err":         err,
+			"identity_id": ctx.ID,
+		}, "failed to load identity")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewNotFoundError("identity", ctx.ID))
+	}
+	// We support "deprovisioned" attribute update for now only
+	if ctx.Payload != nil && ctx.Payload.Data != nil && ctx.Payload.Data.Attributes != nil && ctx.Payload.Data.Attributes.Deprovisioned != nil {
```
sbose78: So, `deprovisioned` is the only attribute supported for updation using svc account.

migration/sql-files/022-add-deprovisioned-to-identity.sql
alexeykazakov: I would expect NULL. But that's OK.

design/account.go
```diff
@@ -212,7 +212,24 @@ var _ = a.Resource("users", func() {
...
+	})
 
+	a.Action("updateByServiceAccount", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.PATCH(":id"),
+		)
+		a.Description("update the user")
+		a.Params(func() {
+			a.Param("id", d.String, "id")
```
sbose78: Gets confusing so let's mention that it's the Identity ID.

migration/sql-files/022-add-deprovisioned-to-identity.sql
alexeykazakov: It will be false by default in account.Identity though.

controller/users.go
alexeykazakov: Yup. Don't want the SA to accidentally update something else :) We can support updating other fields later when we have a usecase for this.

migration/sql-files/022-add-deprovisioned-to-identity.sql
sbose78: Oh yeah, NULL https://www.postgresql.org/docs/9.1/static/datatype-boolean.html
Let's set it to false by default then. Will help us if we have to ever collect data from the db directly :)

design/account.go
alexeykazakov: Well... wanted to be consistent with GET /api/user/:id. We can update the description though: a.Param("id", d.String, "identity ID")

migration/sql-files/022-add-deprovisioned-to-identity.sql
alexeykazakov: I don't think it's a problem to construct a query if we want to collect all false/null or true records :)
Just don't want to add unnecessary data to the table.
@sbryzak maybe there are some good practices for such cases?

design/account.go
sbose78: yea that's what I meant.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-3`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-4`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-5`

alexeykazakov:
@sbose78 ready for review :)

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-6`

migration/sql-files/022-add-deprovisioned-to-identity.sql
sbose78: Do we want the NULL value to imply something like `we don't know if user was deprovisioned` ? If yes, let's keep it NULL. 
Or, if we plan to imply a NULL as `user has not been deprovisioned`, then let's make it `false`

sbose78: Btw, could you please add a migration test for this.

test/account.go
```diff
@@ -88,6 +88,16 @@ var TestTenantIdentity = account.Identity{
 	User:     TestUser,
 }
 
+// CreateLonelyTestIdentity creates an identity not assosiated with any user. For testing purpose only.
+func CreateLonelyTestIdentity(db *gorm.DB, username string) (account.Identity, error) {
```
sbose78: The other methods used to created 'lonely' identities until I 'coupled' them off in one of my PRs :D

migration/sql-files/022-add-deprovisioned-to-identity.sql
alexeykazakov: Currently it's not a big deal to update our table and set false for existing identities. We would like to avoid it when we have too big identity table/DB. So, again I still don't see any issue with treating NULL or FALSE as - not deprovisioned.
Basically if deprovisioned == true then the user is deprovisioned. If not then not. :) Our code works well and handles it properly.
But I will add this "false" assignment to the migration step this time to make our table looks "better" because we can afford it at this point.

alexeykazakov: Actually our identities are quite generic objects in our DB. They can represent remote identities (GitHub, JIRA, etc) or our internal identities (org, user, etc). Not for all of them "deprovisioned" makes sense. So, let's keep it nullable. I've added the migration tests though.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-7`

migration/sql-files/022-add-deprovisioned-to-identity.sql
sbose78: Sorry I didn't realize we had altered IDENTITY table. Why aren't we adding this column to USER table? Because that's where we store cluster information.

alexeykazakov: I was thinking about both options.. To store it in Users or in Identities. Identities seemed to be "closer" to the identity provider created by OSO... But I'm not sure I chose the right option :)

alexeykazakov: If we are going to support multiple clusters per user then it makes sense to keep deprovisioned in the identity. Maybe we will have to move cluster to the identity too. Not sure what's best for now though.

alexeykazakov:
@sbose78 I've moved deprovisioned from Identity to User. Ready for review.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-354-8`

PR #347, opened 2018-02-23
List roles by resource_type

sbose78:
fixes #288 

This PR implements the 
```
GET /api/roles?resource_type=NAME_OF_RESOURCE_TYPE

HTTP/1.1 200 OK
Content-Type: application/json
...
"data": [
  {
    "name": "collaborator",
    "resource_type": "openshift.io/resource/area",
    "scopes": [
      "manage",
      "view"
    ]
  },
  {
    "name": "administrator",
    "resource_type": "openshift.io/resource/area",
    "scopes": [
      "admin"
    ]
  }
]
```

Support for `GET /api/roles` will be added in a different PR.

- [x] Repository for Role/Scope associations in role_scope table
- [x] Repository tests
- [x] Model RoleManagementModel
- [x] Model tests
- [x] Service RoleManagementService
- [x] Service tests
- [x] Controller
- [x] Controller tests
- [x] Migrations / add indexes to columns
- [x] Migration tests

PR #355, opened 2018-02-28
ISSUE-338 initial commit for invitation management

### LINKED ISSUE #338: Implement create invitation endpoint
This endpoint is used to create invitations for users to join an organization, team or security group.

See [this section](https://docs.google.com/document/d/1qmpH8Yex0yKmcTauPqVoGmquRiEbab1E8fFf90tEw3A/edit#heading=h.1ya0msjgkcyd) of the detailed design document for more information.

Sample request:

```
POST /api/invitations/da29bc55-4089-4c6a-95e8-826268e0d616
Content-type: application-json
Authorization: Bearer <token>

{
  "members":[
  {"user-id":"44c9d8b1-f687-4e7b-977d-d87071422f0d"},
  {"user-id":"57c3f642-2d50-4e0a-a071-e8fe480e3aa9"},
  {"user-email":"john_smith@acmecorporation.com"},
  {"user-name":"alice_jones"}]
  "roles":[
  {"user-id":"44c9d8b1-f687-4e7b-977d-d87071422f0d",
   "roles":["admin"]},
  {"user-id":"57c3f642-2d50-4e0a-a071-e8fe480e3aa9",
   "roles":["manager"]}]
}
```

Response:

```
HTTP/1.1 201 Created
Content-Type: application-json

```

sbryzak:
This pull request implements the create invitation endpoint and invitation service.  The scope of the invitations API has been expanded since starting work on this issue to allow for creating invitations for users to accept a role for a resource, this functionality will be covered by a different issue.

Request to invite a user to an organization/team/security group:

```
POST /invitations/<identityID>
or
POST /invitations/<resourceID>

{  "data": [
  {  "identity-id": "aaaa-bbbb-cccc-dddd-1234-5678",
     "member": "true"},
  {  "identity-id": "aaaa-bbbb-cccc-dddd-1234-9999",
      "member": "true"},
  {  "identity-id": "aaaa-bbbb-cccc-dddd-1234-aaaa",
      "member": "false",
      "roles": [ "owner" ]
  }
}
```
On success, returns a `201 Created` response

Fixes #338


sbryzak:
Still a lot more implementation to do, plus testing.  Just wanted to push what I've done so far for review as we're up to 750 lines already and just really getting started on this PR. ;)

account/identity.go
```diff
@@ -312,6 +329,13 @@ func IdentityWithUser() func(db *gorm.DB) *gorm.DB {
 	}
 }
 
+// IdentithWithResource is a gorm filter for preloading the IdentityResource relationship
+func IdentityWithResource() func(db *gorm.DB) *gorm.DB {
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Preload("IdentityResource")
```
sbose78: Need to check if this is a different query :)

authorization/invitation/invitation.go
```diff
@@ -0,0 +1,13 @@
+package invitation
+
+import (
+	uuid "github.com/satori/go.uuid"
+)
+
+type Invitation struct {
```
sbose78: Nice, I've done something similar too here https://github.com/fabric8-services/fabric8-auth/pull/347/files#diff-bd9af08e13589624d0629a20880b3c70 because of the difference between the entity in the db, and the response json.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,166 @@
...
+	invRepo "github.com/fabric8-services/fabric8-auth/authorization/invitation/repository"
+	permissionModel "github.com/fabric8-services/fabric8-auth/authorization/permission/model"
+	"github.com/fabric8-services/fabric8-auth/authorization/repository"
+	"github.com/fabric8-services/fabric8-auth/errors"
+	"github.com/jinzhu/gorm"
+	uuid "github.com/satori/go.uuid"
+)
+
+type InvitationModelService interface {
+	CreateInvitations(ctx context.Context, issuingUserId uuid.UUID, inviteTo uuid.UUID, invitations []invitation.Invitation) error
```
sbose78: Could just use `Create` and `List` ? to avoid stutter :) https://blog.golang.org/package-names

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,166 @@
...
+	ListInvitations(ctx context.Context, id uuid.UUID) ([]invitation.Invitation, error)
+}
+
+// GormInvitationModelService is the implementation of the interface for
+// InvitationService. IMPORTANT NOTE: Transaction control is not provided by this service
```
sbose78: nice note.

migration/sql-files/022-invitations.sql
```diff
@@ -0,0 +1,35 @@
...
+  member boolean NOT NULL,
+  created_at timestamp with time zone,
+  updated_at timestamp with time zone,
+  deleted_at timestamp with time zone
+);
+
+create table invitation_role (
+  invitation_id uuid NOT NULL references invitation (invitation_id),
+  role_id uuid NOT NULL references role (role_id)
+  PRIMARY KEY (invitation_id, role_id)
```
sbose78: Could we have an ID column which is an auto-generated uuid - and then make that the primary key.
We can make `(invitation_id, role_id)` a unique key.

Apart from general consistency with all other tables, this ensures that the database row cleanup logic doesn't break  https://github.com/fabric8-services/fabric8-auth/blob/master/gormsupport/cleaner/db_clean.go 
This helps the test data cleanup to happen 'automatically' using the primary key of the entity.

We don't need to use the ID anywhere in the code - apart from the repository struct.

design/invitation.go
```diff
@@ -0,0 +1,58 @@
...
+		//a.Response(d.OK, invitationArray)
+		a.Response(d.InternalServerError, JSONAPIErrors)
+		a.Response(d.BadRequest, JSONAPIErrors)
+	})
+})
+
+var CreateInvitationRequestMedia = a.MediaType("application/vnd.create_invitation_request+json", func() {
+	a.Description("Request payload required to create new invitations")
+	a.Attributes(func() {
+		a.Attribute("invitees", a.ArrayOf(invitee), "An array of users invited to become members or to accept a role")
```
sbose78: Could we please 'nest' it under `data` ? That'll help us add metadata in future without breaking any of the UI/clients.

sbryzak:
Tests are looking pretty good now, I still need to add a few more (including controller tests) and we should be good for another review.  Will aim to get this completed by the end of the weekend

account/identity.go
sbryzak: I've removed this function, gorm can't preload IdentityResource (and won't tell me why)

authorization/invitation/model/invitation_model_service.go
sbryzak: Agreed, however I've renamed these methods differently now to account for different workflows.

design/invitation.go
sbryzak: I've renamed `invitees` to `data`, will that work?

controller/invitation_blackbox_test.go
```diff
@@ -0,0 +1,89 @@
...
+	payload := &app.CreateGroupInviteInvitationPayload{
+		Data: []*app.Invitee{
+			{
+				Username: &testUsername,
+				Member:   true,
+			},
+		},
+	}
+
+	test.CreateGroupInviteInvitationCreated(s.T(), s.Ctx, service, controller, orgIdentity.ID.String(), payload)
```
alexeykazakov: Use service.Ctx instead of s.Ctx

sbryzak:
[test]

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-23`

design/invitation.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+var _ = a.Resource("invitation", func() {
+
+	a.BasePath("/invitations")
+
+	a.Action("createGroupInvite", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.POST("group/:inviteTo"),
```
alexeykazakov: Why POST `/invitations/group/:id` ? Why not just POST `/invitations/:id`?
/group/ is a bit confusing here IMO.

design/invitation.go
```diff
@@ -0,0 +1,59 @@
...
+	})
+
+	a.Action("list", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.GET(""),
+		)
+		a.Description("Lists invitations for an organization, team or group that the user has the invite_user scope for, or lists invitations for the current user")
+		a.Response(d.Unauthorized, JSONAPIErrors)
+		//a.Response(d.OK, invitationArray)
```
alexeykazakov: Let's remove this from now if we are not implementing it in this PR. Or at least return NotImplemented() for now.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-24`

design/invitation.go
sbryzak: URL has been updated to support both invitation types now.

design/invitation.go
sbryzak: controller returns NotImplemented() now

authorization/invitation/repository/invitation.go
```diff
@@ -0,0 +1,243 @@
...
+	rolerepo "github.com/fabric8-services/fabric8-auth/authorization/role/repository"
+	errs "github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
+)
+
+type Invitation struct {
+	gormsupport.Lifecycle
+
+	// This is the primary key value
+	InvitationID uuid.UUID `sql:"type:uuid default uuid_generate_v4()" gorm:"primary_key" gorm:"column:invitation_id"`
```
alexeykazakov: `gorm:"primary_key;column:invitation_id"`

authorization/invitation/repository/invitation.go
```diff
@@ -0,0 +1,243 @@
...
+	return "invitation"
+}
+
+// GetLastModified returns the last modification time
+func (m Invitation) GetLastModified() time.Time {
+	return m.UpdatedAt
+}
+
+type InvitationRole struct {
+	InvitationID uuid.UUID `sql:"type:uuid" gorm:"primary_key" gorm:"column:invitation_id"`
```
alexeykazakov: `gorm:"primary_key;column:invitation_id"`

authorization/invitation/repository/invitation.go
```diff
@@ -0,0 +1,243 @@
...
+
+	Role   rolerepo.Role `gorm:"ForeignKey:RoleID;AssociationForeignKey:RoleID"`
+	RoleID uuid.UUID     `sql:"type:uuid" gorm:"primary_key" gorm:"column:role_id"`
```
alexeykazakov: `gorm:"primary_key;column:role_id"`

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,47 @@
...
+func (s *InvitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UUID, inviteTo string, invitations []invitation.Invitation) error {
+
+	var err error
+
+	err = application.Transactional(s.db, func(appl application.Application) error {
+		err = s.invModelService.Issue(ctx, issuingUserId, inviteTo, invitations)
+		return err
+	})
+
+	return err
```
alexeykazakov: Add //TODO send emails with a link to accept the invitations

controller/invitation.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// NewInvitationController creates a invitation controller.
+func NewInvitationController(service *goa.Service, invitationService invitationService.InvitationService) *InvitationController {
+	return &InvitationController{Controller: service.NewController("InvitationController"), invService: invitationService}
+}
+
+// Create runs the create action.
+func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationContext) error {
+	currentUser, err := login.ContextIdentity(ctx)
```
alexeykazakov: Use login.ContextIdentityAndUserIfNotDeprovisioned()
and call it currentIdentity instead.

controller/invitation.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+
+	if currentUser == nil {
```
alexeykazakov: currentUser/Identity can't be nil here

design/invitation.go
```diff
@@ -0,0 +1,59 @@
...
+		a.Attribute("data")
+	})
+})
+
+var invitee = a.Type("Invitee", func() {
+	a.Attribute("identity-id", d.String, "unique id for the user identity")
+	a.Attribute("username", d.String, "username of the user")
+	a.Attribute("user-email", d.String, "e-mail address of the user")
+	a.Attribute("member", d.Boolean, "if true invites the user to become a member")
+	a.Attribute("roles", a.ArrayOf(d.String), "An array of role names")
```
alexeykazakov: roles should be required, right?

migration/sql-files/027-invitations.sql
```diff
@@ -0,0 +1,22 @@
...
+  CHECK ((invite_to IS NULL) <> (resource_id IS NULL));
+
+create table invitation_role (
+  invitation_id uuid NOT NULL references invitation (invitation_id),
+  role_id uuid NOT NULL references role (role_id),
+  PRIMARY KEY (invitation_id, role_id)
+);
+
+INSERT INTO resource_type_scope (resource_type_scope_id, resource_type_id, name) SELECT '87c2a3b2-6b7c-4d67-be4d-c73d5f51864b', resource_type_id, 'invite_user' FROM resource_type WHERE name = 'identity/organization';
+INSERT INTO role_scope (scope_id, role_id) SELECT '87c2a3b2-6b7c-4d67-be4d-c73d5f51864b', r.role_id FROM role r, resource_type rt WHERE r.resource_type_id = rt.resource_type_id AND r.name = 'owner' AND rt.name = 'identity/organization';
```
alexeykazakov: We need to keep all scope/role names in some common place so it's easy to track them.
Maybe in role package.

authorization/authorization.go
```diff
@@ -12,4 +12,10 @@ const (
 
 	// IdentityResourceTypeUser defines the string constant to be used for denoting a user managed by the authorization framework
 	IdentityResourceTypeUser = "identity/user"
+
+	// OwnerRole is the constant used to denote the name of the organization, team or security group owner role
+	OwnerRole = "owner"
+
+	// InviteUserScope is the scope required for users wishing to invite other users to an organization, team or security group
+	InviteUserScope = "invite_user"
```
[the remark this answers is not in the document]
sbryzak: I've renamed this scope to "manage_members"

authorization/invitation/invitation.go
```diff
@@ -0,0 +1,23 @@
...
+	IdentityID *uuid.UUID
+	UserEmail  *string
+	UserName   *string
+	Roles      []string
+	Member     bool
+}
+
+/* This struct is used to provide a user with details of the invitations they have received.
```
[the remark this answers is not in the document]
sbryzak: Comments have been updated.

authorization/invitation/invitation.go
```diff
@@ -0,0 +1,23 @@
...
+ * The ResourceType property indicates
```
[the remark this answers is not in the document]
sbryzak: Removed this line.

authorization/invitation/invitation.go
```diff
@@ -0,0 +1,23 @@
...
+ */
+type InvitationDetail struct {
+	ResourceType string
+	Description  string
+	Member       bool
```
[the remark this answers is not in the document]
sbryzak: Descriptions added for all properties.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/fabric8-services/fabric8-auth/authorization/repository"
+	resourceRepo "github.com/fabric8-services/fabric8-auth/authorization/resource/repository"
+	"github.com/fabric8-services/fabric8-auth/errors"
+	"github.com/jinzhu/gorm"
+	"github.com/satori/go.uuid"
+)
+
+type InvitationModelService interface {
+	Issue(ctx context.Context, issuingUserId uuid.UUID, inviteTo string, invitations []invitation.Invitation) error
+	List(ctx context.Context, id uuid.UUID) ([]invitation.Invitation, error)
```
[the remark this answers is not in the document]
sbryzak: I'm not sure how, or even if we will be using these methods yet based on our discussion of simply sending an invitation e-mail for now and not having a dedicated UI screen to display a user's outstanding invitations.  Based on this, I've removed both the List and ListForUser methods for now.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,221 @@
...
+	ListForUser(ctx context.Context, id uuid.UUID) ([]invitation.InvitationDetail, error)
```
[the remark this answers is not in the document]
sbryzak: Method has been removed

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,221 @@
...
+// NewInvitationModelService creates a new service.
+func NewInvitationModelService(db *gorm.DB, repo repository.Repositories, permissionSvc permissionModel.PermissionModelService) InvitationModelService {
+	return &GormInvitationModelService{
+		db:           db,
+		repo:         repo,
+		permModelSvc: permissionSvc,
+	}
+}
+
+// Creates new invitations.  The inviteTo parameter is the unique id of the organization, team or security group for
```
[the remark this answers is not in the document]
sbryzak: Fixed the comment and added more detail.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,221 @@
...
+			if err != nil {
+				return errors.NewInternalError(ctx, err)
+			}
+			// If there is anything other than 1 result found, we have a problem
+			if len(identities) == 0 {
+				// If no users are found, return an error
+				return errors.NewBadParameterErrorFromString("Username", invitation.UserName, "username not found")
+			} else if len(identities) > 1 {
+				// If more than one user is found, return an error
+				return errors.NewBadParameterErrorFromString("Username", invitation.UserName, "more than one user with same username found")
```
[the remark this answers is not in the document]
sbryzak: I've removed the len() > 1 check from here, let's discuss the other point separately.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,221 @@
...
+			} else if len(identities) > 1 {
+				return errors.NewBadParameterErrorFromString("E-mail", invitation.UserEmail, "more than one identity found for user with e-mail address")
+			}
+
+			// Set the IdentityID to that of the identity found
+			invitations[i].IdentityID = &identities[0].ID
+		}
+
+		// Confirm that any specified roles are valid for this resource type
+		for _, roleName := range invitation.Roles {
```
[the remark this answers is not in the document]
sbryzak: I've removed this section.  The point of it was to validate the role names before we started creating records, but I guess that since this entire thing is in a transaction nothing will get created if the transaction fails.

authorization/invitation/repository/invitation.go
sbryzak: Fixed

authorization/invitation/repository/invitation.go
sbryzak: Fixed

authorization/invitation/repository/invitation.go
sbryzak: Fixed

authorization/invitation/service/invitation_service.go
sbryzak: Done.

controller/invitation.go
sbryzak: I'm guessing you meant `LoadContextIdentityIfNotDeprovisioned()` - I've updated to use this function.

controller/invitation.go
sbryzak: Removed this check

design/invitation.go
sbryzak: No, roles aren't required to be assigned, the invitation can be for membership only.

migration/sql-files/027-invitations.sql
sbryzak: They should be in `authorization/authorization.go`, that's where I've been defining the constants.

sbryzak:
So there seems to be an issue here... after removing the role validation check the controller test is now failing at line 207:

`	require.Equal(s.T(), 0, len(invitations))`

This can only be due to invitation records actually being created, however the operation should have been cancelled and rolled back because the service returned an error, is my understanding correct here?

I can confirm that the error was returned because it's in the log:

`time="2018-04-14T12:46:32+07:00" level=error msg="failed to create invitations" err="Bad value for parameter 'Roles': 'foobar' - no such role found for resource type identity/organization" file="/home/shane/go/src/github.com/fabric8-services/fabric8-auth/controller/invitation.go" func="github.com/fabric8-services/fabric8-auth/controller.(*InvitationController).CreateInvite" identity_id=01e97fd3-dd09-4c65-8887-133d843cbe83 line=62 pid=15988 pkg=controller req_id= req_params=map[inviteTo:[05566841-2dcc-44ee-84ee-50a06c3eece2]] 
`

sbryzak:
Ok I found one issue, the controller test was creating the wrong service type (it was creating the model, not the service instance).  I've fixed that, and confirmed with the debugger that it is now calling the transaction-controlled service, however the transaction is still not being rolled back.  I've checked that the service method inside the transaction block is returning an error, and confirmed that it is, however I can query the database from the Postgres sql client and see that the record has still been created inside the table.  @alexeykazakov any ideas what might be happening here?

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,202 @@
...
+				resourceTypeName = inviteToResource.ResourceType.Name
+			}
+
+			role, error := s.repo.RoleRepository().Lookup(ctx, roleName, resourceTypeName)
+
+			if error != nil {
+				return errors.NewBadParameterErrorFromString("Roles", roleName, fmt.Sprintf("no such role found for resource type %s", resourceTypeName))
+			}
+
+			s.repo.InvitationRepository().AddRole(ctx, inv.InvitationID, role.RoleID)
```
alexeykazakov: Check if it returns any error

alexeykazakov:
@sbryzak the bug was in the way you use repos inside a transaction. See the fix - https://github.com/fabric8-services/fabric8-auth/pull/355/commits/01befa326a8965d4acdd0cf03ed9188199ac3bbe

It's required to use the same repo (from appl application.Application) which is passed to the transaction function. We should clean up other model services and repos too.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,201 @@
...
+		if invitation.IdentityID != nil {
+			identity, err := repo.Identities().Load(ctx, *invitation.IdentityID)
+			if err != nil {
+				return errors.NewInternalErrorFromString(ctx, fmt.Sprintf("invalid identity ID specified: %s", invitation.IdentityID))
+			}
+
+			if !identity.IsUser() {
+				return errors.NewBadParameterErrorFromString("Identity ID", invitation.IdentityID, "identity is not a user")
+			}
+		} else if invitation.UserName != nil {
```
alexeykazakov: From my previous review:
> We should not do any look up by usernames or emails here. Our invitation API should accept identity(user) IDs only. In UI we will search for user anyway. So UI will know the identity ID. We don't have to do this search twice.

So, why do you think our API should support invite by usernames or email? Why not Identity ID only?

controller/invitation.go
```diff
@@ -0,0 +1,80 @@
...
+// NewInvitationController creates a invitation controller.
+func NewInvitationController(service *goa.Service, db application.DB, invitationService invitationService.InvitationService) *InvitationController {
+	return &InvitationController{Controller: service.NewController("InvitationController"), db: db, invService: invitationService}
+}
+
+// Create runs the create action.
+func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationContext) error {
+	currentIdentity, err := login.LoadContextIdentityIfNotDeprovisioned(ctx, c.db)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
```
alexeykazakov: Don't need to wrap the error to errors.NewUnauthorizedError(err.Error())). 
login.LoadContextIdentityIfNotDeprovisioned(ctx, c.db) returns UnauthorizedError.
So, return jsonapi.JSONErrorResponse(ctx, err) is enough here.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,201 @@
...
+			// If the user's e-mail address has been provided, confirm the user is valid and that the identity is a user, and set the UserID
+			users, err := repo.Users().Query(account.UserFilterByEmail(*invitation.UserEmail))
+			if err != nil {
+				return errors.NewInternalError(ctx, err)
+			}
+			// We expect exactly 1 user to be found, if not we return an error
+			if len(users) == 0 {
+				return errors.NewBadParameterErrorFromString("E-mail", invitation.UserEmail, "user with e-mail address not found")
+			} else if len(users) > 1 {
+				return errors.NewBadParameterErrorFromString("E-mail", invitation.UserEmail, "more than one user with e-mail address found")
```
alexeykazakov: It can't happen. User emails are unique in the user table.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,201 @@
...
+
+			userID := &users[0].ID
+
+			// Now that we have the user ID, we can lookup the identity ID
+			identities, err := repo.Identities().Query(account.IdentityFilterByUserID(*userID))
+			// If there is anything other than 1 result found, return an error
+			if len(identities) == 0 {
+				return errors.NewBadParameterErrorFromString("E-mail", invitation.UserEmail, "no identity found for user with e-mail address")
+			} else if len(identities) > 1 {
+				return errors.NewBadParameterErrorFromString("E-mail", invitation.UserEmail, "more than one identity found for user with e-mail address")
```
alexeykazakov: It's not an error. We support multiple Identities per user. We don't use it right now though. This is for future use of remote identities (from GitHub, JIRA and other issue trackers).
You need to look up for identities with account.KeycloakIDP identity provider. Identities with that provider come from Keycloak/RHD and they represent OSIO users. There is supposed to be only one identity with such provider per user.
Feel free to add missing methods to the identity repo for that ;) But again, I don't think we should have look up for useremail at all in this API in the first place.

authorization/invitation/model/invitation_model_service_blackbox_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	err = s.invModelService.Issue(s.Ctx, s.Application, identity.ID, orgId.String(), invitations)
+	require.NoError(s.T(), err, "Error creating invitations")
+
+	invs, err := s.invitationRepo.ListForIdentity(s.Ctx, *orgId)
+	require.NoError(s.T(), err, "Error listing invitations")
+
+	require.Equal(s.T(), 1, len(invs))
+	require.True(s.T(), invs[0].Member)
```
alexeykazakov: Check all the relevant fields of the created invitation.

authorization/invitation/model/invitation_model_service_blackbox_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	err = s.invModelService.Issue(s.Ctx, s.Application, identity.ID, orgId.String(), invitations)
+	require.NoError(s.T(), err, "Error creating invitations")
+
+	invs, err := s.invitationRepo.ListForIdentity(s.Ctx, *orgId)
+	require.NoError(s.T(), err, "Error listing invitations")
+
+	require.Equal(s.T(), 1, len(invs))
+	require.Equal(s.T(), invitee.ID, invs[0].IdentityID)
+	require.True(s.T(), invs[0].Member)
```
alexeykazakov: Check all the relevant fields of the created invitation.

authorization/invitation/model/invitation_model_service_blackbox_test.go
```diff
@@ -0,0 +1,256 @@
...
```
alexeykazakov: Check all the relevant fields of the created invitation.

authorization/invitation/model/invitation_model_service_blackbox_test.go
```diff
@@ -0,0 +1,256 @@
...
+	require.NoError(s.T(), err, "Error listing invitations")
+
+	require.Equal(s.T(), 2, len(invs))
+
+	found := false
+
+	for _, inv := range invs {
+		if inv.IdentityID == invitee.ID {
+			found = true
+			require.True(s.T(), inv.Member)
```
alexeykazakov: Check all the relevant fields of the created invitation.

I think it's worth creating a function in this test which will assert the created invitation and use it all the relevant test functions.

authorization/invitation/repository/invitation_blackbox_test.go
```diff
@@ -0,0 +1,131 @@
...
+	_, err = s.repo.Load(s.Ctx, invitation.InvitationID)
+	require.NoError(s.T(), err)
+
+	require.False(s.T(), invitation.Member)
+
+	invitation.Member = true
+	err = s.repo.Save(s.Ctx, &invitation)
+	require.NoError(s.T(), err)
+
+	require.True(s.T(), invitation.Member)
```
alexeykazakov: Check all the relevant field of the saved invitation.

authorization/invitation/repository/invitation_blackbox_test.go
```diff
@@ -0,0 +1,131 @@
...
+func (s *invitationBlackBoxTest) TestAddRole() {
+	invitation, err := s.CreateTestInvitation()
+	require.NoError(s.T(), err)
+
+	roleRepository := roleRepo.NewRoleRepository(s.DB)
+	role, err := roleRepository.Lookup(s.Ctx, authorization.OwnerRole, authorization.IdentityResourceTypeOrganization)
+	require.NoError(s.T(), err)
+
+	err = s.repo.AddRole(s.Ctx, invitation.InvitationID, role.RoleID)
+	require.NoError(s.T(), err)
```
alexeykazakov: Check if the role was added or not.

authorization/invitation/repository/invitation_blackbox_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	roles, err := s.repo.ListRoles(s.Ctx, invitation.InvitationID)
+	require.NoError(s.T(), err)
+
+	require.Equal(s.T(), 1, len(roles))
+	require.Equal(s.T(), authorization.OwnerRole, roles[0].Name)
```
alexeykazakov: This test also tests AddRole(). It's OK to merge it to TestAddRole() test and have just one test TestAddAndListRoles() function which will cover both methods.

authorization/invitation/invitation.go
```diff
@@ -0,0 +1,26 @@
...
+	Roles      []string
+	Member     bool
+}
+
+// InvitationDetail is used to provide a user with details of the invitations they have received.
+type InvitationDetail struct {
+	// ResourceType is the name of the resource type to which the user has been issued an invitation
+	ResourceType string
+	// Description is a description of the organization, team, security group or resource to which the user has been issued an invitation
+	Description string
```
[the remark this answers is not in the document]
sbryzak: This struct is not actually required now since we will be sending an invitation e-mail, not displaying the invitations via UI screen.

authorization/invitation/model/invitation_model_service.go
```diff
@@ -0,0 +1,201 @@
...
+	var inviteToIdentity *account.Identity
+	var identityResource *resourceRepo.Resource
+	var inviteToResource *resourceRepo.Resource
+
+	// First try to convert inviteTo to a uuid
+	inviteToUUID, err := uuid.FromString(inviteTo)
+	// If we get an error here, the value is definitely not for an Identity so we'll treat it as a resource ID
+	if err != nil {
+		// Try to lookup a resource with the same ID value
+		inviteToResource, err = repo.ResourceRepository().Load(ctx, inviteTo)
```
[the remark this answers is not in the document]
sbryzak: Well, since the Resource ID is the primary key for the RESOURCE table, we already have a unique constraint the same as every other table's primary key.  Meaning that attempting to create a resource with an ID that already exists will fail with an error.

authorization/invitation/model/invitation_model_service.go
sbryzak: After our discussion I've removed support for username and e-mail.

controller/invitation.go
sbryzak: Fixed

authorization/invitation/model/invitation_model_service.go
sbryzak: This section has been removed

authorization/invitation/repository/invitation_blackbox_test.go
sbryzak: Fixed

authorization/invitation/repository/invitation_blackbox_test.go
sbryzak: Done

authorization/invitation/repository/invitation_blackbox_test.go
sbryzak: Done

sbryzak:
@alexeykazakov I've cleaned up the tests and hopefully improved coverage quite a bit.  I've refactored the service in line with the changes we are making in #446, we can use the other issue to address naming once we have completed our discussions on it.

sbryzak:
[test]

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-28`

migration/migration_blackbox_test.go
```diff
@@ -299,6 +301,56 @@ func testMigration25ValidMiss(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Confirm that we get a check constraint violation if we try to provide both invite_to and resource_id values
+	_, err = sqlDB.Exec("INSERT INTO invitation (invitation_id, invite_to, resource_id, identity_id, member) VALUES (uuid_generate_v4(), 'c62d77b2-194c-47d0-8bbf-b1308576876d', 'c6a2ee2e-7ec6-4c04-ae7e-5ff8c36b28b9', 'd9161547-5263-4c83-a729-e39ff088978e', false)")
+	require.NotNil(t, err)
+
+	// Cleanup the test data
+	require.Nil(t, runSQLscript(sqlDB, "026-cleanup-test-invitation-data.sql"))
```
sbose78: optional, but good practice - since we anyway drop this db where we test migrations. ( after this test suite is done executing )

account/identity.go
```diff
@@ -97,6 +97,10 @@ func (m Identity) GetLastModified() time.Time {
 	return m.UpdatedAt
 }
 
+func (m Identity) IsUser() bool {
```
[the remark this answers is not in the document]
sbryzak: Done.

authorization/authorization.go
```diff
@@ -12,4 +12,16 @@ const (
...
 	IdentityResourceTypeUser = "identity/user"
+
+	// OwnerRole is the constant used to denote the name of the organization, team or security group owner role
+	OwnerRole = "owner"
+
+	// ManageMembersScope is the scope required for users wishing to invite/remove other users to an organization, team or security group
+	ManageMembersScope = "manage_members"
 )
+
+func CanHaveMembers(resourceTypeName string) bool {
```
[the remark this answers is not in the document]
sbryzak: You won't believe this - this method had a bug which I only discovered when I wrote the test for it. ;)

authorization/invitation/repository/invitation.go
```diff
@@ -0,0 +1,243 @@
...
+func (m Invitation) TableName() string {
+	return "invitation"
+}
+
+// GetLastModified returns the last modification time
+func (m Invitation) GetLastModified() time.Time {
+	return m.UpdatedAt
+}
+
+type InvitationRole struct {
```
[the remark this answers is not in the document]

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-29`

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+				return errors.NewNotFoundError(fmt.Sprintf("invalid identifier '%s' provided for organization, team, security group or resource", inviteTo), inviteTo)
+			}
+		}
+	}
+
+	if inviteToIdentity != nil {
+		// Load the resource for the identity
+		identityResource, err = s.repo.ResourceRepository().Load(ctx, *inviteToIdentity.IdentityResourceID)
+		if err != nil {
+			return errors.NewInternalErrorFromString(ctx, "error loading resource for identity")
```
alexeykazakov: return errors.NewInternalError(ctx, err)

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		// Confirm that the issuing user has the necessary scope to manage members for the organization, team or security group
+		scope, err := s.permSvc.HasScope(ctx, issuingUserId, *inviteToIdentity.IdentityResourceID, authorization.ManageMembersScope)
+		if err != nil {
+			return errors.NewInternalError(ctx, err)
+		}
+
+		if !scope {
+			return errors.NewForbiddenError(fmt.Sprintf("user requires %s scope to invite other users", authorization.ManageMembersScope))
```
alexeykazakov: This is very important use case (when the user don't have permissions to invite) but it's not covered by tests - https://codecov.io/gh/fabric8-services/fabric8-auth/pull/355/diff#D28-79

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		// We only allow membership in some identity types - confirm that we are inviting to an organization, team or security group
+		if !authorization.CanHaveMembers(identityResource.ResourceType.Name) {
+			return errors.NewInternalErrorFromString(ctx, "may only invite a user to an organization, team or security group")
```
alexeykazakov: Not covered by tests.

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	} else if inviteToResource != nil {
+		// Confirm that the issuing user has the manage members scope for the resource
```
alexeykazakov: This whole case when we invite to a resource is not covered by tests.
There should be two tests for this:
1. User has permissions to invite to a resource
2. User doesn't have permissions.

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// Iterate through all of the invitations and confirm that for each one:
+	// 1) a valid user has been specified via its User ID, e-mail address or username
+	// 2) any roles specified are valid roles for the organization, team or security group
+	// For each invitation, ensure that the IdentityID value can be found and set it
+	for _, invitation := range invitations {
+		// Load the identity
+		identity, err := s.repo.Identities().Load(ctx, *invitation.IdentityID)
+		if err != nil {
+			return errors.NewInternalError(ctx, err)
```
alexeykazakov: No tests for this usecase (when we invite unknown user)

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		if !identity.IsUser() {
+			return errors.NewBadParameterErrorFromString("Identity ID", invitation.IdentityID, "identity is not a user")
```
alexeykazakov: No tests for this usecase (when the identity is not a user)

authorization/invitation/service/invitation_service.go
```diff
@@ -0,0 +1,154 @@
...
+			return errors.NewInternalError(ctx, err)
+		}
+
+		if !scope {
+			return errors.NewForbiddenError(fmt.Sprintf("user requires %s scope to invite other users", authorization.ManageMembersScope))
+		}
+	}
+
+	// Iterate through all of the invitations and confirm that for each one:
+	// 1) a valid user has been specified via its User ID, e-mail address or username
```
alexeykazakov: Fix the comment. We now use user identity ID only.

authorization/invitation/repository/invitation.go
```diff
@@ -0,0 +1,244 @@
...
+		SELECT EXISTS (
+			SELECT 1 FROM %[1]s
+			WHERE
+				invitation_id=$1
+				AND deleted_at IS NULL
+		)`, m.TableName())
+
+	err := m.db.CommonDB().QueryRow(query, id).Scan(&exists)
+	if err == nil && !exists {
+		return exists, errors.NewNotFoundError(m.TableName(), id.String())
```
alexeykazakov: Not covered by tests.

authorization/invitation/service/invitation_service.go
```diff
@@ -14,24 +20,135 @@ type InvitationService interface {
...
+func NewInvitationService(repositories repository.Repositories) InvitationService {
+	return &InvitationServiceImpl{repo: repositories, permSvc: repositories.PermissionService()}
 }
 
+// Issue creates new invitations.  The inviteTo parameter is the unique id of the organization, team, security group or resource for
+// which the invitations will be issued, and the invitations parameter contains the users and state for each individual user invitation.
+// This method creates one record in the INVITATION table for each user in the invitations parameter.  Any roles that are issued
+// as part of a user's invitation are created in the INVITATION_ROLE table.
 func (s *InvitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UUID, inviteTo string, invitations []invitation.Invitation) error {
+	var inviteToIdentity *account.Identity
```
alexeykazakov: You removed Model Service (need a better name) and moved its code (as-is) to the Controller Service.
This is not what we agreed to do yesterday. We can remove unnecessary layers but we can't mix them.
Controller Service must use transactions. But this service doesn't use them.
So I see two options here:
1. Move this code back to Model Service (need better name). Remove the Controller Service. Use Model Service directly from Controller. Transaction management should be in the controller then.
2. Wrap the whole thing inside this Issue() method in the Controller Service with a transaction. Remove transaction management from Controller.

IMO the first option is better.

authorization/permission/service/permission_service.go
```diff
@@ -31,7 +32,7 @@ func NewPermissionModelService(db *gorm.DB) PermissionModelService {
 // parent and other ancestor resources, and also takes into account role mappings, which allow roles assigned for a
 // certain type of resource in the resource ancestry to map to a role for a different resource type lower in the
 // resource hierarchy.
-func (s *GormPermissionModelService) HasScope(ctx context.Context, identityID uuid.UUID, resourceID string, scopeName string) (bool, error) {
+func (s *PermissionServiceImpl) HasScope(ctx context.Context, identityID uuid.UUID, resourceID string, scopeName string) (bool, error) {
```
alexeykazakov: Same thing. You are mixing Model Service (need a better name) with Controller Service.
You made this code a controller service. But then it must use transactions.
Just move this code back to the Model Service where it was before :)

controller/invitation.go
```diff
@@ -16,47 +15,47 @@ import (
...
-			*identityID, err = uuid.FromString(*invitee.IdentityID)
+			identityID, err = uuid.FromString(*invitee.IdentityID)
 		}
 
 		// Create the Invitation object, and append it to our list of invitations
 		invitations = append(invitations, invitation.Invitation{
-			IdentityID: identityID,
-			UserName:   invitee.Username,
-			UserEmail:  invitee.UserEmail,
+			IdentityID: &identityID,
```
alexeykazakov: IdentityID here is actually required, right? So, we can make it UUID (not *UUID).

gormapplication/application.go
```diff
@@ -97,6 +99,14 @@ func (g *GormBase) InvitationRepository() invitation.InvitationRepository {
 	return invitation.NewInvitationRepository(g.db)
 }
 
+func (g *GormBase) InvitationService() invitationservice.InvitationService {
+	return invitationservice.NewInvitationService(g)
+}
+
+func (g *GormBase) PermissionService() permissionservice.PermissionService {
+	return permissionservice.NewPermissionService(g.db)
+}
```
alexeykazakov: Both method should return Model Services (need a better name). Not Controller Services.
Model Services is purely a DB layer.
Controller Services is not. They may use DB layer (if they do they should provide transaction control). They may call remote services.

authorization/invitation/service/invitation_service.go
sbryzak: Done.

authorization/invitation/service/invitation_service.go
sbryzak: Test added.

authorization/invitation/service/invitation_service.go
sbryzak: Test added

authorization/invitation/service/invitation_service.go
sbryzak: I've added tests to cover both of these

authorization/invitation/service/invitation_service.go
sbryzak: Test added

authorization/invitation/repository/invitation.go
sbryzak: Test added

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-30`

authorization/invitation/service/invitation_service.go
sbryzak: Test added

authorization/invitation/service/invitation_service.go
sbryzak: Comment has been fixed

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-31`

gormapplication/application.go
sbryzak: I've renamed these for now to put "Model" back into the name.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-355-32`

PR #357, opened 2018-02-28
Add Logging URL to cluster information

stooke:
This PR adds a logging URL to go with the metrics and console URLs.

There will be an accompanying PR for the tenant service to make use of this URL in its APIs, too, and after that, the Deployments API in WIT will be updated to call the tenant to retrieve the console, metrics and logging URLs.


fabric8cd:
@stooke snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-357-1`

PR #371, opened 2018-03-09
API to trigger verification email

### LINKED ISSUE #369: Add API for trigger email for email verification code

sbose78:
Made this a `POST` request because it creates a new verification code and sends out an email.

```
POST /api/users/verificationcode
```

fixes #369

### DIFF
--- a/design/account.go
+++ b/design/account.go
@@ -163,6 +163,17 @@ var _ = a.Resource("users", func() {
 		a.Description("Verify if the new email updated by the user is a valid email")
 		a.Response(d.TemporaryRedirect)
 	})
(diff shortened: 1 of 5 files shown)


design/account.go
```diff
@@ -171,7 +171,6 @@ var _ = a.Resource("users", func() {
 		a.Description("Send a verification code to the user's email address")
 		a.Response(d.NoContent)
 		a.Response(d.InternalServerError, JSONAPIErrors)
-		a.Response(d.NotFound, JSONAPIErrors)
```
alexeykazakov: When there is unknown identity or user in the token then we should return 401. Not 404. So, I removed 404.

PR #373, opened 2018-03-12
Service respond with 500 on expired token. Should be 401 (OSIO#2471)

xcoulon:
Return a GOA error with status code `401 Unauthorized` when the token
is invalid or expired.


goamiddleware/jwt_token_context.go
```diff
@@ -33,7 +33,9 @@ func TokenContext(tokenManager token.Parser, scheme *goa.JWTSecurity) goa.Middle
 
 				token, err := tokenManager.Parse(ctx, incomingToken)
 				if err != nil {
-					return err
+					errUnauthorized := goa.NewErrorClass("validation_failed", 401)
```
alexeykazakov: No need to create a new error class every time.
Let's re-use goa.ErrUnauthorized class or we can create our own error class for the package:
`var errUnauthorized = goa.NewErrorClass("token_validation_failed", 401)`

xcoulon:
thanks, @alexeykazakov !

PR #380, opened 2018-03-14
Remove auth route from OS template

alexeykazakov:


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-380-1`

alexeykazakov:
This route is actually is used by WIT - https://github.com/fabric8-services/fabric8-wit/blob/master/configuration/configuration.go#L822
It's used by WIT->Auth proxy. The full URL is not working in proxy for some reason. This is why we use "auth" route in WIT.

alien-ike:
### Ike Plugins (test-keeper)

alexeykazakov:
It turned out that we don't need a route to reach auth internally. http://auth still will be availability internally and will point to the auth service. So, we can remove the route.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-380-2`

sbose78:
Nice - just saw the discussion too :)

Issue #377, opened 2018-03-14
Don't create preview OSO accounts in prod OSIO/Auth if they were provisioned for preview env


alexeykazakov:
Do not create/provision preview OSO accounts in prod OSIO/Auth if they were provisioned for preview env. No account sharing between prod and preview.

Part of https://github.com/openshiftio/openshift.io/issues/2348

PR #381, opened 2018-03-14
Don't create preview OSO accounts in prod OSIO/Auth if they were provisioned for preview env

alexeykazakov:
When OSO reg app creates a user with email == `<username>+preview*@redhat.com` then Auth ignores such user.
Such accounts are supposed to be used only in Preview. So there will be no account sharing between prod and preview.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/377

### DIFF
--- a/controller/users.go
+++ b/controller/users.go
@@ -25,6 +25,7 @@ import (
 	"github.com/jinzhu/gorm"
 	errs "github.com/pkg/errors"
 	"github.com/satori/go.uuid"
+	"regexp"
 )
 
 // UsersController implements the users resource.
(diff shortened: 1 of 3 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-381-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-381-2`

controller/users.go
```diff
@@ -104,6 +106,21 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
...
+	preview, err := c.checkPreviewUser(ctx.Payload.Data.Attributes.Email)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{"err": err, "email": ctx.Payload.Data.Attributes.Email}, "unable to parse user's email")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+	}
+	if preview {
+		log.Info(ctx, map[string]interface{}{"email": ctx.Payload.Data.Attributes.Email}, "ignoring preview user")
+		user := &account.User{Email: ctx.Payload.Data.Attributes.Email, Cluster: ctx.Payload.Data.Attributes.Cluster}
+		identity := &account.Identity{Username: ctx.Payload.Data.Attributes.Username, ProviderType: account.KeycloakIDP}
+		return ctx.OK(ConvertToAppUser(ctx.RequestData, user, identity, true))
```
sbose78: Good! exactly what I was thinking, we don't need to return alarms to the registration app.

Issue #387, opened 2018-03-16
Allow to search by public emails only.


alexeykazakov:
`/api/users?filter[email]=<email>` should return the result if the email is public.

PR #393, opened 2018-03-20
Don't return private emails in user search result

alexeykazakov:
`/api/users?filter[email]=<email>` and `/api/search/users?q=<email>` look up public emails only.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/387

### DIFF
--- a/account/identity.go
+++ b/account/identity.go
@@ -16,7 +16,7 @@ import (
 	"github.com/goadesign/goa"
 	"github.com/jinzhu/gorm"
 	errs "github.com/pkg/errors"
-	uuid "github.com/satori/go.uuid"
+	"github.com/satori/go.uuid"
 )
 
 const (
@@ -336,13 +336,6 @@ func IdentityFilterByProviderType(providerType string) func(db *gorm.DB) *gorm.D
 	}
 }
 
-// IdentityFilterByRegistrationCompleted is a gorm filter by 'registration_completed'
-func IdentityFilterByRegistrationCompleted(registrationCompleted bool) func(db *gorm.DB) *gorm.DB {
-	return func(db *gorm.DB) *gorm.DB {
-		return db.Where("registration_completed = ?", registrationCompleted)
-	}
-}
-
 // List return all user identities
 func (m *GormIdentityRepository) List(ctx context.Context) ([]Identity, error) {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "list"}, time.Now())
@@ -369,7 +362,8 @@ func (m *GormIdentityRepository) IsValid(ctx context.Context, id uuid.UUID) bool
 	return true
 }
 
-// Search searches for Identites where FullName like %q% or users.email like %q% or users.username like %q%
+// Search searches for Identities where FullName like %q% or users.email like %q% (but ignores private emails)
+// or users.username like %q%
 func (m *GormIdentityRepository) Search(ctx context.Context, q string, start int, limit int) ([]Identity, int, error) {
 
 	db := m.db.Model(&Identity{})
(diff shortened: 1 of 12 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-2`

account/identity.go
```diff
@@ -379,7 +373,7 @@ func (m *GormIdentityRepository) Search(ctx context.Context, q string, start int
 	db = db.Select("count(*) over () as cnt2 ,identities.id as identity_id,identities.username,users.*")
 	db = db.Joins("LEFT JOIN users ON identities.user_id = users.id")
 	db = db.Where("LOWER(users.full_name) like ?", "%"+strings.ToLower(q)+"%")
-	db = db.Or("users.email like ?", "%"+strings.ToLower(q)+"%")
+	db = db.Or("LOWER(users.email) like ? and users.email_private is false", "%"+strings.ToLower(q)+"%")
```
sbose78: nice!

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-3`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-4`

controller/users.go
```diff
@@ -1061,6 +1062,13 @@ func filterUsers(appl application.Application, ctx *app.ListUsersContext) ([]acc
 		// .. Add other filters in future when needed into the userFilters slice in the above manner.
 		if len(userFilters) != 0 {
 			filteredUsers, err = appl.Users().Query(userFilters...)
+			if err != nil {
+				return nil, nil, errs.Wrap(err, "error fetching users")
+			}
+			// Should be zero or the only user in the result. If the user's email is private then ignore the user
+			if len(filteredUsers) > 0 && filteredUsers[0].EmailPrivate {
```
sbose78: I think, it might be a lot cleaner if we introduced a 
```
account.UserFilterByEmailPrivacy(showPrivateEmails bool)
```
instead of adding custom logic. Do you think its worth it?

alexeykazakov:
> One of the objectives was to ensure that every controller mandatorily specify InternalServerError as one of the return types - am I correct ?

Yes. Well.. every "regular" controller :) Status controller is a special one.

> However, should we bother that the /api/status controller doesn't have a return type of InternalServerError ? There's really no scope to return InternalServerError though :)

Yeah... Status doesn't need to return 500 and it doesn't need to use JSONError. So, I think we are good.

controller/users.go
alexeykazakov: Yes. I agree. The code would be cleaner. Will add it.

sbose78:
Agreed, was thinking out aloud.

controller/users.go
alexeykazakov: Done - https://github.com/fabric8-services/fabric8-auth/pull/393/commits/a95d5821603eb9901efe189e6b586d69d3a90654

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-5`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-393-6`

PR #395, opened 2018-03-20
Set resource limits on Container level

aslakknutsen:
Restrict how much memory the container can consume.

Limit to avoid it overtaking the cluster if the process were to go rogue.

Related to openshiftio/openshift.io#2685


fabric8cd:
@aslakknutsen snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-395-1`

PR #399, opened 2018-03-22
Add API to assign a role

### LINKED ISSUE #398: API to associate ( 'assign' ) a user with a role, for a specific resource
Question: Which role does a user need to have to be able to assign roles?

sbose78:
fixes #398 

```
PATCH /api/resources/:resourceID/roles/:roleName
Authorization: Bearer <AccessToken>
{
  "data": [
    { "type": "identities", "id": "<identityID1>" },
    { "type": "identities", "id": "<identityID2>" },
    ...
  ]
}
```

- [x] support for assigning multiple identityIDs
- [x] determine if the user who is assigning the role has sufficient privileges to do so.
- [x] controller
- [x] controller tests
- [x] service 
- [x] service tests
- [x] model
- [x] model tests


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-1`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-2`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-3`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-6`

sbryzak:
The PR itself looks good, however I've been thinking about whether this endpoint is sufficient for the workflow.  In particular, the "edit role" screen in invision:

[https://redhat.invisionapp.com/share/MFGCA0W2ZQA#/screens/289544106](https://redhat.invisionapp.com/share/MFGCA0W2ZQA#/screens/289544106)

has the user selecting which roles the user should have for the resource.  I am sure we want to avoid forcing the client to make multiple calls here, and so the endpoint should support assigning/revoking multiple roles at once.  I would achieve this by having a SetRoles() method (instead of Assign()), and allow an array of role names to be passed in.  This would be the inclusive set of roles to assign for the resource, if any other roles not in the array have previously been assigned then they should be revoked, and all roles in the array should be granted.

sbose78:
Very good point @sbryzak . 

I think we should support the following APIs
 
```
PATCH /:resourceID/roles/:roleName  - assign a new role - only pass the roles to be added in the payload
PUT /:resourceID/roles/:roleName - update/edit roles - pass all roles in the payload
DELETE /:resourceID/roles/:roleName - de-assign a role - only pass the role to be removed, in the payload
```

What do you think ?  
I agree that if we finally go with the UX specified above, we would need to use the one where we pass all roles which we want the user to be assigned to.

sbose78:
However, @sbryzak : 

For the current collaborators API: We need to support assigning roles to a user(s) ( instead of only editing roles ) as done in this PR.
The first step to start using the new authz api is to get rid of the current keycloak authz we use.

sbryzak:
@sbose78 The API you suggested makes sense, PUT should be used to entirely replace the resource's *roles*, which is exactly what we want.  As for the collaborators API, wouldn't the PATCH endpoint address this requirement?

sbose78:
> As for the collaborators API, wouldn't the PATCH endpoint address this requirement?
you mean,
```
PATCH /:resourceID/roles/:roleName  - assign a new role 
```
Yes - have to refactor my PR for the same @sbryzak  :)

sbose78:
@sbryzak 

This design https://redhat.invisionapp.com/share/MFGCA0W2ZQA#/screens/289544113
would need this PR / PATCH request.

sbryzak:
@sbose78 It looks like the UI designs have been updated again, let's discuss at our meeting shortly.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-7`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-8`

authorization/role/service/role_management_service.go
```diff
@@ -95,3 +109,29 @@ func (r *RoleManagementServiceImpl) ListByResourceAndRoleName(ctx context.Contex
...
 	return roles, err
 }
+
+// Assign assigns a user/identity to role, for a specific resource
+func (r *RoleManagementServiceImpl) Assign(ctx context.Context, identityIDs []uuid.UUID, resourceID string, roleName string) error {
+
+	// TODO: Check if the user is authorized to assign roles to other users.
+	var err error
+	return application.Transactional(r.db, func(appl application.Application) error {
+		err = appl.ResourceRepository().CheckExists(ctx, resourceID)
```
alexeykazakov: Another unnecessary DB call IMO.
So, there are four DB calls total instead of just one.

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -467,3 +549,9 @@ func (s *roleManagementServiceBlackboxTest) checkScopeBelongsToResourceType(db *
 	require.NotEmpty(s.T(), scopesReturned)
 	require.NoError(s.T(), err)
 }
+
+func validateIdentityRole(s *roleManagementServiceBlackboxTest, expected identityrole.IdentityRole, actual identityrole.IdentityRole) {
```
alexeykazakov: This code is duplicated in a few places. Just move it to test package and reuse.

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -427,6 +437,78 @@ func (s *roleManagementServiceBlackboxTest) TestCanListRolesWithoutScopes() {
...
+			RoleID:         newRole.RoleID,
+			//Resource:       *createdResource,
+			ResourceID: createdIdentityRole.ResourceID,
+			Identity:   *createdIdentity,
+			IdentityID: createdIdentity.ID,
+		}
+		s.identityRoleRepo.Create(s.Ctx, &newIdentityRole)
+	}
+
+	returnedRoles, err = s.roleManagementService.ListAssignmentsByIdentityAndResource(s.Ctx, createdIdentityRole.ResourceID, createdIdentityRole.IdentityID)
```
alexeykazakov: Almost everything in this test function is duplicated in repo (it's also true for other tests). Let's try to move the code which creates test data to some test package then reuse it in the tests.

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -427,6 +437,78 @@ func (s *roleManagementServiceBlackboxTest) TestCanListRolesWithoutScopes() {
...
+	testR, err := testsupport.CreateTestResource(s.Ctx, s.DB, *testRT, uuid.NewV4().String(), nil)
+	require.NoError(t, err)
+	require.NotNil(t, testR)
+
+	identityList := []uuid.UUID{testIdentity.ID}
+
+	err = s.roleManagementService.Assign(s.Ctx, identityList, testR.ResourceID, testRole.Name)
+	require.NoError(t, err)
+
+	identityRoles, err := s.roleManagementService.ListByResourceAndRoleName(s.Ctx, testR.ResourceID, testRole.Name)
```
alexeykazakov: Same here. Duplication.

controller/resource_roles.go
```diff
@@ -2,29 +2,42 @@ package controller
...
 	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/fabric8-services/fabric8-auth/login"
 	"github.com/goadesign/goa"
+	uuid "github.com/satori/go.uuid"
+)
+
+const (
+	// ROLE_ASSIGNMENT_SCOPE is the scope the user needs to have as part of the roles assigned to her
+	// that will enabled her to assign roles to other users.
+	ROLE_ASSIGNMENT_SCOPE = "assign_role" // The exact name will be refactored in a different PR/commit once we decide on the name of the role.
```
alexeykazakov: Should be "assign_scope" I think. It's a scope. Not a role ;)

controller/resource_roles.go
```diff
@@ -69,6 +82,62 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+	}
+
+	// check if the current user token belongs to a user who has the necessary privileges
+	// for assigning roles to other users.
+	hasScope, err := c.permissionService.HasScope(ctx, *currentUser, ctx.ResourceID, ROLE_ASSIGNMENT_SCOPE)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+	if !hasScope {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewForbiddenError("user is not authorized to assign roles"))
```
alexeykazakov: I'm ok with using both 401 or 403 here. But we must be contestant with our error statuses when rejecting privileged operations because of missing permissions.

@sbryzak can you reminder what we are supposed to return here? 401 or 403?

controller/resource_roles.go
```diff
@@ -69,6 +82,62 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+	}
+
+	// In a batch assignment of roles, all users need to be a part of the resource
+	// Only then futher role assignments would be allowed.
+	for _, identity := range ctx.Payload.Data {
```
alexeykazakov: "Users need to be a part of resource" - this is confusing :)
What you are ding here is just checking if the user has any explicit role for the resource. I see the intention here for this endpoint. It's about managing existing roles rather then inviting to them. But what about inherited roles or roles mapped from the parent resources?
It looks like we need more sophisticated ListAssignmentsByIdentityAndResource method which will take care of it.

@sbryzak thoughts?

controller/resource_roles.go
```diff
@@ -69,6 +82,62 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+		if err != nil {
+			return jsonapi.JSONErrorResponse(ctx, err)
+		}
+		if len(assignedRoles) == 0 {
+			log.Error(ctx, map[string]interface{}{
+				"resource_id": ctx.ResourceID,
+				"identity_id": identityIDAsUUID,
+				"role":        ctx.RoleName,
+			}, "identity not part of a  resource cannot be assigned a role")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewForbiddenError(fmt.Sprintf("identity %s does not belong to the resource", identity.ID)))
```
alexeykazakov: Same question regarding 401 vs 403.

controller/resource_roles.go
```diff
@@ -69,6 +82,62 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
 
+// AssignRole assigns a specific role for a resource, to one or more identities.
+func (c *ResourceRolesController) AssignRole(ctx *app.AssignRoleResourceRolesContext) error {
+
+	currentUser, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+	if currentUser == nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("identity ID not found in token"))
```
alexeykazakov: Missing test for this error. See https://codecov.io/gh/fabric8-services/fabric8-auth/pull/399/diff

controller/resource_roles.go
alexeykazakov: Missing tests for this error (which is a big deal to test that user without needed permissions can't assign roles!). See https://codecov.io/gh/fabric8-services/fabric8-auth/pull/399/diff

authorization/role/model/role_management_model_service.go
```diff
@@ -38,6 +40,32 @@ type GormRoleManagementModelService struct {
...
+
+// Assign assigns a user identity with a role, for a specific resource
+func (r *GormRoleManagementModelService) Assign(ctx context.Context, identityID uuid.UUID, resourceID string, roleName string) error {
+
+	rt, err := r.repository.ResourceRepository().Load(ctx, resourceID)
+	if err != nil {
+		return err
+	}
+
+	roleRef, err := r.repository.RoleRepository().Lookup(ctx, roleName, rt.ResourceType.Name)
```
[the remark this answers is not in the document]
sbryzak: I believe this is because the role name is being passed in, and a lookup must be done on the role table to find the actual role ID for that resource type (which needs to be looked up also).  We could conceivably reduce the number of DB calls by executing a raw query which joins the `resource` and `role` tables by `resource_type_id` and obtaining the `role_id` that way.  This could go inside a helper method somewhere, possibly in the role repository type.

controller/resource_roles.go
sbryzak: My understanding is that this scope is being used to allow a user to assign a role to another user for a resource, so `assign_role` would be correct here.

controller/resource_roles.go
sbryzak: @alexeykazakov I did some quick refresher research on this, there is actually a problem with the 401 response in that the standard seems to incorrectly label is as 'Unauthorized' when in actual fact the error code is used to return an authentication error.  I.e. the user hasn't authenticated, or their authentication is incorrect somehow.  In this instead, I think @sbose78 's use of 403 is correct, as the user is definitely authenticated (otherwise they wouldn't have gotten this far) however they lack the scope necessary to perform the operation.  This is an easy mistake to make due to the confusion surrounding the http response codes, and I wouldn't be surprised if I've made it myself elsewhere in the code.

controller/resource_roles.go
sbryzak: @alexeykazakov very good question about how to handle inherited roles, I think we need to spend some time thinking about what impact this may have on the UI and invitation workflow.  As for the comment, I agree it's confusing. :)  I would rewrite it as something like this:

````
// In batch assignment of roles all selected users must have previously been assigned 
// privileges for the resource, otherwise the invitation workflow should be used instead
````

File: controller/resource_roles.go (hunk 8 above)
Comment (sbryzak) [reply controller/resource_roles.go]: The UI shouldn't actually allow this to happen, I would return a Bad Request error here.  I would also rewrite the string as:

````
"cannot update roles for an identity without an existing role"
````

File: controller/resource_roles.go [hunk 11]
```diff
@@ -69,6 +82,62 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+			log.Error(ctx, map[string]interface{}{
+				"resource_id": ctx.ResourceID,
+				"identity_id": identityIDAsUUID,
+				"role":        ctx.RoleName,
+			}, "identity not part of a  resource cannot be assigned a role")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewForbiddenError(fmt.Sprintf("identity %s does not belong to the resource", identity.ID)))
+		}
+	}
+
+	// Now that we have confirmed that all the users were part of the resource ( ex. space )
```
sbryzak: Same rewording here, I suggest "Now that we have confirmed that all users have pre-existing roles"

controller/resource_roles.go
alexeykazakov: I'm fine with 403. We just have to make sure we always use 403 if there is lack of permissions.

controller/resource_roles.go
alexeykazakov: Ah, right! Didn't think about in this way ;)

controller/resource_roles.go
sbose78: `BadRequest` sounds nice to me.

authorization/role/service/role_management_service.go
```diff
@@ -95,3 +109,29 @@ func (r *RoleManagementServiceImpl) ListByResourceAndRoleName(ctx context.Contex
 
 	return roles, err
 }
+
+// Assign assigns a user/identity to role, for a specific resource
+func (r *RoleManagementServiceImpl) Assign(ctx context.Context, identityIDs []uuid.UUID, resourceID string, roleName string) error {
+
+	// TODO: Check if the user is authorized to assign roles to other users.
```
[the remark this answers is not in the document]
sbose78: yeah, I could move out the `check for scope` as a service method, and then consume it in the controller.

controller/resource_roles.go
alexeykazakov: Maybe some codecov glitch.

authorization/role/model/role_management_model_service.go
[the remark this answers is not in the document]
sbryzak: I just noticed there's also still a .Debug() statement at line 185 in this file.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-9`

authorization/role/identityrole/repository/identity_role_blackbox_test.go
```diff
@@ -89,6 +90,47 @@ func (s *identityRoleBlackBoxTest) TestExistsRole() {
 
 }
 
+func (s *identityRoleBlackBoxTest) TestListByResourceAndIdentity() {
+	t := s.T()
+	createdIdentityRole := createAndLoadIdentityRole(s)
```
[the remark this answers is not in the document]
sbose78: 4a1a8e2 
added a lot of noisy data :)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-10`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-11`

authorization/role/model/role_management_model_service_backbox_test.go
```diff
@@ -202,3 +212,79 @@ func (s *roleManagementModelServiceBlackboxTest) TestGetIdentityRoleByResourceAn
...
+
+	testRole, err := testsupport.CreateTestRole(s.Ctx, s.DB, *testRT, uuid.NewV4().String())
+	require.NoError(t, err)
+	require.NotNil(t, testRole)
+
+	testR, err := testsupport.CreateTestResource(s.Ctx, s.DB, *testRT, uuid.NewV4().String(), nil)
+	require.NoError(t, err)
+	require.NotNil(t, testR)
+
+	err = s.repo.Assign(s.Ctx, testIdentity.ID, testR.ResourceID, testRole.Name)
```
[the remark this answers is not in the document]
sbose78: https://github.com/fabric8-services/fabric8-auth/pull/399/commits/ea9a4ac67ee2441e4c075023d716bc55d8e22c14

authorization/role/model/role_management_model_service_backbox_test.go
```diff
@@ -202,3 +212,79 @@ func (s *roleManagementModelServiceBlackboxTest) TestGetIdentityRoleByResourceAn
...
+	require.NoError(t, err)
+
+	identityRoles, err := s.repo.ListByResourceAndRoleName(s.Ctx, testR.ResourceID, testRole.Name)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(identityRoles))
+	require.Equal(t, testR.ResourceID, identityRoles[0].ResourceID)
+	require.Equal(t, testRole.RoleID, identityRoles[0].RoleID)
```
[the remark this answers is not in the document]
sbose78: 181a261a7332a67f174ff55cb55a2a22cd5a2fa9

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-12`

authorization/role/service/role_management_service.go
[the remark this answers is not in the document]
sbose78: >  why does this service class contain code that accesses the repositories directly?

@sbryzak , because things like checking if resource exists and then reporting the right error felt more like an operation closer to the business logic. But I see your point.

authorization/role/model/role_management_model_service.go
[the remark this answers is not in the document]
sbose78: Correct, I was thinking of something radical. Should we consider getting rid of the role_id and use the ( name , resource_type ) as the primary key ? Because almost everywhere we are going to encourage use of the role name, instead of the role ID.

authorization/role/service/role_management_service.go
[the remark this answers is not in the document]
sbose78: Now that the 'service' layer is gone, the business logic in the controller would include:

- check if the user X who is adding user Y to the space has a role with the `assign_role` scope. If not return, 403
- check if the user already belongs to the space. if not, return 403

authorization/role/service/role_management_service.go
```diff
@@ -51,6 +54,17 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
 	return roles, err
 }
 
+// ListAssignmentsByIdentityAndResource lists all the roles that have been assigned to a specific user/identity for a specific resource
+func (r *RoleManagementServiceImpl) ListAssignmentsByIdentityAndResource(ctx context.Context, resourceID string, identityID uuid.UUID) ([]identityrole.IdentityRole, error) {
```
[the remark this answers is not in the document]
sbose78: correct, updated comment.

authorization/role/service/role_management_service.go
```diff
@@ -95,3 +109,29 @@ func (r *RoleManagementServiceImpl) ListByResourceAndRoleName(ctx context.Contex
 
 	return roles, err
 }
+
+// Assign assigns a user/identity to role, for a specific resource
```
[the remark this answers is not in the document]
sbose78: correct, updated comment.

authorization/role/service/role_management_service.go
sbose78: Could you check again, please ? 
- after the refactoring.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-13`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-14`

sbose78:
refactored  tests https://github.com/fabric8-services/fabric8-auth/pull/399/commits/1f5d4803ef926c7f883008785f38a57da0228830

controller/resource_roles.go
sbose78: df382a6

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-16`

controller/resource_roles.go
sbose78: 2c8e786

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-17`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-399-19`

sbose78:
Will be opening a new PR - since a lot has changed and fixing conflicts wouldn't be productive :)

PR #414, opened 2018-03-28
Fix PK for role_scope table

alexeykazakov:
We missed `gorm:"primary_key"` tag for role_scope table, so, GORM didn't set PKs for that table. It resulted to "DELETE FROM role_table" in our db clean func.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-414-1`

PR #415, opened 2018-03-28
when new user is being pulled from keycloak, set as RELEASED

sbose78:
fixes prod-preview specific issue https://github.com/openshiftio/openshift.io/issues/2692

### DIFF
--- a/login/service.go
+++ b/login/service.go
@@ -624,6 +624,9 @@ func (keycloak *KeycloakOAuthProvider) CreateOrUpdateIdentityInDB(ctx context.Co
 		if identity.User.Cluster == "" {
 			identity.User.Cluster = configuration.GetOpenShiftClientApiUrl()
 		}
+		if identity.User.FeatureLevel == "" {
+			identity.User.FeatureLevel = account.DefaultFeatureLevel
+		}
 		if err != nil {
 			log.Error(ctx, map[string]interface{}{
 				"keycloak_identity_id": keycloakIdentityID,
--- a/migration/sql-files/025-fix-feature-level.sql
+++ b/migration/sql-files/025-fix-feature-level.sql
@@ -1,1 +1,1 @@
-UPDATE users SET feature_level = 'released' WHERE feature_level = '' and email like '%+preview%';
+UPDATE users SET feature_level = 'released' WHERE (feature_level = '' or feature_level is null)  and email like '%+preview%';
--- a/migration/sql-files/025-fix-feature-level.sql
+++ b/migration/sql-files/025-fix-feature-level.sql
@@ -1,0 +1,1 @@
+UPDATE users SET feature_level = 'released' WHERE feature_level = '' and email like '%+preview%';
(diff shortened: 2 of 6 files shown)


alien-ike:
### Ike Plugins (test-keeper)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-415-1`

migration/sql-files/025-fix-feature-level.sql
```diff
@@ -0,0 +1 @@
+UPDATE users SET feature_level = 'released' WHERE feature_level = '' and email like '%+preview%';
```
xcoulon: what about adding a default value to the `feature_level` column as well ?

migration/sql-test-files/025-before-fix-feature-level.sql
```diff
@@ -0,0 +1 @@
+insert into users (id, cluster, email, email_private,feature_level) values ('a2ff2e0a-724c-4318-a8ef-38441f5205f6', 'test', 'migration-test-1025+preview@mail.com', false,'');
```
xcoulon: what about a `nil` value in the `feature_level` column, too ? just to make sure we're not missing a case...

login/service.go
```diff
@@ -624,6 +624,9 @@ func (keycloak *KeycloakOAuthProvider) CreateOrUpdateIdentityInDB(ctx context.Co
 		if identity.User.Cluster == "" {
 			identity.User.Cluster = configuration.GetOpenShiftClientApiUrl()
 		}
+		if identity.User.FeatureLevel == "" {
```
xcoulon: just curious: what's the benefit of setting the default value here instead of using a `default` constraint in the column of the database table ?

migration/sql-files/025-fix-feature-level.sql
sbose78: It already has a default value, right ?

migration/sql-test-files/025-before-fix-feature-level.sql
sbose78: good point!

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-415-2`

migration/sql-test-files/025-before-fix-feature-level.sql
sbose78: the column is set to have NOT NULL

login/service.go
sbose78: because we already have it in the database, and it still doesn't work :)

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-415-3`

login/service.go
xcoulon: ah of course! the value is `<empty>` but not `NULL`, so the `default` is not applied :/

migration/sql-test-files/025-before-fix-feature-level.sql
xcoulon: 😂

Issue #418, opened 2018-04-03
Support offline token generation


alexeykazakov:
Related to https://github.com/fabric8-services/fabric8-auth/issues/229

sbose78:
Hi, what needs to change here? 
```
func (mgm *tokenManager) GenerateUnsignedUserRefreshToken(ctx context.Context, keycloakRefreshToken string, identity *account.Identity) (*jwt.Token, error) {
```
Wouldn't this work ?

PR #427, opened 2018-04-07
Support offline token generation

alexeykazakov:
For refresh tokens generated from a KC token:
- Set `"typ":"Offline"` instead of `"Refresh"`

For refresh tokens generated from an Identity:
- Set `"typ":"Offline"` instead of `"Refresh"`
- Set `"exp":0` (it's already set to o for tokens generated from a KC token)

To be clear. Current token generation (from a KC token) is not broken. It seems that token type is ignored by KC anyway and it's only expiration time which matters. But we need to fix expiration time for tokens generated from an Identity for future use. This is the main reason for this PR. We change the token type just for better visibility.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/418


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-427-3`

PR #432, opened 2018-04-11
Build on RHEL

jmelis:
Enable build on RHEL if TARGET=rhel

### DIFF
--- a/Dockerfile.deploy.rhel
+++ b/Dockerfile.deploy.rhel
@@ -1,7 +1,7 @@
 FROM registry.devshift.net/osio-prod/base/golang:latest
 
 LABEL maintainer "Devtools <devtools@redhat.com>"
-LABEL author "Konrad Kleine <kkleine@redhat.com>"
+LABEL author "Devtools <devtools@redhat.com>"
 ENV LANG=en_US.utf8
 ENV AUTH_INSTALL_PREFIX=/usr/local/auth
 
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @jmelis for this contribution!



sbose78:
hi @jmelis , what needs to happen for the "WIP" tag to go? :)

Dockerfile.deploy.rhel
```diff
@@ -0,0 +1,24 @@
+FROM registry.devshift.net/osio-prod/base/golang:latest
+
+LABEL maintainer "Devtools <devtools@redhat.com>"
+LABEL author "Konrad Kleine <kkleine@redhat.com>"
```
alexeykazakov: Well.. the author should be Jaime Melis I think ;)

kwk: It is a copy of my Dockerfile ;) essentially

alexeykazakov: It's not just a 100% copy, right? :) This docker file is based on your original file but with modifications. So, there are many authors of this file if to be precise...

cico_setup.sh
```diff
@@ -102,26 +102,35 @@ function run_tests_with_coverage() {
 }
 
 function tag_push() {
-  TARGET=$1
-  docker tag fabric8-auth-deploy $TARGET
-  docker push $TARGET
+  local tag
```
kwk: why not `local tag=$1` ?

jmelis: it should have been like that, sorry I didn't see it before

kwk: @jmelis no worries. I'm not even using `local` yet. Will implement it in the future in our shell scripts. ;)

PR #435, opened 2018-04-11
Move 'FetchKeys()' function into a subpkg

alexeykazakov:
It's a copy of https://github.com/fabric8-services/fabric8-auth/pull/434 plus jwt->jwk package renaming.

Needed to get rid of all auth dependencies in fetchKeys function so it can be consumed by other services like WIT without need to pull other Auth dependencies.


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-435-1`

PR #439, opened 2018-04-12
Protect search API

### LINKED ISSUE #438: protect the search API
1. Make auth API changes.
2. See if UI is already sending the token in the search request ( 'Add collaborator' workflow )

sbose78:
fixes #438


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-439-1`

sbose78:
Checked the UI -

1. Login
2. Open Space
3. Settings -> Collaborators
4. Add collaborator by typing in name from drop-down.

Check the network console in Chrome. 
The request has the Authorization header.

controller/search.go
```diff
@@ -33,14 +34,18 @@ func NewSearchController(service *goa.Service, db application.DB, configuration
 // Users runs the user search action.
 func (c *SearchController) Users(ctx *app.UsersSearchContext) error {
 
+	_, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
```
alexeykazakov: We should add jwt security to design instead.

alexeykazakov: Or you actually added it. Good. But then you don't have to check the token here.

sbose78: Yeah, but this makes sure that the IDE-integration token doesn't work here as well.

alexeykazakov: And worth adding tests for deprovisioned users too.

alexeykazakov: We could check the approved claim though.. But rejecting deprovisioned users is better IMO.

sbose78: Done d3e8660a2bbbb6c9349ce4d3ac7c78c4ff2d13a4
Yeah `approved` claim could have been checked, and might have been less expensive.
but, better be a bit expensive than leaky.

controller/search.go
```diff
@@ -33,14 +34,18 @@ func NewSearchController(service *goa.Service, db application.DB, configuration
 // Users runs the user search action.
 func (c *SearchController) Users(ctx *app.UsersSearchContext) error {
 
+	_, err := login.LoadContextIdentityIfNotDeprovisioned(ctx, c.db)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
```
alexeykazakov: Don't have to wrap it in errors.NewUnauthorizedError(). LoadContextIdentityIfNotDeprovisioned() will return NewUnauthorizedError.

sbose78: But the controller should be explicitly making it a 401 irrespective of what is returned from there ?

alexeykazakov: Sure. But "return jsonapi.JSONErrorResponse(ctx, err)" is enough. err is already NewUnauthorizedError.

alexeykazakov: Anyway. It's a minor thing. Merging.

PR #440, opened 2018-04-12
Don't return external tokens for deprovisioned users

alexeykazakov:


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-440-1`

sbose78:
Makes it expensive, but worth it.

alexeykazakov:
A bit more expensive, yeah. But not much.

PR #442, opened 2018-04-13
Lock down writes if deprovisioned

sbose78:
Lock down 
- Add/Delete collaborators
- Add/Delete space
- Patch User

Patch user might take a performance hit because of the numerous requests it gets from the UI to store `contextInformation`


alien-ike:
### Ike Plugins (test-keeper)

PR #443, opened 2018-04-13
Add tests for permission service

sbose78:
The tests are the same as permission model service, but nevertheless, they should be there to ensure that the service level contract is satisfied.


alexeykazakov:
It looks like there some issues with the test data is not properly cleaned up again.

sbose78:
The service layer has been removed.

PR #447, opened 2018-04-16
WIP Refactor service layer

### LINKED ISSUE #446: Refactor/simplify service layer
This issue is to track the discussion and refactoring of the model service layer.

My personal suggestions are here as a starting point:

1) have just one service implementation, e.g. InvitationModelService, remove the default InvitationService implementation however keep the interface definition (which InvitationModelService will implement)
2) Create the service instances on demand, since they are lightweight objects and contain minimal state - add methods like InvitationService() to the Application interface
3) Have GormBase return the "model" implementation of the service
4) Pass extra "stuff" that the service needs to the service constructor, i.e. instead of passing the Repositories as a parameter in every single method, pass it in the constructor and keep the function prototypes clean
5) Move transaction control to the controller layer, where it already is for other operations

sbryzak:
This PR is a WIP proof of concept for the ongoing discussions on the service layer refactoring.  Using the role management service, what this PR does:

- [x] Remove the `/authorization/role/model` directory
- [x] Merges the model service into the role service type
- [x] Combines all service blackbox tests into one file, removes duplicates
- [x] Remove service creation from `/main.go`
- [x] Adds `RoleManagementService()` method to the Application type, implements the method in `gormapplication.go`
- [x] Moves transaction control to `resource_roles.go` controller

This refactoring has had the effect of greatly reducing duplication, reduction of redundant code and significant simplification of the service layer.  In my opinion we should apply the same refactoring to other service types to achieve the same benefits.

Fixes #446


sbryzak:
I'm going to abandon this PR and implement the service layer refactoring in stages, as discussed with @alexeykazakov

PR #449, opened 2018-04-17
Migrate to `dep` for dependency management 

alexeykazakov:
Based on https://github.com/fabric8-services/fabric8-wit/pull/2011


sbose78:
tested, looks good.

PR #450, opened 2018-04-18
Change container path to use RHEL registry

jmelis:

PR #451, opened 2018-04-18
Fix golang image path for RHEL Docker file

alexeykazakov:
This should have been part of https://github.com/fabric8-services/fabric8-auth/pull/450 ;)


jmelis:
> This should have been part of #450 ;)

Totally!

PR #473, opened 2018-04-26
return external tokens in descending order of created_at timestamp

sbose78:
If a user has multiple github or OSO tokens in the token storage, then return the tokens in the descending order of `created_at` because the most relevant token would always be the most recent one.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-473-1`

PR #474, opened 2018-04-27
[WIP] refactor(uuid): parity patch for UUID API upgrade in f8-wit

debloper:
- WIT repo needs to upgrade goa version
- This also means upgrading go.uuid version upgrade
- go.uuid new version has changed its API to multi-value return
- This patch prepares for the dep upgrades and syncs with the new UUID API

Unblocks fabric8-services/fabric8-wit#2059

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -298,8 +298,7 @@
 [[projects]]
   name = "github.com/satori/go.uuid"
   packages = ["."]
-  revision = "f58768cc1a7a7e77a3bd49e98cdd21419399b6a3"
-  version = "v1.2.0"
+  revision = "master"
 
 [[projects]]
   name = "github.com/sergi/go-diff"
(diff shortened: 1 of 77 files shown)


sbose78:
Where is the package version change ?

Gopkg.lock
```diff
@@ -298,8 +298,7 @@
 [[projects]]
   name = "github.com/satori/go.uuid"
   packages = ["."]
-  revision = "f58768cc1a7a7e77a3bd49e98cdd21419399b6a3"
-  version = "v1.2.0"
+  revision = "master"
```
baijum: Using `master` is bit risky. I would suggest to use the latest commit hash here.

debloper:
> Where is the package version change ?

@sbose78 In follow up commit(s). Will have to hold on to that approach, as @kwk has come up with a much better and overall a lot cleaner approach (fabric8-services/fabric8-wit#2063) which will then unblock my PRs to both the repos.

sbose78:
Not relevant for the time being.

PR #475, opened 2018-04-27
Set maximum allowed username to 45 chars

sbose78:
When registration-app calls Auth service to `Create` a new user, return 400 Bad request if username is longer than 45 characters.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -129,6 +129,8 @@ const (
 	// sentry
 	varEnvironment = "environment"
 	varSentryDSN   = "sentry.dsn"
+
+	varMaxUsernameLength = "max.username.length"
 )
 
 type serviceAccountConfig struct {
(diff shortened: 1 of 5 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-475-1`

sbose78:
Possible fix for https://github.com/openshiftio/openshift.io/issues/2318

alexeykazakov:
If we return Bad Request for such users then the users will be stuck in the wait list with zero information about the reason. We will need to coordinate with PM and others first. Putting it on hold for now.

sbose78:
1. PM can send out a communication to those users who are stuck in the waitlist to register with a different ( shorter username ). 

2. And we could ask OSO folks to remove those users with long usernames from the waitlist.

What do you think?

alexeykazakov:
Closing because RHD already enforces 45 character limitation during registration - https://github.com/openshiftio/openshift.io/issues/2318#issuecomment-385066688

PR #481, opened 2018-05-02
Add the "Assign role" API

### LINKED ISSUE #398: API to associate ( 'assign' ) a user with a role, for a specific resource

sbose78:
replaces #399 
```
PUT /api/resources/:resourceID/roles
Authorization: Bearer <AccessToken>
{
  "data": [
    { "role": "admin", "ids": ["<identityID1>"] },
    { "role": "contributor", "ids": ["<identityID2>", "<identityID3>"] },
    ...
  ]
}
```
fixes #398

### DIFF
--- a/authorization/role/repository/identity_role_blackbox_test.go
+++ b/authorization/role/repository/identity_role_blackbox_test.go
@@ -14,6 +14,7 @@ import (
 	errs "github.com/pkg/errors"
 
 	"context"
+	"fmt"
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -62,6 +63,14 @@ func (s *identityRoleBlackBoxTest) TestOKToDelete() {
 		// that none of the role objects returned include the one deleted.
 		require.NotEqual(s.T(), identityRole.IdentityRoleID.String(), data.IdentityRoleID.String())
 	}
+}
+
+func (s *identityRoleBlackBoxTest) TestDeleteUnknownFails() {
+	identityRoleID := uuid.NewV4()
+
+	err := s.repo.Delete(s.Ctx, identityRoleID)
+	require.Error(s.T(), err)
+	assert.Equal(s.T(), fmt.Sprintf("identity_role with id '%s' not found", identityRoleID), err.Error())
 }
 
 func (s *identityRoleBlackBoxTest) TestOKToLoad() {
--- a/authorization/role/repository/identity_role_blackbox_test.go
+++ b/authorization/role/repository/identity_role_blackbox_test.go
@@ -183,20 +183,20 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
 
 func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByIdentityAndResource() {
 	g := s.DBTestSuite.NewTestGraph()
-	newSpace := g.CreateSpace(g.ID("myspace"))
+	newSpace := g.CreateSpace()
 
 	var createdIdentities []uuid.UUID
 
 	for i := 0; i <= 10; i++ {
-		user := g.CreateUser(g.ID("foo"))
+		user := g.CreateUser()
 		newSpace.AddAdmin(user)
 		createdIdentities = append(createdIdentities, user.Identity().ID)
 	}
 
 	// noise
 	for i := 0; i <= 10; i++ {
-		g.CreateSpace(g.ID("myspace")).AddAdmin(g.CreateUser(g.ID("somename")))
-		newSpace.AddContributor(g.CreateUser(g.ID("someothername")))
+		g.CreateSpace().AddAdmin(g.CreateUser())
+		newSpace.AddContributor(g.CreateUser())
 	}
 
 	for _, i := range createdIdentities {
@@ -210,7 +210,7 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByIdentityAndResource() 
 
 func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownIdentity() {
 	g := s.DBTestSuite.NewTestGraph()
-	newSpace := g.CreateSpace(g.ID("myspace"))
+	newSpace := g.CreateSpace()
 
 	knownRoleID := getKnownRoleIDForSpace(s)
 
@@ -226,8 +226,8 @@ func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownIdentity() {
 
 func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownRole() {
 	g := s.DBTestSuite.NewTestGraph()
-	newSpace := g.CreateSpace(g.ID("myspace"))
-	existingUser := g.CreateUser(g.ID("somename"))
+	newSpace := g.CreateSpace()
+	existingUser := g.CreateUser()
 
 	ir := role.IdentityRole{
 		IdentityRoleID: uuid.NewV4(),
@@ -242,7 +242,7 @@ func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownRole() {
 func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownResource() {
 	g := s.DBTestSuite.NewTestGraph()
 
-	existingUser := g.CreateUser(g.ID("somename"))
+	existingUser := g.CreateUser()
 	knownRoleID := getKnownRoleIDForSpace(s)
 
 	ir := role.IdentityRole{
@@ -258,9 +258,9 @@ func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownResource() {
 func (s *identityRoleBlackBoxTest) TestCreateIdentityExistingAssignmentFails() {
 	g := s.DBTestSuite.NewTestGraph()
 
-	existingUser := g.CreateUser(g.ID("somename"))
+	existingUser := g.CreateUser()
 	knownRoleID := getKnownRoleIDForSpace(s)
-	newSpace := g.CreateSpace(g.ID("myspace"))
+	newSpace := g.CreateSpace()
 
 	ir := role.IdentityRole{
 		IdentityRoleID: uuid.NewV4(),
--- a/authorization/role/repository/identity_role_blackbox_test.go
+++ b/authorization/role/repository/identity_role_blackbox_test.go
@@ -1,6 +1,7 @@
 package repository_test
 
 import (
+	"github.com/fabric8-services/fabric8-auth/authorization"
 	"testing"
 
 	account "github.com/fabric8-services/fabric8-auth/account/repository"
@@ -10,7 +11,9 @@ import (
 	"github.com/fabric8-services/fabric8-auth/errors"
 	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
 	testsupport "github.com/fabric8-services/fabric8-auth/test"
+	errs "github.com/pkg/errors"
 
+	"context"
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -176,6 +179,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
 
 	require.Len(s.T(), identityRoles, 1)
 	require.Equal(s.T(), identityRole.IdentityRoleID, identityRoles[0].IdentityRoleID)
+}
+
+func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByIdentityAndResource() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+
+	var createdIdentities []uuid.UUID
+
+	for i := 0; i <= 10; i++ {
+		user := g.CreateUser(g.ID("foo"))
+		newSpace.AddAdmin(user)
+		createdIdentities = append(createdIdentities, user.Identity().ID)
+	}
+
+	// noise
+	for i := 0; i <= 10; i++ {
+		g.CreateSpace(g.ID("myspace")).AddAdmin(g.CreateUser(g.ID("somename")))
+		newSpace.AddContributor(g.CreateUser(g.ID("someothername")))
+	}
+
+	for _, i := range createdIdentities {
+		result, err := s.repo.FindIdentityRolesByIdentityAndResource(context.Background(), newSpace.SpaceID(), i)
+		require.NoError(s.T(), err)
+		require.Len(s.T(), result, 1)
+		require.Equal(s.T(), i, result[0].IdentityID)
+		require.Equal(s.T(), newSpace.SpaceID(), result[0].ResourceID)
+	}
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownIdentity() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+
+	knownRoleID := getKnownRoleIDForSpace(s)
+
+	ir := role.IdentityRole{
+		IdentityRoleID: uuid.NewV4(),
+		IdentityID:     uuid.NewV4(), // unknown identity
+		ResourceID:     newSpace.SpaceID(),
+		RoleID:         knownRoleID,
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownRole() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+	existingUser := g.CreateUser(g.ID("somename"))
+
+	ir := role.IdentityRole{
+		IdentityRoleID: uuid.NewV4(),
+		IdentityID:     existingUser.Identity().ID,
+		ResourceID:     newSpace.SpaceID(),
+		RoleID:         uuid.NewV4(), // unknown role
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownResource() {
+	g := s.DBTestSuite.NewTestGraph()
+
+	existingUser := g.CreateUser(g.ID("somename"))
+	knownRoleID := getKnownRoleIDForSpace(s)
+
+	ir := role.IdentityRole{
+		IdentityRoleID: uuid.NewV4(),
+		IdentityID:     existingUser.Identity().ID,
+		ResourceID:     uuid.NewV4().String(),
+		RoleID:         knownRoleID,
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityExistingAssignmentFails() {
+	g := s.DBTestSuite.NewTestGraph()
+
+	existingUser := g.CreateUser(g.ID("somename"))
+	knownRoleID := getKnownRoleIDForSpace(s)
+	newSpace := g.CreateSpace(g.ID("myspace"))
+
+	ir := role.IdentityRole{
+		IdentityRoleID: uuid.NewV4(),
+		IdentityID:     existingUser.Identity().ID,
+		ResourceID:     newSpace.SpaceID(),
+		RoleID:         knownRoleID,
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.NoError(s.T(), err)
+
+	ir.IdentityRoleID = uuid.NewV4()
+	err = s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.DataConflictError{}, errs.Cause(err))
+}
+
+func getKnownRoleIDForSpace(s *identityRoleBlackBoxTest) uuid.UUID {
+	roles, err := s.roleRepo.FindRolesByResourceType(context.Background(), authorization.ResourceTypeSpace)
+	require.Nil(s.T(), err)
+	require.Len(s.T(), roles, 3)
+
+	knownRoleID, err := uuid.FromString(roles[0].RoleID)
+	require.Nil(s.T(), err)
+	return knownRoleID
 }
 
 func createAndLoadIdentityRole(s *identityRoleBlackBoxTest) *role.IdentityRole {
--- a/authorization/role/service/role_management_service.go
+++ b/authorization/role/service/role_management_service.go
@@ -66,20 +66,9 @@ func (r *RoleManagementServiceImpl) Assign(ctx context.Context, assignedBy uuid.
 	// check if the current user token belongs to a user who has the necessary privileges
 	// for assigning roles to other users.
 	permissionService := permservice.NewPermissionService(r.repo)
-	hasScope, err := permissionService.HasScope(ctx, assignedBy, resourceID, authorization.ScopeForManagingRolesInResourceType(rt.Name))
+	err = permissionService.RequireScope(ctx, assignedBy, resourceID, authorization.ScopeForManagingRolesInResourceType(rt.Name))
 	if err != nil {
-		log.Error(ctx, map[string]interface{}{
-			"resource_id": resourceID,
-			"identity_id": assignedBy,
-		}, "error determining if user may manage roles for resource")
 		return err
-	}
-	if !hasScope {
-		log.Error(ctx, map[string]interface{}{
-			"resource_id": resourceID,
-			"identity_id": assignedBy,
-		}, "user not authorized to assign roles")
-		return errors.NewForbiddenError("user is not authorized to assign roles")
 	}
 
 	// Valid all the roles and user identity IDs, and ensure each user has been previously assigned
--- a/authorization/role/service/role_management_service.go
+++ b/authorization/role/service/role_management_service.go
@@ -19,7 +19,7 @@ type RoleManagementService interface {
 	ListByResource(ctx context.Context, resourceID string) ([]rolerepo.IdentityRole, error)
 	ListAvailableRolesByResourceType(ctx context.Context, resourceType string) ([]role.RoleDescriptor, error)
 	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
-	Assign(ctx context.Context, assignedBy uuid.UUID, identitiesToBeAssigned []uuid.UUID, resourceID string, roleName string) error
+	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string) error
 }
 
 // NewRoleManagementService creates a new service to manage role assignments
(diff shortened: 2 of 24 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/03c169d26a0abffbff315713c55ebf2798673334/test-keeper.yml).

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-1`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-2`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-3`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-4`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-5`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-6`

controller/resource_roles.go
```diff
@@ -87,6 +91,74 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
 
+// AssignRole assigns a specific role for a resource, to one or more identities.
+func (c *ResourceRolesController) AssignRole(ctx *app.AssignRoleResourceRolesContext) error {
+
+	currentUser, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+	if currentUser == nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("identity ID not found in token"))
```
sbose78: Added a test https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-4bd1c185cd704de02aaa93a028975a89R316 for this but doesn't show up in coverage. Must be missing something very obvious.

authorization/role/service/role_management_service.go
```diff
@@ -38,3 +40,25 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
 func (r *RoleManagementServiceImpl) ListAvailableRolesByResourceType(ctx context.Context, resourceType string) ([]role.RoleDescriptor, error) {
 	return r.repo.RoleRepository().FindRolesByResourceType(ctx, resourceType)
 }
+
+// Assign assigns an identity ( users or organizations or teams or groups ) with a role, for a specific resource
+func (r *RoleManagementServiceImpl) Assign(ctx context.Context, identityID uuid.UUID, resourceID string, roleName string) error {
+	rt, err := r.repo.ResourceRepository().Load(ctx, resourceID)
```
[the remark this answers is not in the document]
sbose78: b0dc261

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-10`

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -196,3 +200,52 @@ func (s *roleManagementServiceBlackboxTest) TestGetIdentityRoleByResourceAndRole
...
+	err := s.repo.Assign(context.Background(), uuid.NewV4(), uuid.NewV4().String(), uuid.NewV4().String())
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleWithRoleNotFound() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+	err := s.repo.Assign(context.Background(), uuid.NewV4(), newSpace.SpaceID(), uuid.NewV4().String())
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
```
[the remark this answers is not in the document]
sbose78: Repository level tests for 
- unknown identity
- unknown role
- unknown resource
- identity role assignment already exists
https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-1b9d789ec603b76b1701376ca3fb689dR211

Service level tests for the above:
https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-e6bc49bff4aa33f0cfee81b2275359f8R228

Controller test for existing assignment:
https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-4bd1c185cd704de02aaa93a028975a89R243

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-11`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-12`

authorization/role/service/role_management_service.go
```diff
@@ -38,3 +40,25 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
...
+		return err
+	}
+
+	ir := rolerepo.IdentityRole{
+		ResourceID: resourceID,
+		IdentityID: identityID,
+		RoleID:     roleRef.RoleID,
+	}
+
+	return r.repo.IdentityRoleRepository().Create(ctx, &ir)
```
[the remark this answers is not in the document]
sbose78: https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-e6bc49bff4aa33f0cfee81b2275359f8R228

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -196,3 +200,71 @@ func (s *roleManagementServiceBlackboxTest) TestGetIdentityRoleByResourceAndRole
...
+
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleResourceNotFound() {
+	err := s.repo.Assign(context.Background(), uuid.NewV4(), uuid.NewV4().String(), uuid.NewV4().String())
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleWithRoleNotFound() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+	err := s.repo.Assign(context.Background(), uuid.NewV4(), newSpace.SpaceID(), uuid.NewV4().String())
```
alexeykazakov: This will fail because the identity is not found. Not because the role is not found.
Use some existing identity in this test.

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -196,3 +200,71 @@ func (s *roleManagementServiceBlackboxTest) TestGetIdentityRoleByResourceAndRole
...
+	userToBeAssigned := g.CreateUser(g.ID("somename"))
+	err := s.repo.Assign(context.Background(), userToBeAssigned.Identity().ID, newSpace.SpaceID(), authorization.AdminRole)
+	require.Nil(s.T(), err)
+
+	err = s.repo.Assign(context.Background(), userToBeAssigned.Identity().ID, newSpace.SpaceID(), authorization.AdminRole)
+	require.IsType(s.T(), errors.DataConflictError{}, errs.Cause(err))
+}
+
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleResourceNotFound() {
+	err := s.repo.Assign(context.Background(), uuid.NewV4(), uuid.NewV4().String(), uuid.NewV4().String())
```
alexeykazakov: This will fail because the identity is not found. Not because the resource is not found.
Use some existing identity and existing role in this test.

controller/resource_roles.go
alexeykazakov: ContextIdentity() will return an error if there is no "sub" claim in the token.
So, if err=="nil" then currentUser != nil. You can just remove this check.

controller/resource_roles.go
```diff
@@ -87,6 +91,74 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
 	})
 }
 
+// AssignRole assigns a specific role for a resource, to one or more identities.
+func (c *ResourceRolesController) AssignRole(ctx *app.AssignRoleResourceRolesContext) error {
+
+	currentUser, err := login.ContextIdentity(ctx)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
```
alexeykazakov: Log all errors in the controller.

controller/resource_roles.go
```diff
@@ -87,6 +91,74 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+			log.Error(ctx, map[string]interface{}{
+				"resource_id": ctx.ResourceID,
+				"identity_id": identityIDAsUUID,
+				"role":        ctx.RoleName,
+			}, "assignment failed")
+			return jsonapi.JSONErrorResponse(ctx, err)
+		}
+	}
+
+	if err != nil {
```
alexeykazakov: err can't be nil here, right?

controller/resource_roles_blackbox_test.go
```diff
@@ -205,6 +216,148 @@ func (rest *TestResourceRolesRest) TestListAssignedRolesByRoleNameFromInheritedO
...
+}
+
+func (rest *TestResourceRolesRest) TestAssignRoleBadRequestUserNotInSpace() {
+	g := rest.DBTestSuite.NewTestGraph()
+	res := g.CreateSpace(g.ID("somespacename"))
+
+	var identitiesToBeAssigned []*app.UpdateUserID
+	for i := 0; i <= 2; i++ {
+		testUser := g.CreateUser(g.ID("someusername"))
+		identitiesToBeAssigned = append(identitiesToBeAssigned, &app.UpdateUserID{Type: "identities", ID: testUser.Identity().ID.String()})
```
alexeykazakov: Add at least one identity which has a role in the space to the top of that list. So, we can make sure we don't just check the first identity in the list and actually go through the whole list.

controller/resource_roles_blackbox_test.go
```diff
@@ -205,6 +216,148 @@ func (rest *TestResourceRolesRest) TestListAssignedRolesByRoleNameFromInheritedO
...
+	test.AssignRoleResourceRolesBadRequest(rest.T(), svc.Context, svc, ctrl, res.SpaceID(), authorization.SpaceContributorRole, payload)
+}
+
+func (rest *TestResourceRolesRest) TestAssignRoleForbiddenNotAllowedToAssignRoles() {
+	g := rest.DBTestSuite.NewTestGraph()
+	res := g.CreateSpace(g.ID("somespacename"))
+
+	var identitiesToBeAssigned []*app.UpdateUserID
+	for i := 0; i <= 2; i++ {
+		testUser := g.CreateUser(g.ID("someusername"))
```
alexeykazakov: Add res.AddViewer(testUser) to make sure we are trying to assign roles to good users. Just in case :)

controller/resource_roles_blackbox_test.go
```diff
@@ -205,6 +216,148 @@ func (rest *TestResourceRolesRest) TestListAssignedRolesByRoleNameFromInheritedO
...
+	res := g.CreateSpace(g.ID("somespacename"))
+
+	var identitiesToBeAssigned []*app.UpdateUserID
+	for i := 0; i <= 2; i++ {
+		identitiesToBeAssigned = append(identitiesToBeAssigned, &app.UpdateUserID{Type: "identities", ID: uuid.NewV4().String() + "#$%"})
+	}
+
+	// Create a user who has the privileges to assign roles
+	adminUser := g.CreateUser("adminuser")
+	res.AddAdmin(adminUser) //not really an admin
```
alexeykazakov: It should be an admin. You are testing invalid identity here. Not the lack of permissions ;)

design/role.go
```diff
@@ -70,6 +71,21 @@ var _ = a.Resource("resource_roles", func() {
 		a.Response(d.NotFound, JSONAPIErrors)
 		a.Response(d.Unauthorized, JSONAPIErrors)
 	})
+	a.Action("assignRole", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.PATCH("/:resourceID/roles/:roleName"),
+		)
+		a.Payload(updateUserIDList) // should refactor this variable's name in collaborators design definition too.
+		a.Description("Assigns an identity to a role, for a specific resource")
```
alexeykazakov: Assigns a role to an identity :)

gormapplication/application.go
```diff
@@ -137,7 +137,7 @@ func (g *GormDB) PermissionService() permissionservice.PermissionService {
 }
 
 func (g *GormDB) RoleManagementService() roleservice.RoleManagementService {
-	return roleservice.NewRoleManagementService(g)
+	return roleservice.NewRoleManagementService(g, g)
```
alexeykazakov: I think we just need to create a new interface which will extends both Repositories and Transaction interfaces and just pass GormDB once. It's for a separate PR though. Let's not do it in this PR.

authorization/role/service/role_management_service_blackbox_test.go
sbose78: Not really, the check for role's existence comes first https://github.com/fabric8-services/fabric8-auth/pull/481/files/92b41bdb74998622b4504406c8b41e3f920e5500#diff-98f55d2fa7b5423f02129bc536e43b09R58 .

authorization/role/service/role_management_service_blackbox_test.go
sbose78: Not really, the check for resource's existence is done first https://github.com/fabric8-services/fabric8-auth/pull/481/files/92b41bdb74998622b4504406c8b41e3f920e5500#diff-98f55d2fa7b5423f02129bc536e43b09R52

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-13`

authorization/role/service/role_management_service_blackbox_test.go
alexeykazakov: Ah ok. But anyway let’s test it explicitly so the test is good even if we change the implementation.

gormapplication/application.go
sbose78: Agreed, while doing this - it looked a bit odd to me :)

authorization/role/service/role_management_service_blackbox_test.go
sbose78: Added it - it does make sense from an API contract perspective to have *all but one* valid inputs 56ce11be50f5a3f4a519dbfa0343a562c392c878

authorization/role/service/role_management_service_blackbox_test.go
sbose78: Added it - it does make sense from an API contract perspective to have *all but one* valid inputs 56ce11be50f5a3f4a519dbfa0343a562c392c878

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-14`

controller/resource_roles.go
sbose78: 73ae1a3
3a1ec42940e7a90357b68ba4c59c59cdfb0931d7

controller/resource_roles_blackbox_test.go
sbose78: 9997dd1ef7c1ff44f274a54d3ec237720f37cacd

controller/resource_roles_blackbox_test.go
sbose78: 9997dd1ef7c1ff44f274a54d3ec237720f37cacd

controller/resource_roles_blackbox_test.go
sbose78: 9997dd1ef7c1ff44f274a54d3ec237720f37cacd :D

design/role.go
sbose78: a6581af88ea9d06a6d865829b2d6e03469942bb1 :D

controller/resource_roles.go
sbose78: 3a1ec42940e7a90357b68ba4c59c59cdfb0931d7 right, it gets checked above this already.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-15`

gormapplication/application.go
sbryzak: I am very likely changing this as part of #490

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
 }
 
+func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByIdentityAndResource() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+
+	var createdIdentities []uuid.UUID
+
+	for i := 0; i <= 10; i++ {
+		user := g.CreateUser(g.ID("foo"))
```
sbryzak: You can't give each created user the same ID

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
+		newSpace.AddAdmin(user)
+		createdIdentities = append(createdIdentities, user.Identity().ID)
+	}
+
+	// noise
+	for i := 0; i <= 10; i++ {
+		g.CreateSpace(g.ID("myspace")).AddAdmin(g.CreateUser(g.ID("somename")))
```
sbryzak: Same as above, you can't give each space and user the same ID

authorization/role/repository/identity_role_blackbox_test.go
sbryzak: In fact, I can't see where this identifier is even used - why is it being set?

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
+		require.NoError(s.T(), err)
+		require.Len(s.T(), result, 1)
+		require.Equal(s.T(), i, result[0].IdentityID)
+		require.Equal(s.T(), newSpace.SpaceID(), result[0].ResourceID)
+	}
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownIdentity() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
```
sbryzak: Identifier doesn't seem to be used

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
+		RoleID:         knownRoleID,
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownRole() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace(g.ID("myspace"))
+	existingUser := g.CreateUser(g.ID("somename"))
```
sbryzak: Both these identifiers aren't used.

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
+		RoleID:         uuid.NewV4(), // unknown role
+	}
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityRolesUnknownResource() {
+	g := s.DBTestSuite.NewTestGraph()
+
+	existingUser := g.CreateUser(g.ID("somename"))
```
sbryzak: This identifier not used either.

authorization/role/repository/identity_role_blackbox_test.go
```diff
@@ -178,6 +181,111 @@ func (s *identityRoleBlackBoxTest) TestFindIdentityRolesByResource() {
...
+	err := s.repo.Create(context.Background(), &ir)
+	require.IsType(s.T(), errors.NotFoundError{}, errs.Cause(err))
+}
+
+func (s *identityRoleBlackBoxTest) TestCreateIdentityExistingAssignmentFails() {
+	g := s.DBTestSuite.NewTestGraph()
+
+	existingUser := g.CreateUser(g.ID("somename"))
+	knownRoleID := getKnownRoleIDForSpace(s)
+	newSpace := g.CreateSpace(g.ID("myspace"))
```
sbryzak: Unused identifiers also.

controller/resource_roles.go
```diff
@@ -87,6 +91,92 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+			"resource_id": ctx.ResourceID,
+			"role":        ctx.RoleName,
+		}, "error getting identity information from token")
+		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError(err.Error()))
+	}
+
+	// check if the current user token belongs to a user who has the necessary privileges
+	// for assigning roles to other users.
+
+	hasScope, err := c.app.PermissionService().HasScope(ctx, *currentUser, ctx.ResourceID, authorization.ManageTeamsInSpaceScope)
```
sbryzak: This permission check should be in the service. See [this](https://github.com/fabric8-services/fabric8-auth/blob/master/authorization/team/service/team_service.go#L68) for an example of another service.

controller/resource_roles.go
```diff
@@ -87,6 +91,92 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
...
+				"role":        ctx.RoleName,
+			}, "invalid identity ID")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewBadParameterError("identity", identity.ID).Expected("uuid"))
+		}
+		identitiesToBeAssigned = append(identitiesToBeAssigned, identityIDAsUUID)
+	}
+
+	// In batch assignment of roles all selected users must have previously been assigned
+	// privileges for the resource, otherwise the invitation workflow should be used instead
+	for _, identityIDAsUUID := range identitiesToBeAssigned {
```
sbryzak: This business logic should be in the service, not the controller.

design/role.go
```diff
@@ -70,6 +71,21 @@ var _ = a.Resource("resource_roles", func() {
 		a.Response(d.NotFound, JSONAPIErrors)
 		a.Response(d.Unauthorized, JSONAPIErrors)
 	})
+	a.Action("assignRole", func() {
+		a.Security("jwt")
+		a.Routing(
+			a.PATCH("/:resourceID/roles/:roleName"),
+		)
+		a.Payload(updateUserIDList) // should refactor this variable's name in collaborators design definition too.
```
sbryzak: I'm wondering if the payload structure is ideal.  Will the `"type"` ever be any value other than `"identities"` ?  Also, thinking about the workflow, I can understand how this endpoint is useful for "batch" assigning a single role to a number of users, however how about assigning multiple roles to a single user, as is possible in the current wireframes that we've been reviewing from the UX team?  Will this be done as multiple calls to the auth service? If so, that seems inefficient.

sbose78:
@sbryzak  Moved the business logic to the service - can you review and let me know if this look good? Only after that will I refactor tests ;)

https://github.com/fabric8-services/fabric8-auth/pull/481/commits/a45b8eee58dcfd6e5540a49a3eab5a671ac756a0

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-16`

authorization/role/repository/identity_role_blackbox_test.go
sbose78: Well the method was looking for it, hence passed it. I suppose we are gonna wire up the passed argument somewhere later on?

sbryzak: You don't need to set an ID if you don't intend to look up the object later on.  See [this](https://github.com/fabric8-services/fabric8-auth/blob/master/authorization/team/service/team_service_blackbox_test.go#L80) for an example.

authorization/role/service/role_management_service.go
```diff
@@ -38,3 +47,96 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
...
+				"role":        roleName,
+			}, "identity not part of a resource cannot be assigned a role")
+			return errors.NewBadParameterErrorFromString("identityID", identityIDAsUUID, fmt.Sprintf("cannot update roles for an identity %s without an existing role", identityIDAsUUID))
+		}
+	}
+
+	// Now that we have confirmed that all users have pre-existing role assignments
+	// we can proceed with the assignment of roles.
+	for _, identityIDAsUUID := range identitiesToBeAssigned {
+		err = r.assign(ctx, identityIDAsUUID, resourceID, roleName)
```
sbryzak: You wish to wrap each role assignment in its own transaction?  What if for some reason one of the transactions fails?  Do we wish to support a "partial" success?

controller/resource_roles.go
```diff
@@ -87,6 +89,38 @@ func (c *ResourceRolesController) ListAssignedByRoleName(ctx *app.ListAssignedBy
 	})
 }
 
+// AssignRole assigns a specific role for a resource, to one or more identities.
+func (c *ResourceRolesController) AssignRole(ctx *app.AssignRoleResourceRolesContext) error {
```
sbryzak: Controller method looks much better now, it's simply dealing with unmarshalling the request parameters from the payload and passing them to the service layer. :)

authorization/role/service/role_management_service.go
```diff
@@ -38,3 +47,96 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
...
+}
+
+// Assign assigns an identity ( users or organizations or teams or groups ) with a role, for a specific resource
+// IMPORTANT: This is a transactional method, which manages its own transaction/s internally
+func (r *RoleManagementServiceImpl) Assign(ctx context.Context, assignedBy uuid.UUID, identitiesToBeAssigned []uuid.UUID, resourceID string, roleName string) error {
+	// check if the current user token belongs to a user who has the necessary privileges
+	// for assigning roles to other users.
+
+	permissionService := permservice.NewPermissionService(r.repo)
+	hasScope, err := permissionService.HasScope(ctx, assignedBy, resourceID, authorization.ManageTeamsInSpaceScope)
```
sbryzak: Can we add a new scope here, the name of this one doesn't fit the description of the operation.

sbose78: Ah yeah, this has to be `ManageUsersInSpaceScope`

authorization/role/service/role_management_service.go
sbose78: Initially I did think of having it in one transaction ( that's how the tests are now ) - then yesterday thought of partial success. But now I realize that its a bad idea. Among many other things, it'll be hard to report errors to users while also letting them know of the partial success.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-18`

authorization/role/service/role_management_service.go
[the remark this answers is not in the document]
sbryzak: Since this is (essentially) just 2 lines of code now, can we just roll it into the Assign() method?

authorization/role/service/role_management_service.go
```diff
@@ -38,3 +47,94 @@ func (r *RoleManagementServiceImpl) ListByResource(ctx context.Context, resource
...
+	roleRef, err := r.repo.RoleRepository().Lookup(ctx, roleName, rt.ResourceType.Name)
+	if err != nil {
+		return err
+	}
+
+	// check if the current user token belongs to a user who has the necessary privileges
+	// for assigning roles to other users.
+
+	permissionService := permservice.NewPermissionService(r.repo)
+	hasScope, err := permissionService.HasScope(ctx, assignedBy, resourceID, authorization.ManageRoleAssignmentsInSpaceScope)
```
sbryzak: I'm still a little uncomfortable about the scope name here.  Won't this method be used to assign roles for other types of resources, not just spaces?  E.g. assigning an admin role to a user for an organization.  I think what might be useful here is a helper method, perhaps in the `authorization` package, something like:

`func ManageRoleScopeForResourceType(resourceType string) string`

which accepts the name of a resource type as a parameter, and returns the name of the scope required to be able to manage roles for that resource type.

design/role.go
sbryzak: To add to my previous comment, this is the wireframe for what I'm referring to:

[https://redhat.invisionapp.com/share/QNGBDSOHP6K#/screens/294869601](https://redhat.invisionapp.com/share/QNGBDSOHP6K#/screens/294869601)

In the UI shown here, it is possible to select multiple roles to assign to a single user.  I am certain that we don't wish this operation to require multiple calls to the auth service.

sbose78: > I'm wondering if the payload structure is ideal. Will the "type" ever be any value other than "identities"

For now no. But we can remove the enum, and/or add other constants to the enum as well. I was re-using the existing payload at the moment.

> I am certain that we don't wish this operation to require multiple calls to the auth service.

I don't have a good answer for this honestly - and might go for multiple calls to the auth service for now since we wouldn't be having too many roles to start with. So for assigning 3 roles to 50 users, there's gonna be 3 requests which IMO isn't gonna harm the user experience, if they are all async. 

The alternative would be to include the rolename in the payload itself. Wouldn't that make the transaction too big? Or should we have multiple transactions for each payload item?

authorization/role/service/role_management_service.go
sbose78: Good point, how does this look https://github.com/fabric8-services/fabric8-auth/pull/481/commits/9d01ef67707f27d15566dfad39bc8b55eb009bf8 ?

sbose78: ( hold on )

sbose78: How does this look :) https://github.com/fabric8-services/fabric8-auth/pull/481/files#diff-7dbcc3cc497b6deb6660e41e2eb25becR65

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-20`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-21`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-22`

sbryzak:
I've restructured the payload a bit, to support the assignment of multiple roles.  This provides a significant simplification for the UI team when it comes to assigning multiple roles to multiple users as the result of a single user action (for example, how would they deal with a partial success?)

```
PATCH /api/resources/:resourceID/roles/:roleName
Authorization: Bearer <AccessToken>
{
  "data": [
    { "role": "owner", "ids": ["<identityID1>"] },
    { "role": "contributor", "ids": ["<identityID2>", "<identityID3>"] },
    ...
  ]
}
```

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-23`

authorization/role/service/role_management_service.go
```diff
@@ -50,84 +50,92 @@ func (r *RoleManagementServiceImpl) ListAvailableRolesByResourceType(ctx context
...
+					"identity_id": assignedBy,
+				}, "error looking up existing assignments")
+				return err
+			}
+			if len(assignedRoles) == 0 {
+				log.Error(ctx, map[string]interface{}{
+					"resource_id": resourceID,
+					"identity_id": identityIDAsUUID,
+				}, "identity not previously assigned a resource role cannot be assigned another role")
+				return errors.NewBadParameterErrorFromString("identityID", identityIDAsUUID, fmt.Sprintf("cannot update roles for an identity %s without an existing role", identityIDAsUUID))
```
alexeykazakov: Just realized that we can't actually use this service for initial role assignment (during resource creation for example) when the user doesn't have any role in the resource yet.

We need to add a new method to the service for that or add a param to Assign()

authorization/role/service/role_management_service.go
```diff
@@ -50,84 +50,92 @@ func (r *RoleManagementServiceImpl) ListAvailableRolesByResourceType(ctx context
...
-				return err
+		for roleID, ids := range assignments {
+			for _, identityIDAsUUID := range ids {
+				ir := rolerepo.IdentityRole{
+					ResourceID: resourceID,
+					IdentityID: identityIDAsUUID,
+					RoleID:     roleID,
+				}
+
+				err = r.repo.IdentityRoleRepository().Create(ctx, &ir)
```
alexeykazakov: Another thing I missed in the previous reviews. We just append new roles instead of replacing the existing ones.
So, if the user has "Admin" role and we want to replace it by "Viewer" we can't do it via Assign() (and we don't have anything in the service to remove the existing role). If we assign "Viewer" then the user will have both Admin and Viewer roles.

authorization/role/service/role_management_service.go
alexeykazakov: I will add a new method for that in https://github.com/fabric8-services/fabric8-auth/pull/498

authorization/role/service/role_management_service.go
alexeykazakov: Fixed in https://github.com/fabric8-services/fabric8-auth/pull/481/commits/6cb4606c05f00806f5c36f714ff7e3ec371e2e3a

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-24`

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-481-25`

PR #483, opened 2018-05-03
Reorganize account package

alexeykazakov:
- [x] Move all repositories from /account to /account/repository
- [x] Move all services from /account, /account/email/, /account/userprofile to /account/services


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-483-1`

sbose78:
Scrolled down and saw only package name/location changes. :+1:

Gopkg.lock
```diff
@@ -49,17 +49,17 @@
...
   name = "github.com/fabric8-services/fabric8-tenant"
   packages = ["design"]
-  revision = "master"
+  revision = "2614bf1824754901b7e2f9d7b4ad4d7e3bac61ff"
 
 [[projects]]
   name = "github.com/fabric8-services/fabric8-wit"
   packages = ["design"]
-  revision = "master"
+  revision = "b59d0dc8ae9b2bfea1fb5dc6ff3552c42b735756"
```
sbose78: Goodness, we were using master all along? :) 
I think this happened during the migration from `glide` to `dep` because I'm pretty sure we were using a commit has prior to this.

alexeykazakov: Yes. But it wasn't critical. We control these repos and don't use anything besides generated clients.
But worth fixing anyway :)

PR #487, opened 2018-05-07
Upon dev token generation, create new user in WIT as well

sbose78:
Previously when we generated test tokens, we used to create the identities in WIT as well - that code went away when we started supporting token generation inside Auth service, causing the wit+auth dev workflow to be broken. This PR puts that code back.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-487-1`

PR #489, opened 2018-05-08
WIP - Set ID of dev identity explicitly

sbose78:
To ensure that the identity/user created in WIT has a matching+valid ID.

DO NOT MERGE : trying out a few things with @surajssd to make authorization work with WIT.


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-489-1`

Issue #366, opened 2018-03-08
Create authz resource for new spaces


sbose78:
### Task

When a new space is being created using the existing `Spaces API`, call the resource creation authz APIs inside https://github.com/fabric8-services/fabric8-auth/blob/master/controller/space.go to create a new resource.

### Dependencies
The least that needs to be done on the Authz service:

- [x] Seed the `openshift.io/resource/space` resource_type
     Seed the `contributor`,`viewer`,`admin` role for the above resource_type
     Seed some scope for the `contributor` role (?) https://github.com/fabric8-services/fabric8-auth/issues/383

- [x] Add API to create a new Resource

alexeykazakov:
> Seed the openshift.io/resource/space resource_type

As far as I remember we agreed on using openshift.io/resource/area resource type for spaces. Space == root area.
@sbryzak ?

sbryzak:
@alexeykazakov That is correct, a space is an area.

sbose78:
That sounds wrong to me, because `space` is not equal to root `area`. Whenever a space is created, a root iteration and a root area is created. 

Every `space` has a root `area`. Just like every `space` also has a root `iteration`.
A `codebase` belongs to a space, but does not belong to any area. So, there are things which can be associated with a `space` and not belong to the root `area` or any `area` - example `codebases`.

I would expect us to have resources like 
- openshift.io/resource/space
- openshift.io/resource/area
- openshift.io/resource/iteration

sbose78:
All our discussions have been around 'adding users to space'. Do you think we should have separate resource types for areas/iterations ? The roles/scopes applied to a space can be used for areas/iterations since they are always associated with a space - just the way we do for workitems.

As an example, we can have scopes like `manage` which would employer the roles to create/update iterations & areas.

sbryzak:
@sbose78 it was my understanding that there would be control over the privileges assigned within individual areas within a space, however I don't recall the specific use cases for this.  I guess it depends how an organization structures their spaces/areas.

sbose78:
Exactly, thanks @alexeykazakov

sbose78:
Done.

sbose78:
Dependencies are done. Actual task remains.

PR #495, opened 2018-05-15
Reorganize resource management

alexeykazakov:
This is a prerequisite for https://github.com/fabric8-services/fabric8-auth/issues/366

- [x] Create resource service (and move business logic from controller to service)
- [x] Remove name and owner ID from resource REST API (it doesn't make much sense to store names and owners/creators of the resources created by resource services)
- [x] Remove Update Resource Endpoint. The only thing which would make sense to update in the resource via REST API is the resource parent. But I think we should create some dedicated and explicit endpoint for changing the parent to avoid confusions with empty or missing parent ID in the update payload. Basically in the current implementation an empty Payload (or a payload with missing parent ID) will remove the parent from the resource. Which is not obvious and may cause issues. So, until we really need to support this use case in clients let's just remove this update endpoint.
- [x] More tests.

### DIFF
--- a/authorization/resource/service/resource_service.go
+++ b/authorization/resource/service/resource_service.go
@@ -16,7 +16,6 @@ type ResourceService interface {
 	Delete(ctx context.Context, resourceID string) error
 	Read(ctx context.Context, resourceID string) (*app.Resource, error)
 	Register(ctx context.Context, resourceType string, resourceID, parentResourceID *string) (*resource.Resource, error)
-	Update(ctx context.Context, resourceID, parentResourceID string) (*resource.Resource, error)
 }
 
 // resourceServiceImpl is the implementation of the interface for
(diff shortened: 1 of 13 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-495-1`

authorization/resource/service/resource_service.go
```diff
@@ -0,0 +1,117 @@
...
+// Read reads resource
+func (s *resourceServiceImpl) Read(ctx context.Context, resourceID string) (*app.Resource, error) {
+
+	res, err := s.repo.ResourceRepository().Load(ctx, resourceID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Load the resource type scopes
+	scopes, err := s.repo.ResourceTypeScopeRepository().LookupForType(ctx, res.ResourceTypeID)
```
sbryzak: Do we want to return all the scopes every time?

controller/resource.go
```diff
@@ -61,42 +51,16 @@ func (c *ResourceController) Read(ctx *app.ReadResourceContext) error {
...
-			return error
-		}
-
-		// Load the resource type scopes
-		scopes, error = tr.ResourceTypeScopeRepository().LookupForType(ctx, res.ResourceTypeID)
-
-		return error
-	})
-
+	svc := resourceservice.NewResourceService(c.app, c.app)
```
sbryzak: Add `ResourceService()` method to [Services](https://github.com/fabric8-services/fabric8-auth/blob/master/application/service/services.go) interface, then you can just do:

`svc := c.app.ResourceService()`

(Will also need to add implementation [here](https://github.com/fabric8-services/fabric8-auth/blob/master/gormapplication/application.go))

sbose78:
I agree with the changes, @alexeykazakov .

controller/resource.go
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/495/commits/45158c56db15e82d8b7f3e29a52175f9a1e34025

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-495-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-495-3`

PR #498, opened 2018-05-22
Create authz resource for new spaces

alexeykazakov:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/366

- [x] When a new space is created or deleted then create/delete the corresponding AuthZ resource for the space
- [x] When creating a new resource assign an admin role for the resource creator (requires https://github.com/fabric8-services/fabric8-auth/pull/481 to be merged first)
- [x] Delete relevant references (role mapping, identities roles, identities, etc) when deleting a resource
- [x] Tests

### DIFF
--- a/authorization/resource/service/resource_service_blackbox_test.go
+++ b/authorization/resource/service/resource_service_blackbox_test.go
@@ -142,8 +142,8 @@ func (s *resourceServiceBlackBoxTest) TestDeleteResourceOK() {
 	s.checkIdentityRole(2, teamToDeleteA.ResourceID())
 	s.checkIdentityRole(2, teamToDeleteB.ResourceID())
 
-	s.checkRoleMapping(true, spaceRoleMappingToDelete.RoleMapping().RoleMappingID.String())
-	s.checkRoleMapping(true, grandchildRoleMappingToDelete.RoleMapping().RoleMappingID.String())
+	s.checkRoleMapping(true, spaceRoleMappingToDelete.RoleMapping().RoleMappingID)
+	s.checkRoleMapping(true, grandchildRoleMappingToDelete.RoleMapping().RoleMappingID)
 
 	// Delete the space
 
@@ -166,8 +166,8 @@ func (s *resourceServiceBlackBoxTest) TestDeleteResourceOK() {
 	s.checkIdentityRole(0, teamToDeleteA.ResourceID())
 	s.checkIdentityRole(0, teamToDeleteB.ResourceID())
 
-	s.checkRoleMapping(false, spaceRoleMappingToDelete.RoleMapping().RoleMappingID.String())
-	s.checkRoleMapping(false, grandchildRoleMappingToDelete.RoleMapping().RoleMappingID.String())
+	s.checkRoleMapping(false, spaceRoleMappingToDelete.RoleMapping().RoleMappingID)
+	s.checkRoleMapping(false, grandchildRoleMappingToDelete.RoleMapping().RoleMappingID)
 
 	// Check all not-related artifacts are still present
 
@@ -180,7 +180,7 @@ func (s *resourceServiceBlackBoxTest) TestDeleteResourceOK() {
 	s.checkIdentityRole(1, spaceToStay.SpaceID())
 	s.checkIdentityRole(1, teamToStay.ResourceID())
 
-	s.checkRoleMapping(true, spaceRoleMappingToStay.RoleMapping().RoleMappingID.String())
+	s.checkRoleMapping(true, spaceRoleMappingToStay.RoleMapping().RoleMappingID)
 }
 
 func (s *resourceServiceBlackBoxTest) TestDeleteResourceWithCycleReferencesFails() {
@@ -196,7 +196,7 @@ func (s *resourceServiceBlackBoxTest) TestDeleteResourceWithCycleReferencesFails
 	testsupport.AssertError(s.T(), err, errors.InternalError{}, "cycle resource references detected for resource %s with parent %s", parent.ResourceID(), child.ResourceID())
 }
 
-func (s *resourceServiceBlackBoxTest) checkRoleMapping(shouldExist bool, roleMappingID string) {
+func (s *resourceServiceBlackBoxTest) checkRoleMapping(shouldExist bool, roleMappingID uuid.UUID) {
 	err := s.Application.RoleMappingRepository().CheckExists(s.Ctx, roleMappingID)
 	if shouldExist {
 		assert.NoError(s.T(), err)
--- a/authorization/resource/service/resource_service_blackbox_test.go
+++ b/authorization/resource/service/resource_service_blackbox_test.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/fabric8-services/fabric8-auth/authorization"
-	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
-
 	"github.com/fabric8-services/fabric8-auth/application/service"
+	"github.com/fabric8-services/fabric8-auth/authorization"
+	"github.com/fabric8-services/fabric8-auth/errors"
+	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
+	testsupport "github.com/fabric8-services/fabric8-auth/test"
+
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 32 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-498-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-498-3`

controller/space.go
```diff
@@ -65,14 +67,32 @@ func (c *SpaceController) Create(ctx *app.CreateSpaceContext) error {
 	}
 
 	err = transaction.Transactional(c.app, func(tr transaction.TransactionalResources) error {
-		// Create space resource which will represent the keyclok resource associated with this space
+		// Create space resource which will represent the keycloak resource associated with this space
```
sbryzak: I think this should probably be moved into a standalone service:

- There is a significant amount of business logic here now (coordination of multiple service calls)
- We should be wrapping this operation in a single transaction, this won't work using transaction.Transactional but will work when using the service framework's internal transaction management.

authorization/role/service/role_management_service.go
```diff
@@ -151,3 +152,27 @@ func (s *roleManagementServiceImpl) Assign(ctx context.Context, assignedBy uuid.
 
 	return err
 }
+
+// AssignAsAdmin assigns an identity (users, organizations, teams or groups) with a role for a specific resource.
+// This method doesn't check any permissions and assumes that the caller does all needed permissions checks.
+// As an example: this method is to be used when creating a resource (space) to assign initial admin role to the resource creator.
+// IMPORTANT: This is a transactional method, which manages its own transaction/s internally
+func (s *roleManagementServiceImpl) AssignAsAdmin(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error {
```
sbryzak: I find this method name a little confusing, as there is no actual "admin" role as described, however there are resource admin roles.  Could we perhaps call it something like `ForceAssignNamedRole()` instead?

authorization/role/repository/role_mapping_blackbox_test.go
```diff
@@ -58,6 +58,48 @@ func (s *roleMappingBlackBoxTest) TestOKToDelete() {
...
+	g := s.NewTestGraph()
+
+	// Create 5 role mappings
+	r := g.CreateResource(g.ID("r"))
+	for i := 0; i < 5; i++ {
+		g.CreateRoleMapping(r)
+	}
+
+	// Check all of them are present
+	mappings, err := s.repo.FindForResource(s.Ctx, g.ResourceByID("r").Resource().ResourceID)
```
sbryzak: The "r" resource has already been assigned to a variable `r`, why not use that instead?

`mappings, err := s.repo.FindForResource(s.Ctx, r.Resource().ResourceID)`

authorization/role/repository/role_mapping.go
```diff
@@ -77,26 +78,9 @@ func (m *GormRoleMappingRepository) TableName() string {
 }
 
 // CheckExists returns nil if the given ID exists otherwise returns an error
-func (m *GormRoleMappingRepository) CheckExists(ctx context.Context, ID uuid.UUID) (bool, error) {
+func (m *GormRoleMappingRepository) CheckExists(ctx context.Context, id string) error {
```
sbryzak: Why has this been changed to `string` when the `role_mapping_id` column is a UUID?

controller/space.go
alexeykazakov: We will re-write this controller anyway when switching from Keycloak to AuthZ completely but I think it's a good idea create a space service right now and move all new business logic to that space. But keeping deprecated business logic and the new one in separate controllers/services/transactions is OK IMO...
But anyway let me see what we can do now.

authorization/role/service/role_management_service.go
alexeykazakov: Yeah.. I don't like this name either. Was using it as a temporal name until I come up with a better name but I didn't ;)
I like ForceAssignNamedRole or just ForceAssign.

authorization/role/repository/role_mapping.go
alexeykazakov: Just because base.CheckExists() uses string. We can keep UUID here though.

controller/space.go
```diff
@@ -132,6 +152,24 @@ func (c *SpaceController) Delete(ctx *app.DeleteSpaceContext) error {
...
+	if err != nil {
+		if notFound, _ := errors.IsNotFoundError(err); notFound {
+			log.Warn(ctx, map[string]interface{}{
+				"space_id": ctx.SpaceID,
+			}, "unable to delete authZ space resource: resource not found; that's OK for old spaces")
+		} else {
+			log.Error(ctx, map[string]interface{}{
+				"space_id": ctx.SpaceID,
+			}, "unable to delete authZ space resource")
+			return jsonapi.JSONErrorResponse(ctx, err)
```
sbose78: Usually during soft migration, we don't return an error from the controller if the 'new' code fails. But I think it's fine to do here now since the 'new code' doesn't talk to an external service.

test/graph/team_wrapper.go
```diff
@@ -57,20 +60,21 @@ func newTeamWrapper(g *TestGraph, params []interface{}) teamWrapper {
...
 		Name:             *teamName,
 		ResourceType:     *resourceType,
 		ResourceTypeID:   resourceType.ResourceTypeID,
 		ParentResourceID: &space.ResourceID,
 	}
 
-	err = g.app.ResourceRepository().Create(g.ctx, res)
+	err = g.app.ResourceRepository().Create(g.ctx, w.resource)
 	require.NoError(g.t, err)
+	w.resource.ParentResource = space
```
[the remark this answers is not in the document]
alexeykazakov: No, we can't. This is another GORM weirdness :) For some reason we can set ParentResourceID only when creating a resource. But not ParentResource. Don't remember what error did it cause but the output SQL was not expected. Maybe there is another setting in GORM we don't use and which could solve this problem though..

test/graph/team_wrapper.go
```diff
@@ -2,18 +2,21 @@ package graph
...
 	"github.com/satori/go.uuid"
 	"github.com/stretchr/testify/require"
 )
 
-// teamWrapper represents a space resource domain object
+// teamWrapper represents a team resource domain object
 type teamWrapper struct {
 	baseWrapper
 	identity *account.Identity
+	resource *resource.Resource
```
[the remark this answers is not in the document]
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/498/commits/a82bf3fd432bb4f3f5f7307f60d073c1a1cc3716

test/graph/role_mapping_wrapper.go
```diff
@@ -14,16 +16,20 @@ type roleMappingWrapper struct {
...
 
 	for i := range params {
 		switch t := params[i].(type) {
 		case resourceWrapper:
-			resource = &t
+			resource = t.Resource()
 		case *resourceWrapper:
-			resource = t
+			resource = t.Resource()
+		case spaceWrapper:
```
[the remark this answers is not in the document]
alexeykazakov: It now checks Resource param too. But I decided not to remove resource and space wrappers - https://github.com/fabric8-services/fabric8-auth/pull/498/commits/a82bf3fd432bb4f3f5f7307f60d073c1a1cc3716

authorization/role/service/role_management_service.go
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/498/commits/a82bf3fd432bb4f3f5f7307f60d073c1a1cc3716

authorization/role/repository/role_mapping_blackbox_test.go
alexeykazakov: Right. Done in https://github.com/fabric8-services/fabric8-auth/pull/498/commits/a82bf3fd432bb4f3f5f7307f60d073c1a1cc3716

authorization/role/repository/role_mapping.go
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/498/commits/a82bf3fd432bb4f3f5f7307f60d073c1a1cc3716

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-498-7`

test/graph/team_wrapper.go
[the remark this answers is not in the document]
alexeykazakov: I've created a new space service.

controller/space.go
alexeykazakov: I've created a new service.

sbryzak: I'm still a little concerned that we have two separate transactions in a single operation.  What happens if the second transaction fails?  We don't seem to have any facility for rolling back the first transaction, and if the `space_resources` record is successfully created the first time yet the second transaction fails, subsequent attempts will also fail because of the unique `space_resources_space_id_idx` constraint.  If the continued usage of `space_resources` is only temporary, then this might be acceptable in the meantime - what are the long term plans for this table?

authorization/resource/service/resource_service_blackbox_test.go
```diff
@@ -102,6 +104,131 @@ func (s *resourceServiceBlackBoxTest) TestRegisterReadDeleteResourceWithParentOK
...
+	s.checkIdentity(true, g.UserByID("admin-user").Identity().ID.String())
+	s.checkIdentity(true, teamToStay.TeamID().String())
+
+	s.checkIdentityRole(1, spaceToStay.SpaceID())
+	s.checkIdentityRole(1, teamToStay.ResourceID())
+
+	s.checkRoleMapping(true, spaceRoleMappingToStay.RoleMapping().RoleMappingID)
+}
+
+func (s *resourceServiceBlackBoxTest) TestDeleteResourceWithCycleReferencesFails() {
```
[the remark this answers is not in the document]
alexeykazakov: Yeah... I agree that we need to check cycle references when creating or updating a resource. Let's create an issue for that. But I would still keep the check here too just as an addition safety net. Just because if there is a cycle reference then this method will run in infinite loop.

alexeykazakov:
Not sure why I can't reply to the last comment in the review... Replying here.

> I'm still a little concerned that we have two separate transactions in a single operation. What happens if the second transaction fails? We don't seem to have any facility for rolling back the first transaction, and if the space_resources record is successfully created the first time yet the second transaction fails, subsequent attempts will also fail because of the unique space_resources_space_id_idx constraint. If the continued usage of space_resources is only temporary, then this might be acceptable in the meantime - what are the long term plans for this table?

It's a valid concern. The space controller is a temporal thing. We are going to get rid or the old resource management (so we will remove the first transaction completely) in this train. But during the migration we still need to make sure we handle transaction failure properly. I will add a rollback for the first transaction if the second one fails.

alexeykazakov:
> I will add a rollback for the first transaction if the second one fails.

@sbryzak Done in https://github.com/fabric8-services/fabric8-auth/pull/498/commits/56857c44f88d6523a52cb2e0c0220c0aa0da92a6

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-498-9`

controller/space.go
alexeykazakov: The only information stored in space_resources table is IDs of the remote Keycloak resources by space IDs. Why would we need to keep them when we get rid of KC?

sbose78:
> I'm still a little concerned that we have two separate transactions in a single operation. What happens if the second transaction fails? We don't seem to have any facility for rolling back the first transaction, and if the space_resources record is successfully created the first time yet the second transaction fails, subsequent attempts will also fail because of the unique space_resources_space_id_idx constraint. If the continued usage of space_resources is only temporary, then this might be acceptable in the meantime - what are the long term plans for this table?

Strange, my comment doesn't show up anymore :( 

I was thinking that we shouldn't be removing the `space_resources` table. We ideally should be using a table to store the 'business object' and then create corresponding entities for handling authorization.

Example, the business object might store a lot of extra metadata which could be absolutely irrelevant from an authorization perspective. 

In that case, after migration, the `space_resources` table shouldn't go away, rather it should serve as a table which stores the business object ( with metadata about the space ).

controller/space.go
sbose78: Sorry, my bad! I was thinking about the 'space' table in wit. :)

sbose78:
Ignore my previous comment! I was talking about the `spaces` table in wit.

gormapplication/application.go
```diff
@@ -51,7 +51,7 @@ const (
 
 func NewGormDB(db *gorm.DB) *GormDB {
 	val := new(GormDB)
-	val.db = db
+	val.db = db.Set("gorm:save_associations", false)
```
sbose78: nice :) So, we were inadvertently updating records all this while.

alexeykazakov: Yeah... caught this with Shane when adding tests for this PR. It's not only caused a lot of unnecessary updated on related tables but also caused issues with wrong values in the tables!

PR #500, opened 2018-05-24
Return error if object not found when deleting

alexeykazakov:

### DIFF
--- a/account/repository/user.go
+++ b/account/repository/user.go
@@ -153,14 +153,17 @@ func (m *GormUserRepository) Delete(ctx context.Context, id uuid.UUID) error {
 
 	obj := User{ID: id}
 
-	err := m.db.Delete(&obj).Error
+	db := m.db.Delete(&obj)
 
-	if err != nil {
+	if db.Error != nil {
 		log.Error(ctx, map[string]interface{}{
 			"user_id": id,
-			"err":     err,
+			"err":     db.Error,
 		}, "unable to delete the user")
-		return errs.WithStack(err)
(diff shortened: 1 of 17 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-500-1`

account/repository/user.go
```diff
@@ -153,14 +153,17 @@ func (m *GormUserRepository) Delete(ctx context.Context, id uuid.UUID) error {
 
 	obj := User{ID: id}
 
-	err := m.db.Delete(&obj).Error
+	db := m.db.Delete(&obj)
 
-	if err != nil {
+	if db.Error != nil {
```
sbryzak: Is this thread-safe?

sbryzak: ah nevermind, I didn't notice that you're assigning to a new `db` variable :)

sbryzak: Actually the `db` variable name is a little confusing, could we name it `result` instead, like in the example?

[http://gorm.io/docs/error_handling.html](http://gorm.io/docs/error_handling.html)

alexeykazakov: Yeah.. 'result' would be a better name.

alexeykazakov: Fixed in https://github.com/fabric8-services/fabric8-auth/pull/500/commits/6d5cd7acd5e170920b8f4b35097c72a8fd96bdf5

PR #503, opened 2018-05-29
Automatically create role mappings during resource registration

### LINKED ISSUE #472: Automatically populate role mappings for newly created resources
When a new resource is created, the `DEFAULT_ROLE_MAPPING` table should be queried for any records matching the same resource type of the new resource, and for each one found a corresponding `ROLE_MAPPING` record should be created which maps the roles for the new resource.

Default role mappings should be created for the following roles:

### For space resources, map the admin role to the team admin role

This will allow users that have been assigned the admin role for a space, to also gain admin privileges for any teams within that space.

sbryzak:
This PR fixes #472 

Creates role mapping records for resources during resource registration, based on the default role mappings for the resource type.

### DIFF
--- a/authorization/role/repository/default_role_mapping.go
+++ b/authorization/role/repository/default_role_mapping.go
@@ -2,9 +2,9 @@ package repository
 
 import (
 	"context"
-	"fmt"
 	"time"
 
+	"github.com/fabric8-services/fabric8-auth/application/repository/base"
 	resourcetype "github.com/fabric8-services/fabric8-auth/authorization/resourcetype/repository"
 	"github.com/fabric8-services/fabric8-auth/errors"
 	"github.com/fabric8-services/fabric8-auth/gormsupport"
@@ -62,7 +62,7 @@ func NewDefaultRoleMappingRepository(db *gorm.DB) DefaultRoleMappingRepository {
 
 // DefaultRoleMappingRepository represents the storage interface.
 type DefaultRoleMappingRepository interface {
-	CheckExists(ctx context.Context, ID uuid.UUID) (bool, error)
+	CheckExists(ctx context.Context, ID uuid.UUID) error
 	Load(ctx context.Context, ID uuid.UUID) (*DefaultRoleMapping, error)
 	Create(ctx context.Context, u *DefaultRoleMapping) error
 	Save(ctx context.Context, u *DefaultRoleMapping) error
@@ -78,26 +78,9 @@ func (m *GormDefaultRoleMappingRepository) TableName() string {
 }
 
 // CheckExists returns nil if the given ID exists otherwise returns an error
-func (m *GormDefaultRoleMappingRepository) CheckExists(ctx context.Context, ID uuid.UUID) (bool, error) {
+func (m *GormDefaultRoleMappingRepository) CheckExists(ctx context.Context, id uuid.UUID) error {
 	defer goa.MeasureSince([]string{"goa", "db", "default_role_mapping", "exists"}, time.Now())
-
-	var exists bool
-	query := fmt.Sprintf(`
-		SELECT EXISTS (
-			SELECT 1 FROM %[1]s
-			WHERE
-				default_role_mapping_id=$1
-				AND deleted_at IS NULL
-		)`, m.TableName())
-
-	err := m.db.CommonDB().QueryRow(query, ID.String()).Scan(&exists)
-	if err == nil && !exists {
-		return exists, errors.NewNotFoundError(m.TableName(), ID.String())
-	}
-	if err != nil {
-		return false, errors.NewInternalError(ctx, errs.Wrapf(err, "unable to verify if %s exists", m.TableName()))
-	}
-	return exists, nil
+	return base.CheckExistsWithCustomIDColumn(ctx, m.db, m.TableName(), "default_role_mapping_id", id.String())
 }
 
 // CRUD Functions
@@ -141,24 +124,14 @@ func (m *GormDefaultRoleMappingRepository) Create(ctx context.Context, u *Defaul
 func (m *GormDefaultRoleMappingRepository) Save(ctx context.Context, model *DefaultRoleMapping) error {
 	defer goa.MeasureSince([]string{"goa", "db", "default_role_mapping", "save"}, time.Now())
 
-	obj, err := m.Load(ctx, model.DefaultRoleMappingID)
-	if err != nil {
-		log.Error(ctx, map[string]interface{}{
+	err := m.db.Save(model).Error
+	if err == nil {
+		log.Debug(ctx, map[string]interface{}{
 			"default_role_mapping_id": model.DefaultRoleMappingID,
-			"err": err,
-		}, "unable to update default role mapping")
-		return errs.WithStack(err)
-	}
-	// Select("ResourceTypeID", "FromRoleID", "ToRoleID").Updates(model)
-	err = m.db.Model(obj).Save(model).Error
-	if err != nil {
-		return errs.WithStack(err)
+		}, "default role mapping saved")
 	}
 
-	log.Debug(ctx, map[string]interface{}{
-		"default_role_mapping_id": model.DefaultRoleMappingID,
-	}, "Default role mapping saved!")
-	return nil
+	return errs.WithStack(err)
 }
 
 // List returns all default role mappings
--- a/authorization/role/repository/default_role_mapping.go
+++ b/authorization/role/repository/default_role_mapping.go
@@ -143,7 +143,7 @@ func (m *GormDefaultRoleMappingRepository) Save(ctx context.Context, model *Defa
 		}, "unable to update default role mapping")
 		return errs.WithStack(err)
 	}
-	err = m.db.Debug().Model(obj).Select("ResourceTypeID", "FromRoleID", "ToRoleID").Updates(model).Error
+	err = m.db.Model(obj).Select("ResourceTypeID", "FromRoleID", "ToRoleID").Updates(model).Error
 	if err != nil {
 		return errs.WithStack(err)
 	}
--- a/authorization/role/repository/default_role_mapping.go
+++ b/authorization/role/repository/default_role_mapping.go
@@ -1,0 +1,202 @@
+package repository
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	resourcetype "github.com/fabric8-services/fabric8-auth/authorization/resourcetype/repository"
+	"github.com/fabric8-services/fabric8-auth/errors"
+	"github.com/fabric8-services/fabric8-auth/gormsupport"
+	"github.com/fabric8-services/fabric8-auth/log"
+
+	"github.com/goadesign/goa"
+	"github.com/jinzhu/gorm"
+	errs "github.com/pkg/errors"
+	"github.com/satori/go.uuid"
+)
+
+type DefaultRoleMapping struct {
+	gormsupport.Lifecycle
+
+	// This is the primary key value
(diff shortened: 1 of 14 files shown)


authorization/role/repository/default_role_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+
+	obj, err := m.Load(ctx, model.DefaultRoleMappingID)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"default_role_mapping_id": model.DefaultRoleMappingID,
+			"err": err,
+		}, "unable to update default role mapping")
+		return errs.WithStack(err)
+	}
+	err = m.db.Model(obj).Select("ResourceTypeID", "FromRoleID", "ToRoleID").Updates(model).Error
```
alexeykazakov: We use .Updates() instead of Save() almost everywhere but in Identity and User repos. Main difference is that Update doesn't update empty fields. If the field of "model" is empty then GORM treats it's as not modified and doesn't update it. Which is not critical for default mappings but still..

Save() updates all the fields. So, as result it's impossible to set some field to an empty value (for example an empty string) using Updates().

Updates also require Load() first, so it's two DB calls instead of one in case of Save().
See https://github.com/fabric8-services/fabric8-auth/blob/master/account/repository/identity.go#L232 as an example.

I think we should start using Save() everywhere and use Updates() only if we really need it. WDYT?

authorization/role/repository/default_role_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+
+// Delete removes a single record.
+func (m *GormDefaultRoleMappingRepository) Delete(ctx context.Context, id uuid.UUID) error {
+	defer goa.MeasureSince([]string{"goa", "db", "default_role_mapping", "delete"}, time.Now())
+
+	obj := DefaultRoleMapping{DefaultRoleMappingID: id}
+
+	err := m.db.Delete(&obj).Error
+
+	if err != nil {
```
alexeykazakov: Return NotFoundError if nothing was deleted - https://github.com/fabric8-services/fabric8-auth/blob/master/account/repository/identity.go#L258
And add a test for this.

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+		ResourceTypeID: rt2.ResourceType().ResourceTypeID,
+		FromRoleID:     g.CreateRole(rt).Role().RoleID,
+		ToRoleID:       g.CreateRole().Role().RoleID,
+	}
+
+	err = s.repo.Create(s.Ctx, &rm2)
+	require.NoError(s.T(), err)
+
+	mappings, err := s.repo.List(s.Ctx)
+	require.Nil(s.T(), err, "Could not list role mappings")
```
alexeykazakov: Better to use require.NoError(). In this case the error message will also contain the info about the unexpected error.

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	require.True(s.T(), found1, "first default role mapping not found")
+	require.True(s.T(), found2, "second default role mapping not found")
+
+	err = s.repo.Delete(s.Ctx, rm.DefaultRoleMappingID)
+	assert.Nil(s.T(), err)
+
+	mappings, err = s.repo.List(s.Ctx)
+	require.Nil(s.T(), err, "Could not list role mappings")
```
alexeykazakov: require.NoError()

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+		ResourceTypeID: rt.ResourceType().ResourceTypeID,
+		FromRoleID:     g.CreateRole(rt).Role().RoleID,
+		ToRoleID:       g.CreateRole().Role().RoleID,
+	}
+
+	err := s.repo.Create(s.Ctx, rm)
+	require.NoError(s.T(), err)
+
+	_, err = s.repo.Load(s.Ctx, rm.DefaultRoleMappingID)
+	require.NoError(s.T(), err)
```
alexeykazakov: Check you loaded the expected role mapping.

alexeykazakov: Also check that load() for unknown ID returns NotFoundError.

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+	rm := &rolerepo.DefaultRoleMapping{
+		ResourceTypeID: rt.ResourceType().ResourceTypeID,
+		FromRoleID:     g.CreateRole(rt).Role().RoleID,
+		ToRoleID:       g.CreateRole().Role().RoleID,
+	}
+
+	err := s.repo.Create(s.Ctx, rm)
+	require.NoError(s.T(), err)
+
+	_, err = s.repo.CheckExists(s.Ctx, rm.DefaultRoleMappingID)
```
alexeykazakov: Check both use cases when it returns false and true.

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+	err := s.repo.Create(s.Ctx, rm)
+	require.NoError(s.T(), err)
+
+	rm.ResourceTypeID = g.CreateResourceType().ResourceType().ResourceTypeID
+	rm.ToRoleID = otherRole.Role().RoleID
+	err = s.repo.Save(s.Ctx, rm)
+	require.NoError(s.T(), err)
+
+	updated, err := s.repo.Load(s.Ctx, rm.DefaultRoleMappingID)
+	require.Nil(s.T(), err, "could not load default role mapping")
```
alexeykazakov: require.NoError()

authorization/role/repository/default_role_mapping_blackbox_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.Equal(s.T(), otherRole.Role().RoleID, updated.ToRoleID)
+}
+
+func (s *defaultRoleMappingBlackBoxTest) TestFindForResourceType() {
+	g := s.NewTestGraph()
+	rt := g.CreateResourceType()
+	rm := g.CreateDefaultRoleMapping(rt)
+	g.CreateDefaultRoleMapping()
+
+	mappings, err := s.repo.FindForResourceType(s.Ctx, rt.ResourceType().ResourceTypeID)
```
alexeykazakov: Create some "noise" in the repo to make sure you found the expected mapping only.

authorization/role/repository/role_mapping_blackbox_test.go
```diff
@@ -132,3 +132,16 @@ func (s *roleMappingBlackBoxTest) TestOKToSave() {
 	require.Nil(s.T(), err, "Could not load role mapping")
 	require.Equal(s.T(), rm1.ResourceID, updatedRm.ResourceID)
 }
+
+func (s *roleMappingBlackBoxTest) TestFindForResource() {
+	g := s.NewTestGraph()
```
alexeykazakov: Create some "noise" in the repo to make sure you found the expected mapping only.

authorization/role/repository/default_role_mapping.go
```diff
@@ -0,0 +1,202 @@
...
+		Preload("ToRole").
+		Where("default_role_mapping_id = ?", id).Find(&native).Error
+	if err == gorm.ErrRecordNotFound {
+		return nil, errors.NewNotFoundError("default_role_mapping", id.String())
+	}
+	return &native, errs.WithStack(err)
+}
+
+// Create creates a new record.
+func (m *GormDefaultRoleMappingRepository) Create(ctx context.Context, u *DefaultRoleMapping) error {
```
alexeykazakov: Please add more details to the method comment about the default role mapping object fields. So, it's clear how to create a default role mapping. Some example would be great.
Maybe we should add this description to the Default Role Mapping object itself.

authorization/role/repository/role_mapping.go
```diff
@@ -183,3 +184,15 @@ func (m *GormRoleMappingRepository) Delete(ctx context.Context, id uuid.UUID) er
 
 	return nil
 }
+
```
alexeykazakov: Please also add more details to the Create() method comment of the role mapping repo too. So, it's clear how to create a role mapping. Some example would be great.
Maybe we should add this description to the Role Mapping object itself.

authorization/role/repository/default_role_mapping.go
sbryzak: I've changed this to save() instead.  In fact, I had some issues using Updates() as it was trying to update the related role record also for some reason.

authorization/role/repository/default_role_mapping.go
sbryzak: Done

authorization/role/repository/default_role_mapping_blackbox_test.go
sbryzak: Fixed

authorization/role/repository/default_role_mapping_blackbox_test.go
sbryzak: Fixed

authorization/role/repository/default_role_mapping_blackbox_test.go
sbryzak: Fixed

authorization/role/repository/default_role_mapping_blackbox_test.go
sbryzak: Fixed

authorization/role/repository/default_role_mapping_blackbox_test.go
sbryzak: I was already creating one additional default role mapping record which should have been enough noise to ensure the method was working correctly, I've now upped it to 10.

authorization/role/repository/role_mapping_blackbox_test.go
sbryzak: Done

authorization/role/repository/default_role_mapping.go
sbryzak: I've added quite a large comment to the DefaultRoleMapping struct

authorization/role/repository/role_mapping.go
sbryzak: Done

authorization/role/repository/default_role_mapping.go
alexeykazakov: Looks better but you don't have to load the object before updating it.

authorization/role/repository/default_role_mapping_blackbox_test.go
alexeykazakov: Not really fixed :)

authorization/role/repository/default_role_mapping_blackbox_test.go
alexeykazakov: I've created a dedicated test utility function for such tests recently.
So, you can use something like:
`testsupport.AssertError(s.T(), err, errors.NotFoundError{}, "default_role_mapping with id '%s' not found", id.String())`
It will check if the error in nil or not. Also will check the error type and the message.

authorization/role/repository/default_role_mapping_blackbox_test.go
alexeykazakov: Ah OK. I missed that.

authorization/role/repository/default_role_mapping_blackbox_test.go
alexeykazakov: Still not done.

Issue #511, opened 2018-06-06
Assign/Remove collaborator role when adding/removing space contributors


alexeykazakov:
When we add or remove a space collaborator we should assign contributor role to the collaborator for the corresponding space/resource.

Part of https://github.com/fabric8-services/fabric8-auth/issues/359

PR #512, opened 2018-06-06
Assign/Remove collaborator role when adding/removing space contributors

alexeykazakov:
- [x] Tests

Fixes https://github.com/fabric8-services/fabric8-auth/issues/511

### DIFF
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -52,6 +52,7 @@ type RoleManagementService interface {
 	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
 	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string, appendToExistingRoles bool) error
 	ForceAssign(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error
+	DeleteRoleAssignments(ctx context.Context, byIdentity uuid.UUID, forIdentities []uuid.UUID, resourceID string) error
 }
 
 type TeamService interface {
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -52,7 +52,7 @@ type RoleManagementService interface {
 	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
 	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string, appendToExistingRoles bool) error
 	ForceAssign(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error
-	DeleteRoleAssignments(ctx context.Context, byIdentity uuid.UUID, forIdentities []uuid.UUID, resourceID string) error
+	RevokeResourceRoles(ctx context.Context, byIdentity uuid.UUID, forIdentities []uuid.UUID, resourceID string) error
 }
 
 type TeamService interface {
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -47,9 +47,9 @@ type ResourceService interface {
 }
 
 type RoleManagementService interface {
-	ListByResource(ctx context.Context, resourceID string) ([]rolerepo.IdentityRole, error)
+	ListByResource(ctx context.Context, currentIdentity uuid.UUID, resourceID string) ([]rolerepo.IdentityRole, error)
 	ListAvailableRolesByResourceType(ctx context.Context, resourceType string) ([]role.RoleDescriptor, error)
-	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
+	ListByResourceAndRoleName(ctx context.Context, currentIdentity uuid.UUID, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
 	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string, appendToExistingRoles bool) error
 	ForceAssign(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error
 	RevokeResourceRoles(ctx context.Context, currentIdentity uuid.UUID, identities []uuid.UUID, resourceID string) error
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -52,7 +52,7 @@ type RoleManagementService interface {
 	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
 	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string, appendToExistingRoles bool) error
 	ForceAssign(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error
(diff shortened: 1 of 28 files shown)


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-512-1`

application/service/services.go
```diff
@@ -52,6 +52,7 @@ type RoleManagementService interface {
 	ListByResourceAndRoleName(ctx context.Context, resourceID string, roleName string) ([]rolerepo.IdentityRole, error)
 	Assign(ctx context.Context, assignedBy uuid.UUID, roleAssignments map[string][]uuid.UUID, resourceID string, appendToExistingRoles bool) error
 	ForceAssign(ctx context.Context, assignedTo uuid.UUID, roleName string, res resource.Resource) error
+	DeleteRoleAssignments(ctx context.Context, byIdentity uuid.UUID, forIdentities []uuid.UUID, resourceID string) error
```
sbryzak: Can we call this `RevokeResourceRoles()` instead?

sbryzak: Also, I'm wondering if we could standardize naming a little more, in other places we use `currentUser` to refer to the `identity` object for the currently-authenticated user, could we perhaps use `currentUserID` here instead of `byIdentity`, and then instead of `forIdentities` just call the parameter `identities`?

controller/collaborators.go
```diff
@@ -142,12 +157,59 @@ func (c *CollaboratorsController) AddMany(ctx *app.AddManyCollaboratorsContext)
...
+
+	res := resource.Resource{ResourceType: resourcetype.ResourceType{Name: authorization.ResourceTypeSpace}, ResourceID: spaceID}
+	for _, contributor := range contributors {
+		identityID, err := uuid.FromString(contributor.ID)
+		if err != nil {
+			return autherrors.NewBadParameterError("ids", contributor.ID).Expected("uuid")
+		}
+
+		// Have to use ForceAssign() because Assign() requires assignees to already have any role in the space
+		err = c.app.RoleManagementService().ForceAssign(ctx, identityID, authorization.SpaceContributorRole, res)
```
sbryzak: There is no transaction wrapping this code?

controller/collaborators.go
```diff
@@ -142,12 +157,59 @@ func (c *CollaboratorsController) AddMany(ctx *app.AddManyCollaboratorsContext)
...
 
+func (c *CollaboratorsController) addContributors(ctx context.Context, byIdentityID uuid.UUID, contributors []*app.UpdateUserID, spaceID string) error {
+	err := c.app.ResourceRepository().CheckExists(ctx, spaceID)
+	if err != nil {
+		if notFound, _ := autherrors.IsNotFoundError(err); notFound {
+			log.Warn(ctx, map[string]interface{}{
+				"err":      err,
+				"space_id": spaceID,
+			}, "unable to add contributors to space resource: resource not found; that's OK for old spaces")
+			// Just log the warning. Old spaces doesn't have any registered resources.
```
sbose78: Nice! This block will be executed until old spaces' migration is done. After that we need to make this log.Error(..)

authorization/role/service/role_management_service.go
```diff
@@ -153,6 +153,34 @@ func (s *roleManagementServiceImpl) Assign(ctx context.Context, assignedBy uuid.
 	return err
 }
 
+func (s *roleManagementServiceImpl) DeleteRoleAssignments(ctx context.Context, byIdentity uuid.UUID, forIdentities []uuid.UUID, resourceID string) error {
```
sbose78: Looks good, later on we would need to expose an rest API for this too.

sbose78:
Approach looks good, @alexeykazakov .

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-512-3`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-512-4`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-512-5`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-512-7`

controller/resource_roles.go
```diff
@@ -33,19 +33,14 @@ func NewResourceRolesController(service *goa.Service, app application.Applicatio
...
-			"err":         err,
-		}, "does not exist")
-		return jsonapi.JSONErrorResponse(ctx, errors.NewNotFoundError("resource_id", ctx.ResourceID))
+		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 
-	roles, err = c.app.RoleManagementService().ListByResource(ctx, ctx.ResourceID)
+	var roles []role.IdentityRole
+
+	roles, err = c.app.RoleManagementService().ListByResource(ctx, currentIdentity.ID, ctx.ResourceID)
```
sbose78: Do we need this change for this PR? I'm banking on this (unchanged) API to easily validate the migration of collaborators :)  

I was using a service account token to call this API, but this change would now need a user token.

controller/collaborators.go
```diff
@@ -114,9 +127,102 @@ func (c *CollaboratorsController) List(ctx *app.ListCollaboratorsContext) error
...
+	// Collect all contributors and admins of the space
+	contributors, err := c.listRoles(ctx, currentIdentity, authorization.SpaceContributorRole, isServiceAccount)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+	for _, c := range contributors {
+		cm[c.IdentityID] = c.Identity
+		collaborators = append(collaborators, c.Identity)
+	}
+	admins, err := c.listRoles(ctx, currentIdentity, authorization.SpaceAdminRole, isServiceAccount)
```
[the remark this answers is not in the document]
alexeykazakov: When in the future we have people assigned to view role we won't use collaborators API anymore.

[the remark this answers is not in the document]
sbose78: OK in that case, you could use `roleManagementService().ListByResource(..)` and get all assignments instead of doing it once for admin, and another time for contributors.

controller/resource_roles.go
alexeykazakov: It's much easier to keep consistency if it's part of this PR. It also won't be very easy to extract this change to another PR. All relevant changed were made in the service and many tests.
I would prefer another tiny PR with support of migration SA here which we will revert after migration.

controller/collaborators.go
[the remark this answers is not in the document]
alexeykazakov: OK. Let me clarify what I meant :)
Yes, currently I want to return only admins and contributors as collaborators. It's technically possible for admins to add viewers though if using assign role API directly instead of collaborators API.
But when we start using assign API in our UI then it won't matter anymore because we will remove the collaborator API.

controller/resource_roles.go
sbose78: I understand, sounds good. I'll make the PR after we merge this.

controller/collaborators.go
alexeykazakov: Each forceAssign will create a new transaction. Not optimal but IMO OK for the collaborators API because we are deprecating it anyway.

Issue #365, opened 2018-03-08
existing spaces' migration : Create corresponding resources for existing spaces


sbose78:
### Task

1. Get list of all users ( we do this for updating tenants )
2. For each user, get all spaces `https://api.openshift.io/api/namedspaces/<username>`
3. For each space returned in (2), call `POST https:/auth.openshift.io/api/resources` 
4. For each space, calls Collaborators API to get list of space members, 
5. Call `Assign Role API` to add those users as 'contributors'. ( Q: to do so whose token should we use ? )

Part of https://github.com/fabric8-services/fabric8-auth/issues/359  

### Dependencies

- [x] #383 
- [x] #498 
- [ ] #512

alexeykazakov:
- Drop existing space_resource table.

sbose78:
Done for prod-preview and prod.

PR #513, opened 2018-06-06
temporarily support space creation using service account

sbose78:
If the space creation API call is made using the `space-migration` service account, then the creator ID is passed as a parameter in the request `/api/spaces/spaceID?creator=creator-uuid`.

In this scenario, there would be no keycloak interactions, or `space_resources` table updation.

This commit can be reverted after migration of spaces is done.

Dependency of #365

### DIFF
--- a/controller/space.go
+++ b/controller/space.go
@@ -11,6 +11,7 @@ import (
 	"github.com/fabric8-services/fabric8-auth/log"
 	"github.com/fabric8-services/fabric8-auth/login"
 	"github.com/fabric8-services/fabric8-auth/space"
+	"github.com/fabric8-services/fabric8-auth/token"
 
 	"github.com/goadesign/goa"
 	"github.com/satori/go.uuid"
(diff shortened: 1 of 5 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-513-1`

controller/space.go
```diff
@@ -47,31 +48,55 @@ func NewSpaceController(service *goa.Service, app application.Application, confi
...
-	if err != nil {
-		return jsonapi.JSONErrorResponse(ctx, err)
-	}
 
-	// Create keycloak resource for this space
-	resource, err := c.resourceManager.CreateResource(ctx, ctx.RequestData, ctx.SpaceID.String(), spaceResourceType, nil, &scopes, currentIdentity.ID.String())
-	if err != nil {
-		return jsonapi.JSONErrorResponse(ctx, err)
+	isMigration := false
+	if token.IsSpecificServiceAccount(ctx, "space-migration") {
```
alexeykazakov: `isMigration :=  token.IsSpecificServiceAccount(ctx, "space-migration")` ;)

controller/space.go
```diff
@@ -47,31 +48,55 @@ func NewSpaceController(service *goa.Service, app application.Application, confi
...
 	}
 
-	err = transaction.Transactional(c.app, func(tr transaction.TransactionalResources) error {
-		// Create space resource which will represent the keycloak resource associated with this space
-		_, err = tr.SpaceResources().Create(ctx, spaceResource)
-		return err
-	})
-	if err != nil {
-		return jsonapi.JSONErrorResponse(ctx, err)
+	if !isMigration {
```
alexeykazakov: You could just move this block to `else {...}` above

PR #514, opened 2018-06-08
Send invitation e-mail to user via notification service

### LINKED ISSUE #507: Use notification service to generate an e-mail upon creating a new user invitation
When a user is invited to an organization, team or security group, an invitation e-mail should be generated by the notification service and sent to the user, allowing them to accept the invitation.

sbryzak:
Fixes #507


sbryzak:
Will merge changes from #529 once it is merged with master, in order to improve notification handling.

sbryzak:
Is the low test coverage possibly because the invitation message processing is done in a separate thread (go process)?  I've modified the code (since we now have a SendMessagesAsync method) to not execute these methods in a separate thread, hopefully that will improve the coverage.

authorization/invitation/repository/invitation.go
```diff
@@ -78,6 +81,8 @@ type InvitationRepository interface {
 
 	ListRoles(ctx context.Context, id uuid.UUID) ([]rolerepo.Role, error)
 	AddRole(ctx context.Context, invitationId uuid.UUID, roleId uuid.UUID) error
+
+	FindByToken(ctx context.Context, identityID uuid.UUID, acceptToken uuid.UUID) (*Invitation, error)
```
sbose78: Could please replace the term `token` by `code` ? Because `token` has a very different meaning in our service :)

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +199,84 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+}
+
+// processTeamInviteNotifications sends an e-mail notification to a user.
+func (s *invitationServiceImpl) processTeamInviteNotifications(ctx context.Context, team *account.Identity, inviterName string,
+	notifications []invitationNotification, witURL string) {
+	teamName := team.IdentityResource.Name
+
+	spaceName, err := lookupSpaceName(ctx, witURL, *team.IdentityResource.ParentResourceID)
+	if err != nil {
+		// What to do here?? just log an error and die?
```
sbose78: Let's not just log and die. Let's return an error all the way up to the controller - and then return an 50x/40x response code because without the email the user has not really been invited to the space at all.

This can occur only in an exceptional case. The user who was inviting should then retry inviting the users to the space. ( we will end up with a bunch of unsent invites in the table, but that's better than silently failing and not asking the user to retry )

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +199,84 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+	s.Services().NotificationService().SendMessagesAsync(ctx, messages)
+}
+
+// processSpaceInviteNotifications sends an e-mail notification to a user.
+func (s *invitationServiceImpl) processSpaceInviteNotifications(ctx context.Context, space *resource.Resource,
+	inviterName string, notifications []invitationNotification, witURL string) {
+
+	spaceName, err := lookupSpaceName(ctx, witURL, space.ResourceID)
+	if err != nil {
+		// What to do here?? just log an error and die?
```
sbose78: Same comment as above.

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +199,84 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+			strings.Join(n.roles, ","),
+			acceptURL))
+	}
+
+	s.Services().NotificationService().SendMessagesAsync(ctx, messages)
+}
+
+// lookupSpaceName talks to the WIT service to retrieve a space record for the specified spaceID, then
+// returns the name of the space
+func lookupSpaceName(ctx context.Context, witURL string, spaceID string) (string, error) {
```
sbose78: Thinking out aloud : 

We should ideally make notification-service do this. I agree that might increase the number of calls to the WIT Space API from the notification-service - but that should be a problem which notification-service should solve ( by doing some local caching? ) :) 

The current approach has the advantage of avoiding unnecessary calls for sure. However, if we had to now pass the description, we would have to make code changes to two codebases :) 

For now, we can go ahead with this.

controller/invitation.go
```diff
@@ -52,7 +56,15 @@ func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationConte
 		})
 	}
 
-	err = c.app.InvitationService().Issue(ctx, currentIdentity.ID, ctx.InviteTo, invitations)
+	witURL, err := c.Configuration.GetWITURL(ctx.RequestData)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err": err,
+		}, "failed to create invitations")
```
sbose78: "failed to get WIT URL"

authorization/invitation/service/invitation_service.go
sbryzak: Since this is no longer running in a separate thread I now return an error

authorization/invitation/service/invitation_service.go
sbryzak: Since this is no longer running in a separate thread I now return an error

authorization/invitation/repository/invitation.go

authorization/invitation/service/invitation_service.go
sbryzak: In this case, it makes more sense to do the space lookup in auth once (refactoring the notification service to support caching seems to be a non-trivial exercise).  I think we should probably determine this on a case by case basis, weighing up the benefits of assembling the notification message package ourselves vs getting the notification service to do it.

controller/invitation.go
sbryzak: Fixed

authorization/invitation/service/invitation_service.go
```diff
@@ -151,6 +166,40 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+			}
+
+			if identityResource.ResourceType.Name == authorization.IdentityResourceTypeTeam {
+				err = s.processTeamInviteNotifications(ctx, inviteToIdentity, inviter.User.FullName, notifications, witURL)
+			}
+		} else if inviteToResource != nil && inviteToResource.ResourceType.Name == authorization.ResourceTypeSpace {
+			err = s.processSpaceInviteNotifications(ctx, inviteToResource, inviter.User.FullName, notifications, witURL)
+		}
+
+		if err != nil {
```
alexeykazakov: I guess we should check if we didn't send any notifications (invateToIdentity == nil & inviteToResource == nil or unknown/unsupported resource type (org or whatever)) and return NotFoundError in that case.

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +211,98 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+		spaceName, err = lookupSpaceName(ctx, witURL, *team.IdentityResource.ParentResourceID)
+		if err != nil {
+			return err
+		}
+	}
+
+	var messages []notification.Message
+
+	for _, n := range notifications {
+		acceptURL := fmt.Sprintf("%s/api/invitations/accept?code=%s", s.config.GetAuthServiceURL(), n.invitation.AcceptCode.String())
```
alexeykazakov: It should actually use accept invitation client path instead of hardcoded string. But the client does not exist yet.
It's important to switch to the goa generated client path as soon as we have the accept invitation controller implemented.
Then it will work even if we update the endpoint path in our design doc later.

authorization/invitation/service/invitation_service.go
```diff
@@ -151,6 +166,40 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+		// 2) Invite user to space, roles only, no organization
+		//
+		if inviteToIdentity != nil {
+			identityResource, err := s.Repositories().ResourceRepository().Load(ctx, inviteToIdentity.IdentityResourceID.String)
+			if err != nil {
+				return err
+			}
+
+			if identityResource.ResourceType.Name == authorization.IdentityResourceTypeTeam {
+				err = s.processTeamInviteNotifications(ctx, inviteToIdentity, inviter.User.FullName, notifications, witURL)
```
alexeykazakov: We can't actually call invitation processing here because we are still in a transaction. The notification service is called asynchronously in a new goroutine but the notification will be sent regardless of success or failure of transaction.
So, here is what may happen:

1. We start transaction to send two invitations
1.1. We create the first invitation in DB
1.2. A new goroutine is created to send the invitation
1.3. We created the second invitation in DB
1.4. Something fails and we return an error.
2. Transaction is rolled back and both Invitation 1 and 2 are not comited to the DB. The invitations do not exist.
3. Meanwhile the invitation 1 was sent to the user (it's in a separate goroutine fully independent from the transaction).
4. User get the invitation but it fails when he tries to accept it because it doesn't exist in DB.

I would suggest create all invitations first in a transaction then in a new loop process them.

Update: You actually call WIT to get Space in the invite notification processing methods. So it's even worse. You cal a remote service within a transaction in the same goroutine. One more reason to move the notification processing out of the transaction :)

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +211,98 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+		messages = append(messages, notification.NewTeamInvitationEmail(n.invitation.Identity.UserID.UUID.String(),
+			teamName,
+			inviterName,
+			spaceName,
+			acceptURL))
+	}
+
+	s.Services().NotificationService().SendMessagesAsync(ctx, messages)
+
+	return nil
```
alexeykazakov: return s.Services().NotificationService().SendMessagesAsync(ctx, messages)
instead of 
return nil
Otherwise you will miss an error returned by notification service.

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +211,98 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+		spaceName, err = lookupSpaceName(ctx, witURL, space.ResourceID)
+		if err != nil {
+			return err
+		}
+	}
+
+	var messages []notification.Message
+
+	for _, n := range notifications {
+		acceptURL := fmt.Sprintf("%s/api/invitations/accept?code=%s", s.config.GetAuthServiceURL(), n.invitation.AcceptCode.String())
```
alexeykazakov: Same as above (path).

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +211,98 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+		messages = append(messages, notification.NewSpaceInvitationEmail(n.invitation.Identity.UserID.UUID.String(),
+			spaceName,
+			inviterName,
+			strings.Join(n.roles, ","),
+			acceptURL))
+	}
+
+	s.Services().NotificationService().SendMessagesAsync(ctx, messages)
+
+	return nil
```
alexeykazakov: Same as above (you will lose the error from the notification service).

authorization/invitation/service/invitation_service.go
```diff
@@ -162,3 +211,98 @@ func (s *invitationServiceImpl) Issue(ctx context.Context, issuingUserId uuid.UU
...
+}
+
+// lookupSpaceName talks to the WIT service to retrieve a space record for the specified spaceID, then
+// returns the name of the space
+func lookupSpaceName(ctx context.Context, witURL string, spaceID string) (string, error) {
```
alexeykazakov: This code should move to our WIT service (wit/wit.go). The WIT service needs to be refactored though. The same refactoring as we did for Notification service.
@sbryzak can you maybe create an issue for that? I will try to do such refactoring this week so you can consume it here.

migration/sql-files/032-invitation-code.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE invitation ADD COLUMN accept_code VARCHAR NOT NULL UNIQUE;
```
alexeykazakov: Will UNIQUE constrain automatically create an index for accept_code?

wit/wit.go
```diff
@@ -47,9 +47,11 @@ func (r *RemoteWITServiceCaller) UpdateWITUser(ctx context.Context, updatePayloa
 	}
 
 	remoteWITService, err := CreateSecureRemoteClientAsServiceAccount(ctx, witURL)
+
```
alexeykazakov: This is the WIT service I was talking about above.

authorization/invitation/service/invitation_service.go
```diff
@@ -14,23 +14,36 @@ import (
...
+	"github.com/fabric8-services/fabric8-auth/notification"
+	"github.com/fabric8-services/fabric8-auth/wit"
+	"github.com/fabric8-services/fabric8-auth/wit/witservice"
+	goauuid "github.com/goadesign/goa/uuid"
 	"github.com/satori/go.uuid"
+	"strings"
 )
 
+type InvitationConfiguration interface {
+	GetAuthServiceURL() string
```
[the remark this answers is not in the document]
sbryzak: I would really like to change this method also, if possible.  The `calculateWITURL` method however still requires the request object in order to calculate the correct scheme to use for the current user.  Do you think we can simply hard-code this to https instead?

authorization/invitation/repository/invitation_blackbox_test.go
```diff
@@ -157,6 +157,20 @@ func (s *invitationBlackBoxTest) TestAddRoleFailsForInvalidRoleID() {
...
+	i := g.CreateInvitation()
+
+	// Create a couple more invitations for some noise
+	g.CreateInvitation()
+	g.CreateInvitation()
+
+	invitation, err := s.repo.FindByAcceptCode(s.Ctx, i.Invitation().IdentityID, i.Invitation().AcceptCode)
+	require.NoError(s.T(), err)
+
+	require.Equal(s.T(), i.Invitation().InvitationID, invitation.InvitationID)
```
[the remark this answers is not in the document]
sbryzak: Done

authorization/invitation/repository/invitation.go
```diff
@@ -245,3 +253,16 @@ func (m *GormInvitationRepository) AddRole(ctx context.Context, invitationId uui
...
+func (m *GormInvitationRepository) FindByAcceptCode(ctx context.Context, identityID uuid.UUID, acceptCode uuid.UUID) (*Invitation, error) {
+	defer goa.MeasureSince([]string{"goa", "db", "invitation", "FindByToken"}, time.Now())
+
+	var native Invitation
+	err := m.db.Table(m.TableName()).Where("identity_id = ? AND accept_code = ?", identityID, acceptCode).Find(&native).Error
+	if err == gorm.ErrRecordNotFound {
+		return nil, errors.NewNotFoundErrorWithKey("invitation", "identity:accept_code", identityID.String()+":"+acceptCode.String())
+	}
+	return &native, errs.WithStack(err)
+	return nil, nil
```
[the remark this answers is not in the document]
sbryzak: Fixed

authorization/invitation/repository/invitation.go
```diff
@@ -245,3 +253,16 @@ func (m *GormInvitationRepository) AddRole(ctx context.Context, invitationId uui
 	}, "Invitation role created!")
 	return nil
 }
+
+// FindByAcceptCode returns the Invitation record for the specified identity, with the specified accept code
+func (m *GormInvitationRepository) FindByAcceptCode(ctx context.Context, identityID uuid.UUID, acceptCode uuid.UUID) (*Invitation, error) {
+	defer goa.MeasureSince([]string{"goa", "db", "invitation", "FindByToken"}, time.Now())
```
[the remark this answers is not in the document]
sbryzak: Fixed

authorization/invitation/service/invitation_service.go
sbryzak: I don't think we should error here, while we only support sending _notifications_ for two particular 'types' of invitations, it should still be possible to send other invitation types.  I've updated the comment to reflect this.

authorization/invitation/service/invitation_service.go
sbryzak: I already had the same thought, and I'll roll this change into the accept invitation PR.

authorization/invitation/service/invitation_service.go
sbryzak: Good catch, I've moved this out of the transaction block.

authorization/invitation/service/invitation_service.go
sbryzak: Fixed

authorization/invitation/service/invitation_service.go
sbryzak: Fixed

wit/wit.go
sbryzak: Understood, let's address this in #536 - I'm happy to work on this issue if you like.

authorization/invitation/service/invitation_service.go
alexeykazakov: Sure but for now we support only these two types. I don't think we should silently ignore other types. Let's return an error if we can't actually sent any invitation because we don't support it.

sbryzak:
For some reason the ci/jenkins build is failing with this error:

```
ERROR: These files differ from gofmt's style (run 'make format-go-code' to fix this):
./test/graph/test_graph.go:67:25: expected type, found '='
make: *** [check-go-format] Error 1
```

However the line of code is perfectly valid:

`type wrapperConstructor = func(g *TestGraph, params []interface{}) interface{}`

Any ideas?

alexeykazakov:
@sbryzak just gofmt style check. Just run "make format-go-code".

sbryzak:
@alexeykazakov I've already done that, there were no changes.

authorization/invitation/service/invitation_service_blackbox_test.go
```diff
@@ -35,29 +36,35 @@ func (s *invitationServiceBlackBoxTest) SetupTest() {
 }
 
 func (s *invitationServiceBlackBoxTest) TestIssueInvitationByIdentityID() {
-	// Create a test user - this will be the organization owner
-	identity, err := test.CreateTestIdentityAndUserWithDefaultProviderType(s.DB, "invitationServiceBlackBoxTest-TestIssuingUser")
-	require.Nil(s.T(), err, "Could not create identity")
+	g := s.NewTestGraph()
```
alexeykazakov: Just a note. You could use already available s.Graph instead of creating it explicitly manually in your tests via s.NewTestGraph(). We create a new graph for every test function in DBSuite - https://github.com/fabric8-services/fabric8-auth/blob/master/gormtestsupport/db_test_suite.go#L70
Example of usage - https://github.com/fabric8-services/fabric8-auth/blob/master/authorization/role/service/role_management_service_blackbox_test.go#L45

migration/sql-files/032-invitation-code.sql
sbose78: @sbryzak this constraint is missing in master, right?

PR #516, opened 2018-06-08
 Allow assignment of role even if user not in space, using service account

sbose78:
During migration, the script will be adding a contributor role for every collaborator in a space. However, the existing design disallows use of Assign API if the user isn't already a part of the space.

This PR adds an exception to the above rule if the service account token is passed. After migration, these changes can be safely reverted.

### DIFF
--- a/controller/resource_roles.go
+++ b/controller/resource_roles.go
@@ -3,11 +3,15 @@ package controller
 import (
 	"github.com/fabric8-services/fabric8-auth/app"
 	"github.com/fabric8-services/fabric8-auth/application"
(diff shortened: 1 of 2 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-516-1`

sbose78:
[test]

alexeykazakov:
I was thinking about using collaborators API instead. Then we can keep all migration stuff within space and collaborators controllers.

PR #521, opened 2018-06-13
Fix collaborators API

alexeykazakov:
We can't check if the current identity has permissions to list collaborators because it breaks the existing collaborators API: Currently Planner requires the list of collaborators to show Planner.

Until Planner is fixed we have to allow any user to list collaborators.

- [x] Fix tests


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-521-1`

Issue #526, opened 2018-06-19
Create Notification Service


sbryzak:
It would be useful to have a dedicated notification service which other services may invoke in order to more easily send a message to the notification service.

PR #529, opened 2018-06-20
Notification Service

alexeykazakov:
- [x] Tests

Fixes https://github.com/fabric8-services/fabric8-auth/issues/526


fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-529-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-529-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-529-3`

application/service/services.go
```diff
@@ -70,6 +71,11 @@ type UserService interface {
 	DeprovisionUser(ctx context.Context, username string) (*account.Identity, error)
 }
 
+type NotificationService interface {
```
sbose78: So, as a service we are enforcing that the notification should always be async? ( Ok with this, but just checking )

sbose78:
Looks good to me, will review the tests once you add them.

application/service/services.go
alexeykazakov: So far we need only async calls. When/if we need sync call then we can add it to the service.

alexeykazakov: So far we need only async calls. When/if we need sync call then we can add it to the service.

alexeykazakov:
@sbose78 @sbryzak ready for review.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-529-4`

application/service/base/base_service.go
```diff
@@ -9,6 +9,6 @@ type BaseService struct {
 	context.ServiceContext
 }
 
-func NewBaseService(context *context.ServiceContext) BaseService {
-	return BaseService{*context}
+func NewBaseService(context context.ServiceContext) BaseService {
```
sbryzak: I'm concerned about this change - if we are passing the service context by value, not by pointer, then won't the state of that context be a copy of the service context parameter's state?  If that is so, then member values such as `transactionManager`, `inTransaction` etc will be different instances to those of the original parameter, meaning that an "inner" service will have a different context to an "outer" service, therefore won't be _aware_ of when the other service has started a transaction?  This would mean that all transactional code in services will be in an isolated transaction, right?

My understanding is based on this:

[http://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/](http://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/)

alexeykazakov: As the result it was impossible to create a service passing nil as context.ServiceContext which was needed for tests.

PR #532, opened 2018-06-26
Forward request ID when calling notification service

alexeykazakov:
I noticed that we didn't set request ID in the notification service. This PR fixes that. 
It's very tricky to test this change though.
In our tests we set context via client package and it works. But in runtime a different key is used to store the request ID in the context (middleware vs. client).

So, no tests unfortunately.

### DIFF
--- a/notification/service/notification.go
+++ b/notification/service/notification.go
@@ -79,7 +79,7 @@ func (s *notificationServiceImpl) send(ctx context.Context, c *client.Client, ms
 	msgID := uuid.UUID(msg.MessageID)
 
 	resp, err := c.SendNotify(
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-532-1`

PR #538, opened 2018-07-02
Implement accept invitation endpoint

### LINKED ISSUE #531: Create accept invitation endpoint
This endpoint will be used to process requests from users that click the accept invitation link in their invitation e-mail to join a team or a space.  The link should contain the unique token for their invitation, and be used to locate the invitation record for that user.  Once the invitation has been successfully found, the membership/role assignments for that invitation should be converted into actual membership/role assignments for the user.  After the invitation has been accepted, the user should be redirected to the URL of the space which they are joining.

sbryzak:
Fixes #531 

This PR adds a new endpoint for accepting an invitation.

### DIFF
--- a/account/repository/identity.go
+++ b/account/repository/identity.go
@@ -130,13 +130,22 @@ type IdentityRepository interface {
 	Search(ctx context.Context, q string, start int, limit int) ([]Identity, int, error)
 	FindIdentityMemberships(ctx context.Context, identityID uuid.UUID, resourceType *string) ([]authorization.IdentityAssociation, error)
 	FindIdentitiesByResourceTypeWithParentResource(ctx context.Context, resourceTypeID uuid.UUID, parentResourceID string) ([]Identity, error)
+	AddMember(ctx context.Context, identityID uuid.UUID, memberID uuid.UUID) error
 }
 
 // TableName overrides the table name settings in Gorm to force a specific table name
 // in the database.
 func (m *GormIdentityRepository) TableName() string {
 	return "identities"
+}
 
+type Membership struct {
+	MemberID uuid.UUID `sql:"type:uuid" gorm:"primary_key"`
+	MemberOf uuid.UUID `sql:"type:uuid" gorm:"primary_key"`
+}
+
+func (m *GormIdentityRepository) MembershipTableName() string {
+	return "membership"
 }
 
 // CRUD Functions
@@ -522,4 +531,49 @@ func (m *GormIdentityRepository) FindIdentitiesByResourceTypeWithParentResource(
 	}
 
 	return identities, nil
+}
+
+func (m *GormIdentityRepository) AddMember(ctx context.Context, identityID uuid.UUID, memberID uuid.UUID) error {
+	defer goa.MeasureSince([]string{"goa", "db", "identity", "AddMember"}, time.Now())
+
+	var identity Identity
+	err := m.db.Table(m.TableName()).Preload("IdentityResource").Where("id = ?", identityID).Find(&identity).Error
+	if err == gorm.ErrRecordNotFound {
+		return errs.WithStack(errors.NewNotFoundError("identity", identityID.String()))
+	}
+
+	if !identity.IdentityResourceID.Valid {
+		return errs.WithStack(errors.NewBadParameterErrorFromString("identityID", identityID.String(), "Specified identity has no corresponding resource"))
+	}
+
+	if !authorization.CanHaveMembers(identity.IdentityResource.ResourceType.Name) {
+		return errs.WithStack(errors.NewBadParameterErrorFromString("identityID", identityID.String(), "Specified identity may not have members"))
+	}
+
+	var member Identity
+	err = m.db.Table(m.TableName()).Where("id = ?", memberID).Find(&member).Error
+	if err == gorm.ErrRecordNotFound {
+		return errs.WithStack(errors.NewNotFoundError("identity", memberID.String()))
+	}
+
+	membership := &Membership{
+		MemberOf: identityID,
+		MemberID: memberID,
+	}
+
+	err = m.db.Create(membership).Error
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"member_of": identityID,
+			"member_id": memberID,
+			"err":       err,
+		}, "unable to create the membership")
+		return errs.WithStack(err)
+	}
+	log.Info(ctx, map[string]interface{}{
+		"member_of": identityID,
+		"member_id": memberID,
+	}, "Membership created!")
+
+	return nil
 }
--- a/account/repository/identity.go
+++ b/account/repository/identity.go
@@ -144,7 +144,7 @@ type Membership struct {
 	MemberOf uuid.UUID `sql:"type:uuid" gorm:"primary_key"`
 }
 
-func (m *GormIdentityRepository) MembershipTableName() string {
+func (m Membership) TableName() string {
 	return "membership"
 }
 
@@ -537,7 +537,7 @@ func (m *GormIdentityRepository) AddMember(ctx context.Context, identityID uuid.
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "AddMember"}, time.Now())
 
 	var identity Identity
-	err := m.db.Table(m.TableName()).Preload("IdentityResource").Where("id = ?", identityID).Find(&identity).Error
+	err := m.db.Table(m.TableName()).Preload("IdentityResource").Preload("IdentityResource.ResourceType").Where("id = ?", identityID).Find(&identity).Error
 	if err == gorm.ErrRecordNotFound {
 		return errs.WithStack(errors.NewNotFoundError("identity", identityID.String()))
 	}
--- a/authorization/invitation/repository/invitation.go
+++ b/authorization/invitation/repository/invitation.go
@@ -192,6 +192,26 @@ func (m *GormInvitationRepository) Delete(ctx context.Context, id uuid.UUID) err
 
 	obj := Invitation{InvitationID: id}
 
+	var invitationRoles []InvitationRole
+
+	err := m.db.Where("invitation_id = ?", id).Find(&invitationRoles).Error
+	if err != nil && err != gorm.ErrRecordNotFound {
+		return errs.WithStack(err)
+	}
+
+	for _, r := range invitationRoles {
+		result := m.db.Delete(&r)
+
+		if result.Error != nil {
+			log.Error(ctx, map[string]interface{}{
+				"invitation_id": id,
+				"role_id":       r.RoleID.String(),
+				"err":           result.Error,
+			}, "unable to delete invitation role ")
+			return errs.WithStack(result.Error)
+		}
+	}
+
 	result := m.db.Delete(&obj)
 
 	if result.Error != nil {
--- a/authorization/invitation/service/invitation_service.go
+++ b/authorization/invitation/service/invitation_service.go
@@ -334,9 +334,10 @@ func lookupSpaceName(ctx context.Context, witURL string, spaceID string) (string
 
 // Accept processes an invitation acceptance click, and returns the resource ID of the resource or identity resource which the invitation is for
 func (s *invitationServiceImpl) Accept(ctx context.Context, currentIdentityID uuid.UUID, token uuid.UUID) (string, error) {
-	var resourceID string
+	/*var resourceID string
 
 	inv, err := s.Repositories().InvitationRepository().FindByAcceptCode(ctx, currentIdentityID, token)
+
 	if err != nil {
 		return resourceID, err
 	}
@@ -355,6 +356,6 @@ func (s *invitationServiceImpl) Accept(ctx context.Context, currentIdentityID uu
 	} else if inv.ResourceID != nil {
 
 	}
-
+	*/
 	return "", nil
 }
--- a/authorization/invitation/service/invitation_service.go
+++ b/authorization/invitation/service/invitation_service.go
@@ -366,7 +366,7 @@ func (s *invitationServiceImpl) Accept(ctx context.Context, currentIdentityID uu
 			ir := &repository.IdentityRole{
 				IdentityID: currentIdentityID,
 				RoleID:     role.RoleID,
-				ResourceID: inviteToIdentity.IdentityResource.ResourceID,
+				ResourceID: inviteToIdentity.IdentityResourceID.String,
 			}
 
 			err = s.Repositories().IdentityRoleRepository().Create(ctx, ir)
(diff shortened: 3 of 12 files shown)


authorization/invitation/service/invitation_service.go
```diff
@@ -331,3 +332,85 @@ func lookupSpaceName(ctx context.Context, witURL string, spaceID string) (string
...
+			}
+
+			err = s.Repositories().IdentityRoleRepository().Create(ctx, ir)
+			if err != nil {
+				return resourceID, err
+			}
+		}
+
+		// Delete the invitation
+		s.Repositories().InvitationRepository().Delete(ctx, inv.InvitationID)
```
sbose78: Since a user could have been sent out multiple invites, this needs to delete "All invites to user X who have been invited with Roles Y,Z...". This is to ensure that every other invite mail should lead to "this invite is no longer valid because..."

authorization/invitation/service/invitation_service.go
```diff
@@ -331,3 +332,85 @@ func lookupSpaceName(ctx context.Context, witURL string, spaceID string) (string
...
```
sbose78: Same as above.

authorization/invitation/service/invitation_service.go
sbryzak: The accept invitation request passes a single acceptance code, which itself belongs to a single invitation.  Once the invitation has been accepted, that record alone needs to be removed, there are no circumstances under which we should be deleting any other invitations here.

sbryzak: On a related note however, we do need to be deleting the `invitation_role` records - I just checked and it's not being done.  I'll fix that up in the invitation repository.

sbose78: > The accept invitation request passes a single acceptance code, which itself belongs to a single invitation. Once the invitation has been accepted, that record alone needs to be removed, there are no circumstances under which we should be deleting any other invitations here.

If multiple (re)invites were sent to the same user for the same set of roles, and one of them was accepted by the user - how do we ensure we don't land up with many many unaccepted invites ?

sbryzak: If a reinvite is issued for the same team, org or resource, then the same invitation should be sent again.  In fact, we should enforce this via a database constraint, so we should probably create a new issue for this.

sbryzak: Another thing we don't have is a "Cancel Invitation" endpoint, we likely need that also.  I'll create a new issue for this, and the DB constraint.

sbryzak: Created #543

sbose78: Having it as a db constraint is a good idea. One advantage of allowing multiple invitations in the db is that it helps us record the number of (re)invitations that were created. If you feel we should go with the db constraint, I'm with you on that. Later on, if the need for recording sent invites shows up, we can fix that.

sbryzak: Yes, I believe that a reinvite should simply send another notification for the existing invitation.

authorization/invitation/repository/invitation.go
```diff
@@ -192,6 +192,26 @@ func (m *GormInvitationRepository) Delete(ctx context.Context, id uuid.UUID) err
...
 	obj := Invitation{InvitationID: id}
 
+	var invitationRoles []InvitationRole
+
+	err := m.db.Where("invitation_id = ?", id).Find(&invitationRoles).Error
+	if err != nil && err != gorm.ErrRecordNotFound {
+		return errs.WithStack(err)
+	}
+
+	for _, r := range invitationRoles {
```
sbose78: Could you please add/update the repo test to validate deletion of invitationRoles too?

authorization/invitation/service/invitation_service_blackbox_test.go
```diff
@@ -459,3 +460,89 @@ func (s *invitationServiceBlackBoxTest) TestIssueSpaceInvite() {
...
+	inv := s.Graph.CreateInvitation(space, user, spaceRole)
+
+	otherUser := s.Graph.CreateUser()
+
+	_, err := s.Application.InvitationService().Accept(s.Ctx, otherUser.IdentityID(), inv.Invitation().AcceptCode)
+	require.Error(s.T(), err)
+	require.IsType(s.T(), errors.NotFoundError{}, err)
+}
+
+func (s *invitationServiceBlackBoxTest) TestAcceptFailsForUnknownAcceptCode() {
```
sbose78: Could you please add a test for "Accept code already used" ?

controller/invitation_blackbox_test.go
```diff
@@ -245,6 +246,33 @@ func (s *TestInvitationREST) TestCreateOrganizationInvalidUserInvitation() {
...
+
+	test.AcceptInviteInvitationTemporaryRedirect(s.T(), service.Context, service, controller, inv.Invitation().AcceptCode.String())
+
+	// The invitation should no longer be there after acceptance
+	_, err := s.Application.InvitationRepository().FindByAcceptCode(s.Ctx, s.testIdentity.ID, inv.Invitation().AcceptCode)
+	require.Error(s.T(), err)
+	require.IsType(s.T(), errors.NotFoundError{}, err)
+}
+
+func (s *TestInvitationREST) TestAcceptInvitationFailsForInvalidCode() {
```
sbose78: Would be nice to validate the response in case the same code is being re-used - could you please add a test for the same?

controller/invitation.go
```diff
@@ -73,3 +73,28 @@ func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationConte
...
+	}
+
+	c.app.InvitationService().Accept(ctx, currentIdentity.ID, acceptCode)
+
+	log.Debug(ctx, map[string]interface{}{
+		"accepting-user-id": *currentIdentity,
+		"accept-code":       ctx.AcceptCode,
+	}, "invitation accepted")
+
+	return ctx.TemporaryRedirect()
```
sbose78: Set the `Location` header for the redirect? Add the redirect url to configuration, and also set it in the configmap https://github.com/fabric8-services/fabric8-auth/blob/master/openshift/auth.config.yaml#L53 . We could keep the redirect url as `/_home` for now. Could you please assign an issue to @rohitkrai03 and work with UX to get the page designed?

fyi: This is what we use for the email verification redirect https://console.rh-idev.openshift.com/console/project/dsaas-preview/browse/config-maps/auth

controller/invitation.go
```diff
@@ -73,3 +73,28 @@ func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationConte
 
 	return ctx.Created()
 }
+
+func (c *InvitationController) AcceptInvite(ctx *app.AcceptInviteInvitationContext) error {
+	currentIdentity, err := login.LoadContextIdentityIfNotDeprovisioned(ctx, c.app)
```
sbose78: We could keep it along the same lines as 'email verification'. This endpoint should not be protected by a jwt. The user needn't be logged in. The 'code' will indirectly vouch for the validity and then the identity should be looked up using the acceptCode ( in the database ).

authorization/invitation/repository/invitation.go

authorization/invitation/service/invitation_service_blackbox_test.go
sbryzak: What should this test do?  Attempt to create a second invitation with the same accept code?

controller/invitation_blackbox_test.go
sbryzak: Do you mean add a test to try and accept an invitation that has already been accepted?

controller/invitation.go
sbryzak: So can anyone with the code accept the invitation?  The invitation repository has the following method for locating the accept code:

`InvitationRepository().FindByAcceptCode(ctx, currentIdentityID, token)`

This requires the user for which the invitation is for to be logged in, as an added level of protection.  If we wish to change this behaviour, I'll need to rewrite these methods and their tests.

authorization/invitation/service/invitation_service_blackbox_test.go
sbose78: Attempt to accept the same invitation code twice.

controller/invitation_blackbox_test.go
sbose78: Yes

controller/invitation.go
sbose78: The code is *like* a secret token which should provide the necessary protection. In any case, whoever clicks on the 'accept' link, it is always going to associate the 'role' only with the identities originally intended - so we should be good. Would be great if you could re-write this to not need jwt.

authorization/invitation/service/invitation_service.go
sbose78: Cool, let's roll with that, then :+1:

authorization/invitation/service/invitation_service_blackbox_test.go
sbryzak: Done

controller/invitation_blackbox_test.go
sbryzak: Same as previous comment.

controller/invitation.go
sbryzak: I've introduced a new configuration parameter, `invitation.accepted.url` which can be used to configure where the user will be redirected to after accepting an invitation.

sbose78: Thanks, looks good !
> Add the redirect url to configuration, and also set it in the configmap https://github.com/fabric8-services/fabric8-auth/blob/master/openshift/auth.config.yaml#L53 

 Could you please also update the  openshift templates? I've opened a housekeeping issue for this  : https://gitlab.cee.redhat.com/dtsd/housekeeping/issues/2114 . Please ensure we merge that, before the auth PR :) 

Btw, for a separate discussion: https://github.com/fabric8-services/fabric8-auth/issues/546

controller/invitation.go
```diff
@@ -73,3 +81,29 @@ func (c *InvitationController) CreateInvite(ctx *app.CreateInviteInvitationConte
...
+	}
+
+	c.app.InvitationService().Accept(ctx, currentIdentity.ID, acceptCode)
+
+	log.Debug(ctx, map[string]interface{}{
+		"accepting-user-id": *currentIdentity,
+		"accept-code":       ctx.AcceptCode,
+	}, "invitation accepted")
+
+	ctx.ResponseData.Header().Set("Location", c.config.GetInvitationAcceptedRedirectURL())
```
sbose78: I missed passing on this review feedback to you @sbryzak - sorry about that.

When there's a success in the 'Accept invite' controller, we redirect to `c.config.GetInvitationAcceptedRedirectURL()`
When there's an error in the 'Accept invite' controller, we still *redirect* to `c.config.GetInvitationAcceptedRedirectURL()+"?error="+errorString`  but with this extra error parameter in the end.

We need to do a 
```
_,err = c.app.InvitationService().Accept(ctx, currentIdentity.ID, acceptCode)
```
and then if `err != nil` we need to pass the `error` URL parameter. That way, the UI can show the user a failure/success.

Could you please make the above changes, and also update the tests to validate the presence of the "Location" header, with an `error` URL parameter conditionally?

controller/invitation.go
sbose78: ignore: https://github.com/fabric8-services/fabric8-auth/pull/547/files

authorization/invitation/repository/invitation.go
```diff
@@ -192,6 +192,26 @@ func (m *GormInvitationRepository) Delete(ctx context.Context, id uuid.UUID) err
 
 	obj := Invitation{InvitationID: id}
 
+	var invitationRoles []InvitationRole
+
+	err := m.db.Where("invitation_id = ?", id).Find(&invitationRoles).Error
+	if err != nil && err != gorm.ErrRecordNotFound {
```
sbose78: You have a test missing for `Invitation not found`.

account/repository/identity.go
```diff
@@ -523,3 +532,48 @@ func (m *GormIdentityRepository) FindIdentitiesByResourceTypeWithParentResource(
...
+	err := m.db.Table(m.TableName()).Preload("IdentityResource").Preload("IdentityResource.ResourceType").Where("id = ?", identityID).Find(&identity).Error
+	if err == gorm.ErrRecordNotFound {
+		return errs.WithStack(errors.NewNotFoundError("identity", identityID.String()))
+	}
+
+	if !identity.IdentityResourceID.Valid {
+		return errs.WithStack(errors.NewBadParameterErrorFromString("identityID", identityID.String(), "Specified identity has no corresponding resource"))
+	}
+
+	if !authorization.CanHaveMembers(identity.IdentityResource.ResourceType.Name) {
```
sbose78: Would it be possible to add a test for this scenario?

sbose78:
Good job @sbryzak with multiple conditional blocks in various parts of the code to handle all possible scenarios. Apart from the missed tests for a few conditional blocks, and the `Location` header changes, this PR looks great.

controller/invitation.go
sbryzak: Done

authorization/invitation/repository/invitation.go
sbryzak: Test added.

controller/invitation.go
sbryzak: I'm still not convinced this would be a good thing - without requiring the user to be authenticated, anyone with the accept code could accept the invitation on another user's behalf.  Also, the most obvious landing page after accepting the invitation would be a redirect to the space page  (which we only support so far in the form of team and space invites) which requires the user to be authenticated anyway.  I can't see any benefits from removing the added protection of requiring the user to be already authenticated, also to change this will require changing the service method, repository find method, rewriting tests, etc.

sbryzak: I've made this change, however I'm thinking that it might be useful to have two separate URLs configured for this - one for success, one for failure.  For the success, it is probably useful to include the resource ID of the space, assuming that we will be redirecting to the space page.  For a failure, we probably don't want to redirect to the space page at all.  Perhaps @rohitkrai03 could describe what would be best here?

sbose78: Thanks, I would suggest we always redirect to the same url. On success, the UI can take care of rendering the needful, to provide the user an option to navigate to the space page.

On success, we won't be passing the `error` param. So everything passed as URL params can be used to render a nice success message - and then in the same page we could have something like a "Congratulations, you've accepted the invite, CLICK HERE to visit Space X"

On failure, we *will* be passing be the `error` param. So everything passed as URL params can be used to render a nice 'did not succeed' page.

sbose78: @sbryzak , you also need to add the redirect in case of auth error / invalid jwt here

```
        currentIdentity, err := login.LoadContextIdentityIfNotDeprovisioned(ctx, c.app)
	if err != nil {
                // redirect here as well. If you return a 401/500, the user would see some plain text json  		
	}
```

sbryzak: I overlooked this one, thanks.  Fixed now.

PR #549, opened 2018-07-10
Updated SHA checksum for dep-darwin-amd64

rohitkrai03:
- For macOS dep package was updated with a new SHA. This SHA value is checked in `Makefile` with a hardcoded SHA checksum to verify dep package. 
- Updated the SHA value to new one.
- Added initialization of `$(UNAME_S)` variable `UNAME_S=$(shell uname -s)` since it was being used but never set which lead to explicitly exporting the variable from shell.

Note: We can probably look into automating the process where SHA value is fetched dynamically instead of hard coding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ DOCKER_COMPOSE_BIN := $(shell command -v $(DOCKER_COMPOSE_BIN_NAME) 2> /dev/null
 DOCKER_BIN := $(shell command -v $(DOCKER_BIN_NAME) 2> /dev/null)
 ASCIIDOCTOR_BIN := $(shell command -v $(ASCIIDOCTOR_BIN_NAME) 2> /dev/null)
 
+# Define and get the vakue for UNAME_S variable from shell
+UNAME_S := $(shell uname -s)
+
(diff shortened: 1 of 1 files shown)


sbose78:
[test]

Makefile
```diff
@@ -244,7 +244,7 @@ $(DEP_BIN):
 ifeq ($(UNAME_S),Darwin)
```
dipak-pawar: I can't see where `$(UNAME_S)` is set if we are calling `make build` goal. I believe we should set it so that user doesn't have to explicitly set using `export UNAME_S=Darwin`. So the simple thing to do is define `UNAME_S=$(shell uname -s)`

dipak-pawar:
@rohitkrai03 As we are already fixing it here, Can we fix for fabric8-wit as I can see same issue there as well. See [here](https://github.com/fabric8-services/fabric8-wit/blob/master/Makefile#L252)

Makefile
rohitkrai03: @dipak-pawar Added UNAME_S variable initialization.

rohitkrai03:
@dipak-pawar Will raise a PR for the same in fabric8-wit as well.

PR #550, opened 2018-07-12
Pass service account token to notification service client/req

sbose78:
Notification service needs to be called using a service account token, instead of a user account token - for the user's email updation verification email.

This is needed because of notification-service side changes in https://openshift.io/openshiftio/openshiftio/plan/detail/2646


sbose78:
[test]

Issue #556, opened 2018-07-19
Should we add 'new' assignees and ignore the existing ones..


sbose78:
In our `Add Assignee` API request with a payload of multiple assignments, if one or more users already have the specified role(s) for the resource, we return a non 200 response and ignore the legitimate role assignments that could have taken place.

This is leading to this situation https://github.com/openshiftio/openshift.io/issues/4011#issuecomment-406180012

Should we instead do the legitimate role assignments , and ignore the ones already added?
Should we refactor our API to assign roles to those users who don't already have the role ? and not return an error in case somebody already has a role ?

@sbryzak

sbose78:
https://github.com/openshiftio/openshift.io/issues/4011#issuecomment-406267794

I'm gonna go ahead and modify the API and return a 200 OK even if one assignment is possible, even though others are not possible because the assignments already exist.

FYI the 'Add many collaborators' API uses the `ForceAssign()` from the roles service.

Example, 
If out of 4 requested assignments, 1 is a legit case, and 3 already exist, the API should return a 200 with one assignment successful.
If out of N requested assignments, N already exist, the API would return a 409 conflict ( or anything which says 'resource already exists' )

sbryzak:
@sbose78 That approach seems sensible to me, especially after reading the comments on the other issue.

sbose78:
I'm making this change only for the 'Add collaborators' API for now.

PR #558, opened 2018-07-23
If user is already a collaborator, log and move on

sbose78:
The `Add Collaborators` used to return an error ( non 200 ) if one or more users being added was already a contributor. 

This PR ensures that if a user is already a contributor, the API does not complain and returns a 200, after logging the info that the user is already a contributor. This is analogous to any update API which overwrites existing data even if nothing has changed.

Fixes https://github.com/fabric8-services/fabric8-auth/issues/556
Related to https://github.com/openshiftio/openshift.io/issues/4011

### DIFF
--- a/authorization/role/service/role_management_service_blackbox_test.go
+++ b/authorization/role/service/role_management_service_blackbox_test.go
@@ -356,6 +356,24 @@ func (s *roleManagementServiceBlackboxTest) TestAssignRoleAsAdminOK() {
 	s.checkRoleAssignments([]uuid.UUID{spaceCreator.Identity().ID}, "admin", newSpace.SpaceID())
 }
 
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleAsAdminFailsExistingAssignment() {
+	g := s.DBTestSuite.NewTestGraph()
+	newSpace := g.CreateSpace()
+	spaceCreator := g.CreateUser()
+	s.addNoisyAssignments()
+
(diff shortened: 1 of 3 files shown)


jiekang:
Shouldn't the API still return some kind of notice that the result was not a successful addition, but the case where collaborator already exists? I'm not sure why returning an error code is bad as the UX can choose how to handle it appropriately, whether to ignore it or to emit some notification to the user.

sbose78:
If 3 out of 5 collaborators were already present, and 2 were added anew, that is considered a success, and should be a 200 ok.

In future, we would consider adding a response body with the information about which collaborators were already present. But for now, the UI doesn't really care. Hence we'll add if absent, ignore if already present.

jiekang:
Ah; I did not consider that this supported addition of multiple collaborators at the same time. Sorry about that. This makes sense then, thanks.

authorization/role/service/role_management_service_blackbox_test.go
```diff
@@ -356,6 +356,24 @@ func (s *roleManagementServiceBlackboxTest) TestAssignRoleAsAdminOK() {
 	s.checkRoleAssignments([]uuid.UUID{spaceCreator.Identity().ID}, "admin", newSpace.SpaceID())
 }
 
+func (s *roleManagementServiceBlackboxTest) TestAssignRoleAsAdminFailsExistingAssignment() {
+	g := s.DBTestSuite.NewTestGraph()
```
sbryzak: Unnecessary line, just use `s.Graph`

Issue #559, opened 2018-07-23
Refactor RemoteWITService 


dipak-pawar:
Currently, we have RemoteWITService implementation which is not as per other available services. Also due to which we can't test other services which are dependent on `RemoteWITService`. 

Refactor it such a way that we can test it and avoid duplicate logic.

Issue #536, opened 2018-06-28
Add lookup space name method to WIT API


sbryzak:
Currently we have a `lookupSpaceName` method added to the invitation service in PR #514, however we should add this to the WIT API instead. (`/wit/wit.go`)

Issue #546, opened 2018-07-09
On acceptance of invitation, should we redirect to 'resource' home?


sbose78:
Example, On clicking on 'accept invite' to space/team link in my email, do you think we should redirect the user to the corresponding space home? Basically we would need to construct URLs based on the space name, user owning the space. 

Example, the space https://openshift.io/openshiftio/openshift_io/ is basically https://openshift.io/:owner/:spaceName

@sbryzak what are your thoughts on this?
@rohitkrai03 FYI. since you are gonna work with UX to get a design of 'invitation acceptance' page. We need to check what are their views on this.

sbryzak:
I agree with this, it makes sense to me that accepting an invite to contribute a space or join a team would then take you to the space home page.

sbose78:
Thanks @sbryzak . I was thinking along the same lines as well.

Effectively,
1. Check if resource type is space, if yes, call the `spaces API` to figure out the owner.
2. Check if resource type is team, if yes, lookup the parent resource ( aka space ), and then call the spaces API to figure out the owner.

Once the space owner is figured out, based on `config.GetEnvironment()` , construct the space URL ,
https://[environment]openshift.io/:ownerUserName/:spaceName

sbose78:
@rohitkrai03  that is correct.

rohitkrai03:
@sbose78 But what about the cases like 
- The user clicks on the link where his invitation has been rescinded.
- The user clicks on the link for second time where he has already accepted the invitation earlier.

sbose78:
> The user clicks on the link where his invitation has been rescinded.
> The user clicks on the link for second time where he has already accepted the invitation earlier.

@rohitkrai03 Great question, sorry I wasn't clear. We would still need a proper redirected page to show that the invitation has expired.

rohitkrai03:
@sbose78 So, for successful requests we redirect user to the space and for failed ones we show a landing page with error. Got it. Thanks.

PR #560, opened 2018-07-23
refactor(#559): refactor RemoteWITService as per new design of service

dipak-pawar:
*Changes Proposed in this PR*
* Adds RemoteWITService as new service under all available services for application.
* Renamed `lookupSpaceName` to `GetSpaceNameAndOwnedBy` and moved it under `wit_service.go`
* Modified tests as per new logic

Fixes: #559, #536


application/service/factory/service_factory.go
```diff
@@ -139,7 +140,7 @@ func (f *ServiceFactory) OrganizationService() service.OrganizationService {
 }
 
 func (f *ServiceFactory) InvitationService() service.InvitationService {
-	return invitationservice.NewInvitationService(f.getContext(), f.config)
+	return invitationservice.NewInvitationService(f.getContext(), f.config, f.getContext().Services().WITService())
```
sbryzak: If `WITService` is now available via the service layer, we don't need to pass in the parameter here - simply call `s.Services.WitService()` wherever it is needed.

sbryzak: I should also add that this is the _proper_ way to do service to service interaction in the auth service, as going via the services interface gives us automatic transaction control (which is not needed in this case, however is a good practice to follow).

dipak-pawar: @sbryzak we can simply call `s.Services.WitService()`, But in that case we can't assign `dummyWITService` for it in case of testing. this is why I am passing it as parameter.

sbose78: I understand, mocking the service is the concern. @dipak-pawar , could you explore the possibility of using a mock'd service without having to pass it as a parameter? We are gonna need that anyway soon in a many many places.

application/service/services.go
```diff
@@ -79,6 +79,12 @@ type NotificationService interface {
 	SendMessagesAsync(ctx context.Context, messages []notification.Message) error
 }
 
+type WITService interface {
+	UpdateWITUser(ctx context.Context, updatePayload *app.UpdateUsersPayload, identityID string) error
+	CreateWITUser(ctx context.Context, identity *account.Identity, identityID string) error
+	GetSpaceNameAndOwnedBy(ctx context.Context, spaceID string) (name, ownedBy string, e error)
```
sbose78: Very soon we might need more info from the space service - let's return a struct here. 
```
type Space struct {
	ID              uuid.UUID
	Name            string
	Description     string
	OwnerID         uuid.UUID 
}
```

application/service/factory/service_factory.go
sbose78: @dipak-pawar , you could try doing a 
```
type TestGormDB struct {
    GormDB
}

func (testGormDB *TestGormDB) () service.WITService {
    return myMockWITService
}
```

Or, as @sbryzak suggested
do something like https://github.com/fabric8-services/fabric8-auth/blob/master/notification/service/notification.go#L42

dipak-pawar: @sbryzak, We can use this as defined in [notification.go](https://github.com/fabric8-services/fabric8-auth/blob/master/notification/service/notification.go#L42). But with this way, you can't provide a dynamic mock of dependent services for that service. I would keep this simple for now. Maybe I will try to find more generic solution for this.

@sbose78 I am not sure about approach you suggested will work or not. I will try this in next PR when needed.

dipak-pawar:
@sbryzak `dummyWITService` requirement is for work which I am doing for #546

xcoulon:
@dipak-pawar @sbryzak @sbose78  I opened https://github.com/dipak-pawar/fabric8-auth/pull/1 to fix the issue with `TestRunInvitationServiceBlackBoxTest/TestIssueInvitationOKForResource`. Please let me know if this approach is ok for you

dipak-pawar:
@sbose78, @sbryzak I have addressed review comments. Please have a look into it.

sbose78:
@dipak-pawar  I've addressed some of my comments in https://github.com/dipak-pawar/fabric8-auth/pull/2

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-560-7`

xcoulon:
@dipak-pawar @sbose78 @sbryzak on WIT/Toggles Services/Tenant, we've used https://github.com/dnaeon/go-vcr/ to handle requests and responses to remote services and it's been quite helpful to test multiple scenarii, including `4xx` and `500` errors returned by the remote service: https://github.com/fabric8-services/fabric8-toggles-service/blob/master/controller/features_test.go#L181-L190. The idea would be to use the same `wit` client, but with a different HTTP transport layer (handled by `go-vcr` in that case). 

We even have a custom matcher based on the user's token in the request: https://github.com/fabric8-services/fabric8-toggles-service/blob/master/controller/features_test.go#L677-L705, which would certainly deserve to be moved in https://github.com/fabric8-services/fabric8-common.

sbose78:
That looks good @xcoulon , let's try that @dipak-pawar if that makes it easier for you now.

alexeykazakov:
I don't see any tests for the WIT service.

alexeykazakov:
@xcoulon go-vcr looks nice. Currently we are using our own home-made tiny test lib to mock the http client/transport and to test both incoming requests and responses - https://github.com/fabric8-services/fabric8-auth/blob/master/notification/service/notification_whitebox_test.go#L111

dipak-pawar:
@alexeykazakov, Before refactoring there were no tests. now I have added tests around `WITService` using the same approach as we are using in auth.

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-560-9`

dipak-pawar:
@sbose78, I have tested this PR locally to make sure that we have fresh `user` and `identity` in `auth` and `wit`

**Steps followed**
* Deploy `auth`, `WIT`, `UI` on minishift.
* After sucessfully login from UI connect to auth DB and WIT DB.
* Verified that we have new `user` and `identity` created in both DB using `select * from users;` and `select * from identities`

sbose78:
Thanks @dipak-pawar !

test/witservice.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	account "github.com/fabric8-services/fabric8-auth/account/repository"
+	"github.com/fabric8-services/fabric8-auth/app"
+	"github.com/fabric8-services/fabric8-auth/wit"
+	"github.com/goadesign/goa/uuid"
+)
+
+// DevWITService is the default dev service implementation for WIT.
```
xcoulon: I think it would be better to use `go-vcr` (or another http mocking lib) so that we don't need to introduce such a `DevWITService` in our tests. Instead, we could try to configure the `Application` with a custom HTTP client that would be used for all remote communications (here, `WIT`) and the tests would only have to provide a set of HTTP interactions (if we use go-vcr, for example). WDYT?

dipak-pawar:
@xcoulon we can use `go-vcr`. But how about creating a new issue for it and work on it, instead of adding  it as a part of this PR.

xcoulon:
@dipak-pawar yes, I'm fine with introducing go-vcr in another PR.

test/witservice.go
dipak-pawar: opened https://github.com/fabric8-services/fabric8-auth/issues/575

dipak-pawar:
@xcoulon I have addressed your comments

xcoulon:
@dipak-pawar I approved the PR but you'll need to fix the format on `./authorization/invitation/service/invitation_service_blackbox_test.go` before so the build can pass.

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-560-11`

PR #565, opened 2018-07-27
oauth2.0 Return 403 forbidden is user not provisioned on openshift

sbose78:


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-565-1`

Issue #587, opened 2018-08-09
Add templates and documentaion to run notification svc locally


dipak-pawar:
Currently, we have documentation to setup `auth`, `wit`, `f8ui` to deploy it on minishift. However it'll be really helpful to have documentation and make tasks to deploy notification on minishift. 

This way we can test notifcation related work locally as well.

PR #588, opened 2018-08-09
docs(#587): add documentation for deploying notification on minishift

dipak-pawar:

- Adds openshift template for notification
- Adds make target for `deploy-notification`, `clean-notification`.
- Adds documentation about the same

Fixes: #587

### DIFF
--- a/.make/Makefile.dev
+++ b/.make/Makefile.dev
@@ -36,6 +36,11 @@ deploy-wit: deploy-wit-secrets deploy-wit-db
 	@echo "Deploying wit service"
 	@oc process -f $(TEMPLATE_DIR)/wit.app.yaml -o yaml | oc apply -f -
 
+.PHONY: deploy-notification
+deploy-notification: deploy-notification-secrets
+	@echo "Deploying notification service"
+	@oc process -f $(TEMPLATE_DIR)/f8notification.app.yaml -o yaml | oc apply -f -
+
 .PHONY: deploy-f8ui
 deploy-f8ui:
 	@echo "Deploying fabric8-ui service"
@@ -70,6 +75,10 @@ deploy-auth-secrets:
 deploy-wit-secrets:
 	oc process -f $(TEMPLATE_DIR)/wit.config.yaml -o yaml | oc apply -f -
 
+.PHONY: deploy-notification-secrets
+deploy-notification-secrets:
+	oc process -f $(TEMPLATE_DIR)/f8notification.config.yaml -o yaml | oc apply -f -
+
 .PHONY: clean-all
 clean-all:
 	oc delete all -l env=dev
@@ -86,6 +95,10 @@ clean-auth:
 .PHONY: clean-wit
 clean-wit:
 	oc delete all -l belongsTo=wit
+
+.PHONY: clean-notification
+clean-notification:
+	oc delete all -l belongsTo=f8notification
 
 .PHONY: clean-f8ui
 clean-f8ui:
--- a/.make/Makefile.dev
+++ b/.make/Makefile.dev
@@ -77,7 +77,10 @@ deploy-wit-secrets:
 
 .PHONY: deploy-notification-secrets
 deploy-notification-secrets:
-	oc process -f $(TEMPLATE_DIR)/f8notification.config.yaml -o yaml | oc apply -f -
+	@echo "Hint: set F8_MANDRILL_API_KEY env variable with your base64 encoded mandrill api key"
+	@oc process -f $(TEMPLATE_DIR)/f8notification.config.yaml \
+		-p F8_MANDRILL_API_KEY=$(F8_MANDRILL_API_KEY) \
+		-o yaml | oc apply -f -
 
 .PHONY: clean-all
 clean-all:
--- a/minishift/dev_setup.md
+++ b/minishift/dev_setup.md
@@ -121,7 +121,7 @@ make deploy-notification
 Look for running pods using `oc get pods`. You should be able to see two pods(f8notification-*). First time it will take some time as it has download required container images.
 
 ##### Mandrill API key
-If you have mandrill api key. Make sure you updated secrets data by key `madrill.api` from `f8notification.config.yaml` with base64 encoded value of your key.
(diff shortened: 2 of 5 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @dipak-pawar for this contribution!



For more information please head over to official [documentation](http://arquillian.org/ike-prow-plugins/#_test_keeper_plugin). You can find there how to configure the plugin.

.make/Makefile.dev
```diff
@@ -70,6 +75,10 @@ deploy-auth-secrets:
 deploy-wit-secrets:
 	oc process -f $(TEMPLATE_DIR)/wit.config.yaml -o yaml | oc apply -f -
 
+.PHONY: deploy-notification-secrets
```
xcoulon: you could probably have the single command below in the `deploy-notification` goal, to keep the Makefile more straightforward

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-588-1`

minishift/dev_setup.md
```diff
@@ -110,6 +110,23 @@ You can try logging by hitting this in browser `http://f8ui-fabric8-services.${m
...
+
+To deploy Notification service, we have following make target which will deploy service.
+```bash
+make deploy-notification
+```
+
+Look for running pods using `oc get pods`. You should be able to see two pods(f8notification-*). First time it will take some time as it has download required container images.
+
+##### Mandrill API key
+If you have mandrill api key. Make sure you updated secrets data by key `madrill.api` from `f8notification.config.yaml` with base64 encoded value of your key.
```
xcoulon: is there a risk that the `f8notification.config.yaml` file was committed and pushed with someone's Mandrill API key? If so, could we avoid that and maybe rely on env variable instead of a config file committed in the repo? or, since `minishift/openshift/f8notification.config.yaml` is an OpenShift template, we could use to generate the `secrets` manifest by including this environment variable

minishift/openshift/f8notification.app.yaml
```diff
@@ -0,0 +1,121 @@
+apiVersion: v1
+kind: Template
```
xcoulon: here you used a template, but what's the benefit compared to a `DeploymentConfig`?

minishift/openshift/f8notification.config.yaml
```diff
@@ -0,0 +1,22 @@
...
+objects:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: f8notification
+  type: Opaque
+  data:
+    madrill.api: eHh4eHh4
+    service.account.id: NGM4M2NhMmQtNmRjYy00MWM5LWFjN2QtYjA2OGFkNGQxN2M1
+    service.account.secret: bm90aWZpY2F0aW9uc2VjcmV0
```
xcoulon: this could be a parameter, so we never commit an API key in the repository

dipak-pawar: I copied this template from notification repository. If someone has to try it then there should have some way using which we can share required keys with a user who wanted to try this locally

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-588-2`

minishift/openshift/f8notification.config.yaml
xcoulon: yes, but what about the API key that is committed in this file ? Can *anyone* use it, really?

minishift/openshift/f8notification.app.yaml
dipak-pawar: If you see you can have few parameters here like `IMAGE_NAME`, `TAG`, `MEMORY`. This way it's easy for a user to just overwrite this parameter using `-p` flag.

minishift/dev_setup.md
dipak-pawar: yes, there is a risk. Let me findgenerate it using env.

.make/Makefile.dev
dipak-pawar: I believe we already have command `deploy-notification` in the makefile, also from the name it's clear what it's doing.

minishift/openshift/f8notification.config.yaml
dipak-pawar: If you are referring to `madrill.api`, I think it's not a valid key. I already described it in the documentation https://github.com/fabric8-services/fabric8-auth/pull/588/files#diff-cad2fe3b8c90b796038ba2d828cb8e4cR123

minishift/openshift/f8notification.app.yaml
xcoulon: ok

minishift/openshift/f8notification.config.yaml
xcoulon: ok, so that part of the docs should be rewritten to explain how to set it as an env var, instead.

.make/Makefile.dev
xcoulon: sure, but my point is: is the `deploy-notification-secret` something that we want to use independently of the `deploy-notification` goal?

minishift/openshift/f8notification.config.yaml
dipak-pawar: Addressed in d291bd56af2c0b37038c22386ec2d2b5a8bdd352

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-588-3`

.make/Makefile.dev
dipak-pawar: If anyone wants to just deploy secrets he can deploy it using `deploy-notification-secret` goal. Sometimes it's useful. But we are not forcing a user to run two independent goals.

xcoulon: sure, I just don't see why someone would want to deploy secrets without the service, but fair-enough, we can keep it like that.

PR #599, opened 2018-08-14
Add API to to check if a user has a specific scope (#492)

### LINKED ISSUE #492: Add API to to check if a user has a specific scope
in a specific space/resource.

xcoulon:
Added endpoint as /api/resources/:resourceID/scopes/:scopeName

Also:
- refactored `controller/resource_roles_blackbox_test.go` by
  grouping in subtests

Fixes #492

### DIFF
--- a/authorization/role/service/role_management_service.go
+++ b/authorization/role/service/role_management_service.go
@@ -46,11 +46,6 @@ func (s *roleManagementServiceImpl) ListByResource(ctx context.Context, currentI
 	return s.Repositories().IdentityRoleRepository().FindIdentityRolesByResource(ctx, resourceID, false)
 }
 
-// ListByResource lists all identity roles for the resource if the current user has permissions to view the roles
-func (s *roleManagementServiceImpl) ListByIdentityAndResource(ctx context.Context, currentIdentity uuid.UUID, resourceID string) ([]role.RoleDescriptor, error) {
-	return s.ListByIdentityAndResource(ctx, currentIdentity, resourceID)
-}
-
 func (s *roleManagementServiceImpl) requireViewRolesScope(ctx context.Context, currentIdentity uuid.UUID, resourceID string) error {
 	// Lookup the resourceID and ensure the resource is valid
 	rt, err := s.Repositories().ResourceRepository().Load(ctx, resourceID)
--- a/authorization/role/service/role_management_service.go
+++ b/authorization/role/service/role_management_service.go
@@ -46,6 +46,11 @@ func (s *roleManagementServiceImpl) ListByResource(ctx context.Context, currentI
 	return s.Repositories().IdentityRoleRepository().FindIdentityRolesByResource(ctx, resourceID, false)
 }
 
+// ListByResource lists all identity roles for the resource if the current user has permissions to view the roles
+func (s *roleManagementServiceImpl) ListByIdentityAndResource(ctx context.Context, currentIdentity uuid.UUID, resourceID string) ([]role.RoleDescriptor, error) {
+	return s.ListByIdentityAndResource(ctx, currentIdentity, resourceID)
+}
+
 func (s *roleManagementServiceImpl) requireViewRolesScope(ctx context.Context, currentIdentity uuid.UUID, resourceID string) error {
 	// Lookup the resourceID and ensure the resource is valid
 	rt, err := s.Repositories().ResourceRepository().Load(ctx, resourceID)
--- a/design/role.go
+++ b/design/role.go
@@ -87,6 +87,21 @@ var _ = a.Resource("resource_roles", func() {
 		a.Response(d.Unauthorized, JSONAPIErrors)
 		a.Response(d.Conflict, JSONAPIErrors)
 	})
+	a.Action("hasScope", func() {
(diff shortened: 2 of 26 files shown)


fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-599-1`

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-599-2`

authorization/role/service/role_management_service.go
```diff
@@ -46,6 +46,11 @@ func (s *roleManagementServiceImpl) ListByResource(ctx context.Context, currentI
 	return s.Repositories().IdentityRoleRepository().FindIdentityRolesByResource(ctx, resourceID, false)
 }
 
+// ListByResource lists all identity roles for the resource if the current user has permissions to view the roles
+func (s *roleManagementServiceImpl) ListByIdentityAndResource(ctx context.Context, currentIdentity uuid.UUID, resourceID string) ([]role.RoleDescriptor, error) {
+	return s.ListByIdentityAndResource(ctx, currentIdentity, resourceID)
```
alexeykazakov: This method is
- broken (looks like an infinite recursive loop?)
- not covered by tests
- not used anywhere :)

Wonder if it was some temporal experiments you forgot to clean up.

alexeykazakov:
@sbose78 @xcoulon @sbryzak I have concerns regarding exposing HasScope() method to a public endpoint.
Don't want clients to relay on that endpoint instead of using RPT tokens because this method doesn't support any cache/optimization. It was designed to be used internally. So Auth won't be able to control usage of that heavy and slow method. But I also understand that implementing the PRT flow (so it's easy to be used by clients) will take awhile.

xcoulon:
@alexeykazakov good point for the lack of cache/optimization here. Maybe there's something to add in the `PermissionService` layer to benefit from caching, too?

alexeykazakov:
@sbryzak is working on a cache for RPT tokens. Wondering if we could reuse it here as well if you guys believe we really do need to expose that endpoint :)

sbose78:
Exactly. That was on my mind as well :)

Here's how we could go about it -

1. Expose the API as here. ( Easy to do, and makes our authz service usable, but comes with a cost)

2. Finish the RPT workflow ( Complex to do, but optimizes our service - and with any such optimization things might take a while to mature )

3. Utilize the optimization in the exposed Rest API as well.

That way our authorization service will give the resource servers the option to either use the RPT token workflow or the API endpoint.

alexeykazakov:
@sbose78 good plan. +1

xcoulon:
@sbose78 ok, let's do that way. I was also thinking about the RPT workflow and the DB caching, but I'm not sure if we'll be able to re-use it as-is or if it'll need some refactoring. 

Also, in its current form, the PR provides a new endpoint to check if the user has a specific scope in a requested resource, i.e., something like `/api/resources/:resourceID/scopes/:scopeName` and it returns a JSON-API response with a boolean attribute (`hasScope`). This is the way we can directly expose the `HasScope` service method. If we want to list all scopes for the current user in the request resource as was suggested in https://github.com/fabric8-services/fabric8-auth/issues/492#issuecomment-389024147, then we'll need another SQL request which @sbryzak is already working on as part of the RPT workflow.

sbose78:
@xcoulon I think this API is nice. Just checking if a user has the needed scope is enough. Let's start with this.

xcoulon:
thanks for your feedback, @sbose78. Could you approve the PR then, please ?

authorization/role/service/role_management_service.go
sbose78: @xcoulon , have you taken care of this?

xcoulon: oh.... my bad! Let me take care of it now!

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-599-3`

xcoulon:
[test]

authorization/role/service/role_management_service.go
xcoulon: @sbose78 it's now been taken care of. See https://github.com/fabric8-services/fabric8-auth/pull/599/commits/3f93465f00366a503a8542d1fe357310213be32c

xcoulon:
[test]

xcoulon:
[test]

xcoulon:
thanks, @sbose78 !

design/role.go
```diff
@@ -87,6 +87,21 @@ var _ = a.Resource("resource_roles", func() {
 		a.Response(d.Unauthorized, JSONAPIErrors)
 		a.Response(d.Conflict, JSONAPIErrors)
 	})
+	a.Action("hasScope", func() {
```
sbose78: You've missed the jwt protection here, @xcoulon .

xcoulon: ouch! Let me fix that immediately!

PR #626, opened 2018-08-28
chore: add config file to skip validation for minishift folder

dipak-pawar:
In Ike-Prow we have a feature where we can add our customized location to skip test check. http://arquillian.org/ike-prow-plugins/#test-keeper-config

Currently, ike is crying each time if we change something in `minishift/` folder. so reviewer has to comment `ok-without-test`.

After this ike won't check for tests with change in all paths mentioned in config file.(in our case `minishift` so no more `ok-without-tests` if any change in `minishift/`).

### DIFF
--- a/.ike-prow/test-keeper.yaml
+++ b/.ike-prow/test-keeper.yaml
@@ -1,0 +1,2 @@
+skip_validation_for:
+  - 'minishift/'


fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-626-1`

PR #628, opened 2018-08-28
Use existing type as query param for listing resources a user has a role in

### LINKED ISSUE #623: Use existing type as query param for listing resources a user has a role in
Instead of querying spaces with something like `/api/user/resources?type=spaces`, we should use the existing `openshift.io/resource/space`, which avoids the unnecessary conversion.

xcoulon:
Replace `spaces` with `openshift.io/resource/space` as
the resource type for spaces so there's no need for
a transation from `spaces`, but keep checking the `type`
query param for invalid values, to avoid
running an expensive query for nothing (and then return
a 400 error)

Also, `openshift.io/resource/space` is a valid query param
 as defined in https://tools.ietf.org/html/rfc3986#section-3.4,
even if it contains `/` characters

Fixes #623

### DIFF
--- a/controller/user.go
+++ b/controller/user.go
@@ -83,11 +83,13 @@ func (c *UserController) ListResources(ctx *app.ListResourcesUserContext) error 
 		}, "Bad Token")
 		return jsonapi.JSONErrorResponse(ctx, errors.NewUnauthorizedError("bad or missing token"))
 	}
-	var resourceType string
+	resourceType := ctx.Type
 	switch ctx.Type {
-	case "spaces":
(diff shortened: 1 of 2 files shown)


fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-628-1`

sbose78:
>Also, openshift.io/resource/space is a valid query param
as defined in https://tools.ietf.org/html/rfc3986#section-3.4,
even if it contains / characters


:+1:

PR #632, opened 2018-08-29
Remove role names and scopes from API

### LINKED ISSUE #629: Remove role names from API
- As a follow-up from https://github.com/fabric8-services/fabric8-auth/pull/586
We should remove the role names ( and the scope names as well since its not needed ) from the API response - and only stick to returning the resource IDs. Since the above API is not for making permission decisions. 

- On the WIT side, the response should be similar to the response in /api/namespaces, ie - only have the space metadata returned in the response, without specifying the role/scopes associated with it.

This is to address the concern that clients/resource-servers might inadvertently be using the role names to make decisions, instead of the scope names.

xcoulon:
Roles and scopes should not be provided in the response
to the endpoint that lists all resources of a given type
for which the current user has a role, as this may lead
to bad usage of the API. Other endpoints already exists
to check the permissions on a given resource.

Response now looks like this:
````
{
    "data": [{
        "id": "114c89ec-8ff2-4c65-adb4-161c9b505be9",
        "links": {
            "related": "http:///api/resource/114c89ec-8ff2-4c65-adb4-161c9b505be9"
        },
        "type": "resources"
    }, {
        "id": "5d7d3f9b-5a53-4d7e-b321-de3999542626",
        "links": {
            "related": "http:///api/resource/5d7d3f9b-5a53-4d7e-b321-de3999542626"
        },
        "type": "resources"
    }]
}
````
Also: needed to rename the `Read` action to `Show` so that Goa
can generate the `Href` utility function to provide a link to the
resource in the JSON-API response.

Fixes #629

### CODE REVIEW

**Comment by fabric8cd** [comment]
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-632-1`

**Comment by sbose78** [comment]
``
"type": "openshift.io/resource/space"
```
Is this the JSON-API type? Then, we are listing resources, so the type=resource.

xcoulon:
@sbose78 I replaced the `"openshift.io/resource/space"` with `"resources"` and also included a "related" link to each resource in the response. The PR description has been updated with the latest response format.

fabric8cd:
@xcoulon snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-632-2`

xcoulon:
[test]

xcoulon:
thanks for the review, @sbose78 !

PR #635, opened 2018-08-30
Add scopes endpoint for resources

### LINKED ISSUE #634: Add endpoint for reading the current user's scopes for a resource
We need an endpoint that will return all of the scopes available for a specified resource, for the current user.  This will be a simpler alternative than using RPT tokens for certain resource servers.

The proposed api endpoint for this is:

Request:

`GET /api/resource/<resource_id>/scopes`

Response:

```
{
  "data": {
    "id": "<resource_id>",
    "type": "resource",
    "scopes": ["manage","viewer"]
  }
}

```
Original issue:
[https://openshift.io/openshiftio/Openshift_io/plan/detail/528](https://openshift.io/openshiftio/Openshift_io/plan/detail/528)

sbryzak:
Fixes #634 

This PR adds support for reading the current user's scopes for a resource.

Request:

`GET /api/resource/<resource_id>/scopes`

Response:

```
{
  "data": [
    {
      "id": "alpha",
      "type": "user_resource_scope"
    },
    {
      "id": "bravo",
      "type": "user_resource_scope"
    }
  ]
}

```

### DIFF
--- a/controller/resource.go
+++ b/controller/resource.go
@@ -103,12 +103,17 @@ func (c *ResourceController) Scopes(ctx *app.ScopesResourceContext) error {
 		return jsonapi.JSONErrorResponse(ctx, err)
 	}
 
+	scopes := []*app.ResourceScopes{}
+
+	for _, scope := range pc.ScopesAsArray() {
+		scopes = append(scopes, &app.ResourceScopes{
+			ID:   scope,
+			Type: "user_resource_scope",
+		})
+	}
+
 	res := &app.ResourceScopesData{
-		Data: &app.ResourceScopes{
-			ID:     ctx.ResourceID,
-			Type:   "resource",
-			Scopes: pc.ScopesAsArray(),
-		},
+		Data: scopes,
 	}
 
 	return ctx.OK(res)
--- a/controller/resource_blackbox_test.go
+++ b/controller/resource_blackbox_test.go
@@ -221,11 +221,19 @@ func (rest *TestResourceREST) TestScopesOK() {
 
 	// Invoke the endpoint
 	_, scopes := test.ScopesResourceOK(rest.T(), svc.Context, svc, ctrl, res.ResourceID())
-	require.Equal(rest.T(), scopes.Data.ID, res.ResourceID())
-	require.Equal(rest.T(), scopes.Data.Type, "resource")
-	require.Len(rest.T(), scopes.Data.Scopes, 2)
-	require.Contains(rest.T(), scopes.Data.Scopes, "foo")
-	require.Contains(rest.T(), scopes.Data.Scopes, "bar")
+	require.Len(rest.T(), scopes.Data, 2)
+	fooFound := false
+	barFound := false
+	for _, scope := range scopes.Data {
+		require.Equal(rest.T(), "user_resource_scope", scope.Type)
+		if scope.ID == "foo" {
+			fooFound = true
+		} else if scope.ID == "bar" {
+			barFound = true
+		}
+	}
+	require.True(rest.T(), fooFound)
+	require.True(rest.T(), barFound)
 
 	// Create another user
 	user2 := rest.Graph.CreateUser()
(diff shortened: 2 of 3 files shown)


controller/resource.go
```diff
@@ -81,3 +82,34 @@ func (c *ResourceController) Register(ctx *app.RegisterResourceContext) error {
...
+	// Load the scopes
+	pc, err := c.app.PrivilegeCacheService().CachedPrivileges(ctx, currentIdentity.ID, ctx.ResourceID)
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+
+	res := &app.ResourceScopesData{
+		Data: &app.ResourceScopes{
+			ID:     ctx.ResourceID,
+			Type:   "resource",
```
sbose78: Should be of type "scopes", what do you think @xcoulon ?

xcoulon: right, it makes sense to use `scopes` here, indeed.

sbryzak:
@xcoulon I do have test coverage for multiple roles in the privilege cache tests, however I have no problem adding the same here.

controller/resource.go
sbryzak: @sbose78 @xcoulon This doesn't make sense according to the [JSON API](http://jsonapi.org/format/#document-resource-object-identification) specification:

> Within a given API, each resource object’s type and id pair MUST identify a single, unique resource. 

We are not identifying a scope here, we are identifying a resource.  Technically, we could make the case that the scopes are an *attribute* of the resource, however I think this is a special case as this endpoint produces a dynamic result depending on who is doing the calling, which is why I don't think that using attributes fits here.

sbryzak:
@xcoulon test updated with multiple roles.

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-635-2`

controller/resource.go
xcoulon: ah, I missed that in the design: the endpoint response is something like 
````
{
    "data": {
        "id": "74372086-9b16-4cef-a100-70e3e2a32194",
        "scopes": ["foo", "bar"],
        "type": "resource"
    }
}
````
but I'm not sure that having a `scopes` attribute as a child of `data` is valid either, as explained in here:http://jsonapi.org/format/#document-resource-objects. My understanding is that the `id` and `type` elements are mandatory, and `attributes`, `relationships`, `links` and `meta` are optional, but no other element is allowed. So I would suggest moving the `scopes` under the `attributes` element. As for the `type` value to use, maybe `resource_scopes` would be more accurate? My reasoning is that the `ID` refers to the resource, response does not contain the resource's attributes, but rather an array of scopes (based on the user's roles, indeed).

Comment (sbryzak) [reply controller/resource.go]: My reluctance to put `scopes` under the `attributes` element is because it is not a static attribute of the resource, rather a *calculated* set of values dependent on the current user, and as such not really an attribute of the resource itself.  

As for the `type` value, `resource_scopes` is still not compliant with the JSON API spec - the identifier is for a resource, and the spec does not allow us to stray from this.  I.e. the "MUST" in:

> **MUST** identify a single, unique resource.

is final and doesn't allow for any other alternative.  Just because we don't include other *usual* attributes of the resource in the request doesn't mean it's still not resource-centric.

Back on where to place the `scopes` value - if we were worried about strict compliance here, then I suggest we place the `scopes` under the [meta](http://jsonapi.org/format/#document-meta) tag instead.

Comment (sbose78) [reply controller/resource.go]: The API `GET /api/resource/<resource_id>/scopes` is returning scopes for a specific resource - which is why I feel `scopes` could be more appropriate or as @xcoulon mention resource_scopes could be a good option.

File: controller/resource.go [hunk 2]
```diff
@@ -81,3 +82,39 @@ func (c *ResourceController) Register(ctx *app.RegisterResourceContext) error {
...
+	if err != nil {
+		return jsonapi.JSONErrorResponse(ctx, err)
+	}
+
+	scopes := []*app.ResourceScopes{}
+
+	for _, scope := range pc.ScopesAsArray() {
+		scopes = append(scopes, &app.ResourceScopes{
+			ID:   scope,
+			Type: "user_resource_scope",
```
sbose78: Looks good!

fabric8cd:
@sbryzak snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-635-3`

PR #659, opened 2018-09-25
Proxy /api/clusters to Cluster Management Service

alexeykazakov:
- [x] Proxy `auth.openshift.io/api/clusters` to `cluster.openshift.io/api/clusters`
- [x] Call `cluster.openshift.io/api/clusters/auth` to get additional cluster info (oauth client id/secret, etc) when resolving OpenShift provider (during account linking or new user registration)
- [x] Tests

Some usecases is quite tricky to cover by automated tests by 100%. We will need to test it carefully in prod-preview:
- Registration flow
- OSO-account linking


application/service/factory/service_factory.go
```diff
@@ -128,6 +130,7 @@ type ServiceFactory struct {
 	config                  *configuration.ConfigurationData
 	witServiceFunc          func() service.WITService          // the function to call when `WITService()` is called on this factory
 	notificationServiceFunc func() service.NotificationService // the function to call when `NotificationService()` is called on this factory
+	clusterServiceFunc      func() service.ClusterService      // the function to call when `ClusterService()` is called on this factory
```
golangcibot: `clusterServiceFunc` is unused

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-1`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-2`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-3`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-4`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-5`

alexeykazakov:
cc: @nurali-techie @MatousJobanek

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-6`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-7`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-8`

token/token.go
```diff
@@ -51,8 +51,20 @@ const (
 	GeminiServer       = "fabric8-gemini-server"
 )
 
+var defaultManager Manager
+
+// DefaultManager creates the default manager if it has not created yet.
+// This function must be called in main to make sure the default manager is created during service startup.
+func DefaultManager(config Configuration) (Manager, error) {
+	var err error
+	if defaultManager == nil {
```
sbryzak: Do you think it's worth putting some synchronization around this?

alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/659/commits/8668cfa5b59bbce00e19baf6a0f85a02968e4294

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-9`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-10`

application/service/factory/service_factory.go
```diff
@@ -159,6 +170,10 @@ func NewServiceFactory(producer ServiceContextProducer, config *configuration.Co
 	f.notificationServiceFunc = func() service.NotificationService {
 		return notificationservice.NewNotificationService(f.getContext(), f.config)
 	}
+	// default function to return an instance of Cluster Service
+	f.clusterServiceFunc = func() service.ClusterService {
+		return clusterservice.NewClusterService(f.getContext(), config)
```
nurali-techie: `f.config` would be better than `config`

alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/659/commits/7db39812d4b5e9743dff65a22c87ce9660cf8946

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-11`

cluster/service/cluster.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+			return err
+		}
+	}
+	return nil
+}
+
+// Clusters converts the given cluster map to an array slice
+func Clusters(clusters map[string]*cluster.Cluster) []cluster.Cluster {
+	var cs []cluster.Cluster
```
nurali-techie: allocating array (with target length) before loop would be better.

main.go
```diff
@@ -183,8 +172,17 @@ func main() {
 
 	keycloakProfileService := login.NewKeycloakUserProfileClient()
 
+	// Try to fetch the initial list of clusters and start Cluster Service cache refresher
+	err = clusterservice.Start(context.Background(), config)
+	if err != nil {
+		// It's not a critical error. Cluster management service can be offline. We will try to fetch clusters later when we need them.
+		log.Warn(nil, map[string]interface{}{
+			"err": err,
+		}, "failed to fetch clusters")
```
nurali-techie: just want to know, why it's not critical?

cluster/service/cluster.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, cls := range clusters {
+		cs = append(cs, *cls)
+	}
+	return cs
+}
+
+func ClusterByURL(clusters map[string]*cluster.Cluster, url string) *cluster.Cluster {
+	for apiURL, cluster := range clusters {
+		if strings.HasPrefix(rest.AddTrailingSlashToURL(url), apiURL) {
+			return cluster
```
nurali-techie: I guess, by mistake, you returning ref to cluster cache object.

cluster/service/cluster_whitebox_test.go
```diff
@@ -0,0 +1,165 @@
...
+func (s *TestClusterSuite) TearDownTest() {
+	if clusterCache != nil {
+		clusterCache.stop()
+	}
+}
+
+func (s *TestClusterSuite) TestClustersFail() {
+	ctx, _, reqID := tokentestsupport.ContextWithTokenAndRequestID(s.T())
+
+	s.T().Run("clusters() fails if can't get clusters", func(t *testing.T) {
```
xcoulon: OH WAIT! Did you use subtests here ?!? 😂

controller/clusters_blackbox_test.go
```diff
@@ -23,45 +19,13 @@ func TestRunClustersREST(t *testing.T) {
 	suite.Run(t, &TestClustersREST{UnitTestSuite: testsuite.NewUnitTestSuite()})
 }
 
-func (rest *TestClustersREST) SecuredControllerWithServiceAccount(serviceAccount account.Identity) (*goa.Service, *ClustersController) {
-	svc := testsupport.ServiceAsServiceAccountUser("Token-Service", serviceAccount)
+func (rest *TestClustersREST) UnsecuredController() (*goa.Service, *ClustersController) {
```
xcoulon: why are you changing the controller to `unsecured` if you need a SA token in https://github.com/fabric8-services/fabric8-auth/pull/659/files#diff-28264df4d2b218e146d7d2540ac6c26dR92 ? (unless I missed something?)

main.go
xcoulon: is there any risk of error if we load a user info but can't associate the user's account with a cluster in the cache?

nurali-techie:
@alexeykazakov one design query,
I see that we have cluster_service_client sort of stuff to which auth_service interact to serve '/api/clusters' API.

Instead, why can't we have auth_service just forwarding call for `/api/clusters` API to cluster_service and returned whatever response comes from clsuter_service to caller.  Something like, auth_service serves as http_proxy for `/api/clusters` API calls.

main.go
alexeykazakov: There is one usecase when it's ok when we fail here: Cluster management service is not started yet because it's wait for Auth service to start to fetch public keys.
If we required Cluster service to be ready to start Auth then we would create cycle dependency when both Auth and Cluster wait for each other to start up.

If cluster loading fails in main then we will try to load it again later when we actually need it:
- when user tries to link OSO-OSIO accounts
- when user tries to sign-up

If Cluster management still down then users will see an error message.

cluster/service/cluster.go
alexeykazakov: @nurali-techie This is was the intent. ClusterByURL(clusters map[string]) doesn't care - it returns a pointer to the found cluster in the giving cluster map. But yes, https://github.com/fabric8-services/fabric8-auth/pull/659/files#diff-9a78061a303f1c442fb407b8ce4b9906R55 will pass the cached map so eventually it will return the cached object.
What is your concern here? You think it's safer to return a copy of the cluster instead of a pointer to the original cached object? We could use ClusterByURL() (cluster, ok bool) instead...

cluster/service/cluster_whitebox_test.go
alexeykazakov: I was thinking about you when adding them ;)
Just to clarify my position. It's OK to use subtests when you don't need to sacrifice automated test setup/cleanup to achieve proper test isolation.

In this case I actually needed subtests to depend on each other. I need the second subtest to be executed after the first one.
The alternative of using subtests would be to have one big test function TestClustersFail() and merge both subtests there. It's the case when I do not want subtests to be isolated. Subtests here is just to represent block of codes/tests more nicely.

cluster/service/cluster.go
nurali-techie: Yes, it's better to return copy rather pointer to cache object.

main.go
nurali-techie: thanks for clarification and this should be captured somewhere else if possible.  may be in code itself ??

cluster/service/cluster.go
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/659/commits/6b66114a88021f393a170ab200e7653e652268db

cluster/service/cluster.go
alexeykazakov: Good point. Done in https://github.com/fabric8-services/fabric8-auth/pull/659/commits/6b66114a88021f393a170ab200e7653e652268db

main.go
alexeykazakov: Add more detailed comment in https://github.com/fabric8-services/fabric8-auth/pull/659/commits/ff60248226b5ca67e651abf7ce65d43a3994efc3

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-13`

fabric8cd:
@alexeykazakov snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-659-14`

xcoulon:
> > but there's already some logic to fetch and cache (with scheduled refresh) the list of clusters in fabric8-tenant, so I wonder if this part of your PR should be in fabric8-common instead, so it can be reused by `auth` and` tenant` directly.
> 
> @xcoulon The cluster cache used in this PR is very specific to Auth. It's using a special endpoint and cluster info which should be available to Auth service only. If we want to move any cluster cache mechanism to common for re-using then I would consider moving it from tenant service instead. We potentially could need to re-use it in other services.

Ok, I missed that. So, is this `cluster` endpoint that is specific to `auth` restricted via a specific kind of token for example?

alexeykazakov:
> So, is this cluster endpoint that is specific to auth restricted via a specific kind of token for example?

Yes - https://github.com/fabric8-services/fabric8-cluster/blob/master/controller/clusters.go#L62

Issue #674, opened 2018-10-01
fix user creation during developer mode /api/token/generate


sbose78:
Instead of  
```
	idpService := login.NewIdentityProvider(c.Configuration)
	_, token, err := c.Auth.CreateOrUpdateIdentityAndUser(ctx, ctx.RequestData.URL, generatedToken, ctx.RequestData, idpService, c.Configuration)
	if err != nil {
		return jsonapi.JSONErrorResponse(ctx, err)
	}
```
in https://github.com/fabric8-services/fabric8-auth/blob/master/controller/token.go#L138 which tries to reach out to keycloak ( which would no longer have the user ), call the gorm repository methods to create/update the user in the db.

( By the way, this is another API we need to consider removing. )

PR #676, opened 2018-10-01
fix(#674): create token for dev user in developer mode

dipak-pawar:
* Creating user and identity in DB for the first time in developer mode for dev user

Fixes: #674

### DIFF
--- a/controller/token.go
+++ b/controller/token.go
@@ -30,6 +30,11 @@ import (
 	"golang.org/x/oauth2"
 )
 
+const (
+	DevUsername = "developer"
+	DevEmail    = "osio-developer@email.com"
+)
+
 // TokenController implements the login resource.
 type TokenController struct {
 	*goa.Controller
@@ -101,11 +106,10 @@ func (c *TokenController) Generate(ctx *app.GenerateTokenContext) error {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, errs.New("postgres developer mode is not enabled")))
 	}
 
-	devUsername := "developer"
 	var identities []account.Identity
 	err := transaction.Transactional(c.app, func(tr transaction.TransactionalResources) error {
 		var err error
-		identities, err = tr.Identities().Query(account.IdentityWithUser(), account.IdentityFilterByUsername(devUsername), account.IdentityFilterByProviderType(account.KeycloakIDP))
+		identities, err = tr.Identities().Query(account.IdentityWithUser(), account.IdentityFilterByUsername(DevUsername), account.IdentityFilterByProviderType(account.KeycloakIDP))
 		return err
 	})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-676-1`

controller/token.go
```diff
@@ -126,6 +126,26 @@ func (c *TokenController) Generate(ctx *app.GenerateTokenContext) error {
...
+				return err
+			}
+			return nil
+		})
+
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"err": err,
+			}, "failed to create a user and identity for dev user")
+			return jsonapi.JSONErrorResponse(ctx, err)
```
xcoulon: you should return an `InternalServerError` with a message like `"failed to create a user and identity for dev user"`, otherwise, it's the repository error that will be returned to the user, which is not what we may want.

controller/token.go
```diff
@@ -126,6 +126,26 @@ func (c *TokenController) Generate(ctx *app.GenerateTokenContext) error {
 			ProviderType:          account.KeycloakIDP,
 			RegistrationCompleted: true,
 		}
+
+		err = transaction.Transactional(c.app, func(tr transaction.TransactionalResources) error {
```
xcoulon: how does this apply to `dev user in developer mode` only (as the title of the PR suggests)?

controller/token.go
```diff
@@ -126,6 +126,26 @@ func (c *TokenController) Generate(ctx *app.GenerateTokenContext) error {
...
+			// Using the old-fashioned service
+			err := tr.Identities().Save(ctx, &devIdentity)
```
xcoulon: You could add tests to verify that the user and identity records are created, and that they are correctly linked (eg: `Identity.UserID` == generated `User.ID`)

controller/token.go
dipak-pawar: This is the code which makes sure that you have developer mode enabled where we are looking only for `developer` user https://github.com/fabric8-services/fabric8-auth/blob/c5d5e91e3729356ac779f01184c0353b9ddd8ef6/controller/token.go#L99-L113

controller/token.go
dipak-pawar: sure let me try to add that

controller/token.go
xcoulon: ah of course, I missed that :/ Thanks for the clarification ;)

fabric8cd:
@dipak-pawar snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-676-2`

controller/token.go
```diff
@@ -30,6 +30,11 @@ import (
 	"golang.org/x/oauth2"
 )
 
+const (
+	DevUsername = "developer"
```
sbose78: Could we make these internal to the package?

xcoulon: actually, it's also used in the test, which is not in the same pkg

dipak-pawar: no we can't as we used it in test like @xcoulon  mentioned

alexeykazakov: You could convert your test to white_box then.

alexeykazakov: Oh.. it's actually just a string. Then you really should have just use the same string in your tests instead of exporting internal constants. But it's not that critical.

xcoulon: yes, that could work, too.

PR #697, opened 2018-10-16
Dont carry over params as fragments

sbose78:
fixes https://github.com/fabric8-services/fabric8-auth/issues/698

### DIFF
--- a/login/service.go
+++ b/login/service.go
@@ -356,30 +356,32 @@ func (keycloak *KeycloakOAuthProvider) AuthCodeCallback(ctx *app.CallbackAuthori
 	if err != nil {
 		return nil, err
 	}
-	var redirectTo string
+
+	redirectTo := buildRedirectURL(ctx.Code, ctx.State, referrerURL, responseMode)
+	return &redirectTo, nil
+}
+
+func buildRedirectURL(code string, state string, referrerURL *url.URL, responseMode *string) string {
 
 	var parameters url.Values
 
 	if responseMode != nil && *responseMode == "fragment" {
-
 		parameters = url.Values{}
-		parameters.Add("code", ctx.Code)
-		parameters.Add("state", ctx.State)
+		parameters.Add("code", code)
+		parameters.Add("state", state)
 		if referrerURL.Fragment != "" {
 			referrerURL.Fragment = referrerURL.Fragment + "&" + parameters.Encode()
 		} else {
 			referrerURL.Fragment = parameters.Encode()
 		}
-
 	} else {
 		parameters = referrerURL.Query()
-		parameters.Add("code", ctx.Code)
-		parameters.Add("state", ctx.State)
+		parameters.Add("code", code)
+		parameters.Add("state", state)
 		referrerURL.RawQuery = parameters.Encode()
 	}
-	redirectTo = referrerURL.String()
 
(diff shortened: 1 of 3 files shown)


fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-697-2`

login/service.go
```diff
@@ -357,18 +357,24 @@ func (keycloak *KeycloakOAuthProvider) AuthCodeCallback(ctx *app.CallbackAuthori
 		return nil, err
 	}
 	var redirectTo string
-	parameters := referrerURL.Query()
-	parameters.Add("code", ctx.Code)
-	parameters.Add("state", ctx.State)
+
+	var parameters url.Values
 
 	if responseMode != nil && *responseMode == "fragment" {
```
xcoulon: wouldn't it be simpler to move the logic of checking the response mode to build the referrerURL in a separate function, that could be exhaustively tested in a simpler manner ?

sbose78: addressed in https://github.com/fabric8-services/fabric8-auth/pull/697/commits/d8e19d215298611f8b070df7fd4d96acbe866c52

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-697-3`

alexeykazakov:
@sbose78 @xcoulon since It was me who was so aggressively against the sub-tests in the first place let me try to clarify my thoughts:
1. I'm personally OK to use sub-tests in pure Unit tests where there is no any automated test setups and teardowns needed and where we don't need to isolate subtests at all. In same cases we actually need to run some tests in a sequence (so we want some parts of the test to depend on other parts of the same test). So subtests are fine for such cases. They should not be forced though IMO. It's up to the test author do use them or not.
2. If we need to isolate subtests (there are DB interactions or any other statefull things) then I'm still strongly against using subtests for such cases. Automated test isolation (via test startups, teardowns) is much more important then nice looking blocks of subtests.

sbose78:
Makes sense @alexeykazakov , I'll go with @xcoulon 's suggestion to use sub-tests here.

fabric8cd:
@sbose78 snapshot fabric8-auth image is available for testing.  `docker pull docker.io/fabric8/fabric8-auth:SNAPSHOT-PR-697-4`

xcoulon:
> @sbose78 @xcoulon since It was me who was so aggressively against the sub-tests in the first place let me try to clarify my thoughts:
> 
> 1. I'm personally OK to use sub-tests in pure Unit tests where there is no any automated test setups and teardowns needed and where we don't need to isolate subtests at all. In same cases we actually need to run some tests in a sequence (so we want some parts of the test to depend on other parts of the same test). So subtests are fine for such cases. They should not be forced though IMO. It's up to the test author do use them or not.
> 2. If we need to isolate subtests (there are DB interactions or any other statefull things) then I'm still strongly against using subtests for such cases. Automated test isolation (via test startups, teardowns) is much more important than nice looking blocks of subtests.

I also agree, @alexeykazakov. I thought subtests would apply well here, because we don't have any DB setup to take care of, and the initial test was covering multiple uses cases.

PR #701, opened 2018-10-22
Return deprovisioned attr in User API response

sbose78:
This will help services verify if the user has been de-provisioned though this might mean auth service would now end up servicing a lot more HTTP GET /user requests.

### DIFF
--- a/controller/users.go
+++ b/controller/users.go
@@ -852,6 +852,7 @@ func ConvertToAppUser(request *goa.RequestData, user *accountrepo.User, identity
 	var userURL string
 	var email string
 	var isEmailPrivate bool
+	var isUserDeprovisioned bool
 	var createdAt time.Time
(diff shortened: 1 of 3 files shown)


sbose78:
This was on request from @davidfestal / Che . I'm opening this up for discussion.

davidfestal:
Thanks @sbose78 !

That would allow the RhChe server to verify that a user is deprovisioned, and then refuse access to the server if it is the case.
We could verify that at each login, or cache the result for a given period of time and verify it on a regular basis, in order to avoid too many network accesses during Che REST authentication. But in any case this would allow forbidding relatively quickly the use of tokens that correspond to deprovisioned users.

davidfestal:
I'll open an issue on the RhChe side to use this.

sbose78:
Yup, that sounds great. Closing this.

PR #705, opened 2018-10-30
chore: call generate_client_setup before deployment of master

dipak-pawar:
Depends on #https://github.com/openshiftio/openshiftio-cico-jobs/pull/875

### DIFF
--- a/cico_build_deploy.sh
+++ b/cico_build_deploy.sh
@@ -10,6 +10,8 @@ if [ ! -f .cico-prepare ]; then
 
     run_tests_without_coverage;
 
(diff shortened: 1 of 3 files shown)


cico_build_deploy.sh
```diff
@@ -10,6 +10,8 @@ if [ ! -f .cico-prepare ]; then
 
     run_tests_without_coverage;
 
+    generate_client_setup fabric8-auth auth tool fabric8-services fabric8-auth-client
```
sbose78: So-
service,package,tool dir,org,client repo

dipak-pawar: yes

Issue #708, opened 2018-11-01
Fix broken code coverage


sbryzak:
There seems to be a bug in Go relating to the `coverpkg` parameter when generating a code coverage report - see [this issue](https://github.com/golang/go/issues/25093).  The fix is to upgrade Go to version 1.11.

PR #709, opened 2018-11-01
Upgrade to go v11.1 for test-coverage CI

alexeykazakov:
Fixes https://github.com/fabric8-services/fabric8-auth/issues/708


alexeykazakov:
Coverage build fails because gofmt was changed in v.1.11.
The coverage build uses go v1.11.1 but the main CI job uses go from centos image which is 1.9 (as far as I remember).
I think we have to skip gofmt check for the coverage build and keep it for the main build only.

alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/43638339927115d12380cb65dc17cd38ca584ca2/.ike-prow/test-keeper.yaml).

alexeykazakov:
I'm switched all builds to go v.11.1 from the site in that PR and fixed formatting. Let's see if it's acceptable. If not then we can just skip formatting check when run tests with coverage.

sbose78:
> I think we have to skip gofmt check for the coverage build and keep it for the main build only.
Agreed, coverage build shouldn't really have an business with gofmt checks.

I'm fine with this change - upgrading to go 11.1 sounds good to me. Good find @sbryzak !

PR #689, opened 2018-10-04
Refactor authentication codebase

sbryzak:
This PR refactors the authentication codebase.  

###  Introduction of new services

Many legacy "services" have been refactored as proper service-layer services.  This allows them to take advantage of:

- Consistent API for accessing repository layer and other services.
- Built-in support for atomic transactions via the `ExecuteInTransaction` function, ensuring service integrity
- Reduced setup burden (no longer required to instantiate services in `main.go`)

The new services that are added in this PR are:

- AuthenticationProviderService
- LinkService 
- LinkingProviderFactory
- LogoutService
- OSOSubscriptionService
- UserProfileService

### Removal of Keycloak configuration, normalisation of authentication provider configuration parameters.

Configuration parameter name changes:

`keycloak.client.id`         -> `auth.provider.client.id`
`keycloak.secret`            -> `auth.provider.client.secret`
`keycloak.endpoint.admin`    -> `auth.provider.endpoint.admin`
`keycloak.endpoint.auth`     -> `auth.provider.endpoint.auth`
`keycloak.endpoint.token`    -> `auth.provider.endpoint.token`
`keycloak.endpoint.userinfo` -> `auth.provider.endpoint.userinfo`

Configuration parameters removed:

- `keycloak.testuser.name`
- `keycloak.testuser.secret`
- `keycloak.testuser2.name`
- `keycloak.testuser2.secret`
- `keycloak.domain.prefix`
- `keycloak.realm`
- `keycloak.url`

### Restructured codebase

A new base-level package, `authentication` has been added, with a similar structure to the `authorization` package however dealing with authentication-related functions.

These old packages have been removed:

- `/auth` (redundant code)
- `/account` (now in `/authentication/account`)
- `/login` (various packages now under `/authentication` package)
- `/token` (merged with `/authorization/token`)

Issue #646, opened 2018-09-12
Implement native logout & umbrella issue for authentication codebase refactor


sbryzak:
Original issue: https://openshift.io/openshiftio/Openshift_io/plan/detail/573

Implement token management for access and refresh tokens, allowing all user tokens to be invalidated upon user logout.

sbryzak:
Implemented in #746

PR #712, opened 2018-11-16
fix: add 3s timeout for HttpClient using option for first attempt

dipak-pawar:
Adding timeout of 10s in case if host is not ready/available instead of waiting for 30 seconds.

This is useful when you will be deploying auth and cluster as both are dependent  on each other.
e.g. auth needs to fetch clusters `http://f8cluster/api/clusters/auth` from cluster service. cluster needs to fetch public keys from auth `http://auth/api/token/keys`

Here auth will wait for 30s to fetch cluster details which needs to be up. At the same time cluster will try to fetch public keys from auth but as auth is not in ready state this will again go in multple restarts.

Fixes: https://github.com/fabric8-services/fabric8-auth/issues/713

**To be merged after https://github.com/fabric8-services/fabric8-auth/pull/689/ (to avoid conflicts).**


alien-ike:
### Ike Plugins (test-keeper)

Thank you @dipak-pawar for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/ed7ae296ac162cc631b7dcf0db85a3c333e1a8f1/.ike-prow/test-keeper.yaml).

cluster/service/cache.go
```diff
@@ -160,6 +160,7 @@ func (c *cache) createClient(ctx context.Context) (*client.Client, error) {
 	}
 
 	httpClient := http.DefaultClient
+	httpClient.Timeout = 10 * time.Second
```
xcoulon: that would actually let the `auth` service wait for 10s until the `cluster` service is up, but what about using an env var for this, so we can increase the timeout from the config maps (we could have a default value of 10s in the configuration)

alexeykazakov: Well.. the first attempt of getting the list of the clusters happens during startup in main(). And this attempt is not critical. If it fails then we move on and will try to get the list again when we really need it or every 5 minutes by the cluster cache refresher whichever occurs first. It's done in that way to avoid cycle dependencies between auth and cluster during startup.
 
The first main() attempt is what we would like to make as short as possible. But the cache refresh and lazy initialization things is more critical and I would keep them 30 seconds as we have now.

So, my suggestion is to use `options` parameter in `Start()` function - https://github.com/fabric8-services/fabric8-auth/blob/master/cluster/service/cluster.go#L60 when called from `main()` - https://github.com/fabric8-services/fabric8-auth/blob/master/main.go#L176

Something like this:
```
err = clusterservice.Start(context.Background(), config, func(c *http.Client){
	c.Timeout = 3 * time.Second
})
```
I would keep the timeout in main() very short. 3 or maybe 5 seconds, not longer. And would keep the rest of the code intact.

alexeykazakov:
//ok-without-tests

cluster/service/cache.go
dipak-pawar: you are right. Now using `options` with Timeout of 3s for first attempt from main

dipak-pawar:
@alexeykazakov I have addressed it in 1a89a6b

xcoulon:
May I suggest that we keep this PR on hold until #646 is fixed ?

main.go
```diff
@@ -173,7 +173,9 @@ func main() {
 	keycloakProfileService := login.NewKeycloakUserProfileClient()
 
 	// Try to fetch the initial list of clusters and start Cluster Service cache refresher
-	err = clusterservice.Start(context.Background(), config)
+	err = clusterservice.Start(context.Background(), config, func(c *http.Client) {
+		c.Timeout = 3 * time.Second
```
xcoulon: so is this the timeout that is going to be used for all fetches or only the first one ?

xcoulon: so, if I understand correctly, the `auth` service may start with an empty list of clusters in case the `cluster` service was not available at the startup time. Then, if the list is empty when a user request is processed, then `auth` would explicitly call the `cluster` service to get the list of clusters, and return an error if no data was available ? My point is: maybe we could check that the list of cluster is loaded before marking the `auth` service as `ready` (in the `readiness` probes), but I'm not sure it's such a good idea, because there are other parts of the service (eg: roles/permissions) that would still work (i.e., `auth` would be only partially unavailable). WDYT?

alexeykazakov: I hate Lenova touchpad! It's impossible to type without accidentally touching it and messing things up... OK. I need to type again my long response to this comment with disabled touchpad now :(

> so is this the timeout that is going to be used for all fetches or only the first one ?

It's for the first one only.

> so, if I understand correctly, the auth service may start with an empty list of clusters in case the cluster service was not available at the startup time. Then, if the list is empty when a user request is processed, then auth would explicitly call the cluster service to get the list of clusters, and return an error if no data was available ?

There are three places where we try to load the clusters.
- In main() during service startup. It can fail if the cluster service is not available for some reason. Not critical. We can leave it as is (the empty list) and start the service.
- Cluster cache. We pull a new list of clusters every few minutes in a separate goroutine in case there is any changes. It can fail as well if the cluster service is temporally unavailable. In this case we keep using the old cached list of clusters and will try to pull it again in a few minutes.
- When we really need the list of clusters (when handling a user request for example). In this case we use the cached list but if it's empty then we load it from the cluster service. If it fails then we return an error to the user.

>My point is: maybe we could check that the list of cluster is loaded before marking the auth service as ready (in the readiness probes), but I'm not sure it's such a good idea, because there are other parts of the service (eg: roles/permissions) that would still work (i.e., auth would be only partially unavailable). WDYT?

We can't relay on cluster service availability when starting Auth. It will bring a cycle dependency. Cluster service needs Auth service during startup to load the keys. But if Auth needs Cluster service to startup then you can start them both up. They will wait for each other to pass the redness prob.

dipak-pawar:
[test]

PR #729, opened 2018-12-10
Add username to authentication error message

### LINKED ISSUE #683: Add the user name in authentication error message
Currently, when trying to login with a user that was not provisioned to OSIO, the following error is returned in the request body:
```javascript
{ errors: [
    error : {
        code : 'forbidden_error',
        detail : 'user is not authorized to access OpenShift'
    }
] }
```

It would be great if you could include the name of the user (RHD username) in this error message, like this:
```javascript
{ errors: [
    error : {
        code : 'forbidden_error',
        detail : 'user is not authorized to access OpenShift: dfestal'
    }
] }
```

That would allow a much better tracking of provisioning failures.

sbryzak:
This PR uses the context to propagate username information up the stack from the user profile call (in the authentication service provider) to the token controller, where it is then used to generate the error message returned to the client.

Fixes #683

### DIFF
--- a/authentication/provider/service/authentication_provider_service.go
+++ b/authentication/provider/service/authentication_provider_service.go
@@ -365,8 +365,8 @@ func (s *authenticationProviderServiceImpl) UpdateIdentityUsingUserInfoEndPoint(
 
 	// Set the user profile's username in the context
 	profileCtx := ctx.Value(provider.UserProfileContextKey)
-	if profileCtx != nil {
-		profileCtx.(*provider.UserProfileContext).Username = &userProfile.Username
+	if pCtx, ok := profileCtx.(*provider.UserProfileContext); ok {
+		pCtx.Username = &userProfile.Username
 	}
 
 	identity := &account.Identity{}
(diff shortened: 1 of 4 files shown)


controller/token.go
```diff
@@ -212,6 +218,11 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
 
 	if notApprovedRedirect != nil && token == nil {
 		// the code enters this block only if the user is not provisioned on OSO.
+		userProfileContext := profileCtx.Value(provider.UserProfileContextKey).(*provider.UserProfileContext)
```
dipak-pawar: How about using `userProfileContext, ok := profileCtx.Value(provider.UserProfileContextKey).(*provider.UserProfileContext)` to avoid panic trigger?

authentication/provider/service/authentication_provider_service.go
```diff
@@ -363,6 +363,12 @@ func (s *authenticationProviderServiceImpl) UpdateIdentityUsingUserInfoEndPoint(
 		return nil, errors.New("unable to get user profile " + err.Error())
 	}
 
+	// Set the user profile's username in the context
+	profileCtx := ctx.Value(provider.UserProfileContextKey)
+	if profileCtx != nil {
+		profileCtx.(*provider.UserProfileContext).Username = &userProfile.Username
```
dipak-pawar: How about using
```go
if pCtx, ok := profileCtx.(*provider.UserProfileContext); ok {
pCtx.Username = &userProfile.Username
``` 
to avoid panic trigger?

sbryzak:
@dipak-pawar thanks for the review! I've implemented your suggested changes, they make for more elegant code.

PR #738, opened 2018-12-13
Re-provision user by reg app request

alexeykazakov:
If some user was manually deprovisioned in the OSO registration app but was not blocked (due to inactivity for example) then the user can reactivate the account by using signup again. In this case the reg app will call Auth to create a new user. Currently Auth will return 409 for such users. This PR re-provision such users and returns 200 to the reg app.

### DIFF
--- a/controller/users.go
+++ b/controller/users.go
@@ -133,7 +133,10 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
 			return jsonapi.JSONErrorResponse(ctx, errors.NewVersionConflictError("user with such email or username already exists"))
 		}
 		if idn.User.Deprovisioned {
-			c.app.UserService().ResetDeprovision(ctx, idn.User)
+			err := c.app.UserService().ResetDeprovision(ctx, idn.User)
+			if err != nil {
+				return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+			}
 		}
 		// User/identity already exist. Just return it.
 		return ctx.OK(ConvertToAppUser(ctx.RequestData, &idn.User, idn, true))
--- a/controller/users.go
+++ b/controller/users.go
@@ -135,6 +135,7 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
 		if idn.User.Deprovisioned {
 			err := c.app.UserService().ResetDeprovision(ctx, idn.User)
 			if err != nil {
+				log.Error(ctx, map[string]interface{}{"err": err, "username": ctx.Payload.Data.Attributes.Username, "email": ctx.Payload.Data.Attributes.Email}, "unable to re-provision user")
 				return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
 			}
 		}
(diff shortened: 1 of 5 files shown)


controller/users.go
```diff
@@ -118,7 +118,25 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
...
+				"username": ctx.Payload.Data.Attributes.Username,
+				"email":    ctx.Payload.Data.Attributes.Email,
+			}, "unable to lookup identity by username and email")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+		}
+		if idn == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewVersionConflictError("user with such email or username already exists"))
+		}
+		if idn.User.Deprovisioned {
+			c.app.UserService().ResetDeprovision(ctx, idn.User)
```
golangcibot: Error return value of `(github.com/fabric8-services/fabric8-auth/application/service.UserService).ResetDeprovision` is not checked (from `errcheck`)

alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/738/commits/186490418758494d2af8707b6ac7288e5e71dc12

controller/users.go
```diff
@@ -118,7 +118,29 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
 		return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
 	}
 	if userExists {
-		return jsonapi.JSONErrorResponse(ctx, errors.NewVersionConflictError("user with such email or username already exists"))
+		// This may happen for manually deprovisioned users which are reactivating their account via the registration app
+		// We should re-deprovision such user
+		idn, err := c.app.UserService().IdentityByUsernameAndEmail(ctx, ctx.Payload.Data.Attributes.Username, ctx.Payload.Data.Attributes.Email)
```
sbose78: The other `userExistsInDB(..) method could be replaced by this. If an identity is returned, then this retrieval of the user/identity second time is not needed.

controller/users.go
```diff
@@ -118,7 +118,29 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
...
+		if err != nil {
+			log.Error(ctx, map[string]interface{}{
+				"err":      err,
+				"username": ctx.Payload.Data.Attributes.Username,
+				"email":    ctx.Payload.Data.Attributes.Email,
+			}, "unable to lookup identity by username and email")
+			return jsonapi.JSONErrorResponse(ctx, errors.NewInternalError(ctx, err))
+		}
+		if idn == nil {
+			return jsonapi.JSONErrorResponse(ctx, errors.NewVersionConflictError("user with such email or username already exists"))
```
sbose78: Given that control entered this block because the call to `userExistsInDB(ctx context.Context, email string, username string) ` had returned a true, `idn==nil` is very unlikely to happen. If we retrieve only once ( as per my previous comment ), then we can get rid of this altogether.

alexeykazakov: It can happen. If the reg app tries to create a user with already taken email (emails are not unique in oso) then we should return 409. We should deffer it from the case when we have a perfect match - username&email to some existing deprovisioned user.

sbose78: Understood!

PR #745, opened 2018-12-17
feat: call cluster service to create identity cluster relation while creating user

dipak-pawar:
Fixes: #744

### DIFF
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -62,6 +62,8 @@ type ClusterService interface {
 	Clusters(ctx context.Context, options ...rest.HTTPClientOption) ([]cluster.Cluster, error)
 	ClusterByURL(ctx context.Context, url string, options ...rest.HTTPClientOption) (*cluster.Cluster, error)
 	Status(ctx context.Context, options ...rest.HTTPClientOption) (bool, error)
+	RemoveIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
+	AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
 	Stop()
 }
 
--- a/application/service/services.go
+++ b/application/service/services.go
@@ -62,8 +62,8 @@ type ClusterService interface {
 	Clusters(ctx context.Context, options ...rest.HTTPClientOption) ([]cluster.Cluster, error)
 	ClusterByURL(ctx context.Context, url string, options ...rest.HTTPClientOption) (*cluster.Cluster, error)
 	Status(ctx context.Context, options ...rest.HTTPClientOption) (bool, error)
-	RemoveIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
-	AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
+	UnlinkIdentityFromCluster(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
+	LinkIdentityToCluster(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
 	Stop()
 }
 
--- a/cluster/cache.go
+++ b/cluster/cache.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"net/http"
-	"net/url"
 	"sync"
 	"time"
 
@@ -13,7 +12,6 @@ import (
 	"github.com/fabric8-services/fabric8-auth/rest"
 	"github.com/fabric8-services/fabric8-cluster-client/cluster"
 
-	goaclient "github.com/goadesign/goa/client"
 	"github.com/pkg/errors"
 )
 
@@ -100,7 +98,8 @@ func (c *cache) refreshCache(ctx context.Context) error {
 
 // fetchClusters fetches a new list of clusters from Cluster Management Service
 func (c *cache) fetchClusters(ctx context.Context) (map[string]Cluster, error) {
-	cln, err := c.createClientWithServiceAccountSigner(ctx)
+	signer := NewJWTSASigner(ctx, c.config, c.options...)
+	cln, err := signer.CreateSignedClient()
 	if err != nil {
 		return nil, err
 	}
@@ -147,39 +146,4 @@ func (c *cache) fetchClusters(ctx context.Context) (map[string]Cluster, error) {
 		}
 	}
 	return clusterMap, nil
-}
-
-// createClientWithSASigner creates a client with a JWT signer which uses the Auth Service Account token
-func (c *cache) createClientWithServiceAccountSigner(ctx context.Context) (*cluster.Client, error) {
-	cln, err := c.createClient(ctx)
-	if err != nil {
-		return nil, err
-	}
-	m, err := manager.DefaultManager(c.config)
-	if err != nil {
-		return nil, err
-	}
-	signer := m.AuthServiceAccountSigner()
-	cln.SetJWTSigner(signer)
-	return cln, nil
-}
-
-func (c *cache) createClient(ctx context.Context) (*cluster.Client, error) {
-	u, err := url.Parse(c.config.GetClusterServiceURL())
-	if err != nil {
-		return nil, err
-	}
-
-	httpClient := http.DefaultClient
-
-	if c.options != nil {
-		for _, opt := range c.options {
-			opt(httpClient)
-		}
-	}
-	cln := cluster.New(goaclient.HTTPClientDoer(httpClient))
-
-	cln.Host = u.Host
-	cln.Scheme = u.Scheme
-	return cln, nil
 }
--- a/cluster/service/cluster.go
+++ b/cluster/service/cluster.go
@@ -17,9 +17,11 @@ import (
 	"github.com/fabric8-services/fabric8-auth/log"
 	"github.com/fabric8-services/fabric8-auth/rest"
 	clusterclient "github.com/fabric8-services/fabric8-cluster-client/cluster"
+	goaclient "github.com/goadesign/goa/client"
 	"github.com/pkg/errors"
 	"github.com/satori/go.uuid"
 	"net/http"
+	"net/url"
 )
 
 type clusterServiceConfig interface {
@@ -83,11 +85,12 @@ func (s *clusterService) Stop() {
 	}
 }
 
-func (s *clusterService) AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error {
-	signer := cluster.NewJWTSASigner(ctx, s.config, options...)
-	remoteClusterService, err := signer.CreateSignedClient()
+// LinkIdentityToCluster links Identity To Cluster using Cluster URL
+func (s *clusterService) LinkIdentityToCluster(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error {
+	signer := newJWTSASigner(ctx, s.config, options...)
+	remoteClusterService, err := signer.createSignedClient()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to create JWT signer for cluster service")
 	}
 	identityToClusterData := &clusterclient.LinkIdentityToClusterData{
 		ClusterURL: clusterURL,
@@ -95,7 +98,7 @@ func (s *clusterService) AddIdentityToClusterLink(ctx context.Context, identityI
 	}
 	res, err := remoteClusterService.LinkIdentityToClusterClusters(goasupport.ForwardContextRequestID(ctx), clusterclient.LinkIdentityToClusterClustersPath(), identityToClusterData)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to link identity %s to cluster having url %s", identityID, clusterURL)
 	}
 	defer rest.CloseResponse(res)
 	bodyString := rest.ReadBody(res.Body) // To prevent FDs leaks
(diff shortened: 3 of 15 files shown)


cluster/service/cluster.go
```diff
@@ -77,6 +83,62 @@ func (s *clusterService) Stop() {
 	}
 }
 
+func (s *clusterService) AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error {
+	signer := cluster.NewJWTSASigner(ctx, s.config, options...)
+	remoteClusterService, err := signer.CreateSignedClient()
+	if err != nil {
+		return err
```
xcoulon: maybe wrap this error?

cluster/service/cluster.go
```diff
@@ -77,6 +83,62 @@ func (s *clusterService) Stop() {
...
+	}
+	identityToClusterData := &clusterclient.LinkIdentityToClusterData{
+		ClusterURL: clusterURL,
+		IdentityID: identityID.String(),
+	}
+	res, err := remoteClusterService.LinkIdentityToClusterClusters(goasupport.ForwardContextRequestID(ctx), clusterclient.LinkIdentityToClusterClustersPath(), identityToClusterData)
+	if err != nil {
+		return err
```
xcoulon: same here, wrap the error?

cluster/service/cluster.go
```diff
@@ -77,6 +83,62 @@ func (s *clusterService) Stop() {
...
+	}
+	identityToClusterData := &clusterclient.UnLinkIdentityToClusterdata{
+		ClusterURL: clusterURL,
+		IdentityID: identityID.String(),
+	}
+	res, err := remoteClusterService.RemoveIdentityToClusterLinkClusters(goasupport.ForwardContextRequestID(ctx), clusterclient.RemoveIdentityToClusterLinkClustersPath(), identityToClusterData)
+	if err != nil {
+		return err
```
xcoulon: same here

cluster/service/cluster.go
```diff
@@ -93,10 +155,10 @@ func Start(ctx context.Context, factory service.ClusterCacheFactory, options ...
 			} else {
 				clusterCache = nil
 			}
-			return (clusterCache != nil && started == uint32(1)), err
+			return clusterCache != nil && started == uint32(1), err
```
xcoulon: if this code tries to start the cache only once, then maybe https://golang.org/pkg/sync/#Once.Do would be more appropriate ?

cluster/service/cluster.go
```diff
@@ -109,9 +171,9 @@ func Clusters(clusters map[string]cluster.Cluster) []cluster.Cluster {
 }
 
 func ClusterByURL(clusters map[string]cluster.Cluster, url string) *cluster.Cluster {
-	for apiURL, cluster := range clusters {
+	for apiURL, c := range clusters {
```
xcoulon: good work in avoiding var names with package names 👍

controller/users.go
```diff
@@ -155,6 +155,18 @@ func (c *UsersController) Create(ctx *app.CreateUsersContext) error {
...
+	// we create a identity cluster relationship in cluster management service.
+	clusterURL := ctx.Payload.Data.Attributes.Cluster
+	err = c.app.ClusterService().AddIdentityToClusterLink(ctx.Context, identityID, clusterURL)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{
+			"err":         err,
+			"identity_id": identityID,
+			"cluster_url": clusterURL,
+		}, "failed to link identity to cluster in cluster service")
+		// Not a blocker. Log the error and proceed.
```
xcoulon: so what happens next if this error occurs? is there a way to retry the Identity/cluster link creation ?

alexeykazakov: This error is critical enough to fail here instead of proceed I think.
We should roll back user creation here (hard delete user&identity from DB) and return an error, so, the reg app can repeat provisioning later.
But this is not that easy to implement and will complicate that PR. So, let's add a TODO here and create a new issue to address this in a separate PR.

application/service/services.go
```diff
@@ -62,6 +62,8 @@ type ClusterService interface {
 	Clusters(ctx context.Context, options ...rest.HTTPClientOption) ([]cluster.Cluster, error)
 	ClusterByURL(ctx context.Context, url string, options ...rest.HTTPClientOption) (*cluster.Cluster, error)
 	Status(ctx context.Context, options ...rest.HTTPClientOption) (bool, error)
+	RemoveIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
+	AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error
```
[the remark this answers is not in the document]
sbryzak: @alexeykazakov yes, I like your naming suggestions better

controller/users.go
dipak-pawar: Created https://github.com/fabric8-services/fabric8-auth/issues/747

test/recorder/cluster.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/dnaeon/go-vcr/cassette"
+	"github.com/dnaeon/go-vcr/recorder"
+	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/fabric8-services/fabric8-cluster-client/cluster"
+	"net/http"
+	"strings"
+)
+
+type RequestParamMatcher interface {
+	MatchClusterURL(r cassette.Request, url string) bool
```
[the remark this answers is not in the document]
dipak-pawar: Ideally yes we should match `identityID`, but I am using same identityID and different clusterURL, to match request from yaml file

cluster/cache.go
```diff
@@ -100,7 +98,8 @@ func (c *cache) refreshCache(ctx context.Context) error {
 
 // fetchClusters fetches a new list of clusters from Cluster Management Service
 func (c *cache) fetchClusters(ctx context.Context) (map[string]Cluster, error) {
-	cln, err := c.createClientWithServiceAccountSigner(ctx)
+	signer := NewJWTSASigner(ctx, c.config, c.options...)
```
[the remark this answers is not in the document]
dipak-pawar: but how do we get a list of all cluster URL to fetch it once? Or are you saying when there is a request for any cluster explicitly from any client, then we should get it from cluster service each time?

test/recorder/cluster.go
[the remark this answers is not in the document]
dipak-pawar: Created https://github.com/fabric8-services/fabric8-auth/issues/748

test/recorder/recorder.go
```diff
@@ -50,46 +38,45 @@ func New(cassetteName string, options ...Option) (*recorder.Recorder, error) {
...
+	authorization := httpRequest.Header.Get("Authorization")
+	reqID := httpRequest.Header.Get("X-Request-Id")
+
+	return "Bearer "+r.saToken == authorization && reqID == r.requestID
+}
 
-		if NotifyRequestPayloadMatcher(messageID)(httpRequest, cassetteRequest) {
-			authorization := httpRequest.Header.Get("Authorization")
-			reqID := httpRequest.Header.Get("X-Request-Id")
+// WithLinkIdentityToClusterRequestPayloadMatcher an option to specify the RequestPayload matcher for the recorder
```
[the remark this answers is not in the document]
dipak-pawar: Addressed in ebaf3d2

test/recorder/cluster.go
```diff
@@ -0,0 +1,96 @@
...
+// WithLinkIdentityToClusterRequestPayloadMatcher an option to specify the RequestPayload matcher for the recorder
+func WithLinkIdentityToClusterRequestPayloadMatcher(clusterURL, requestID, saToken string) Option {
+	linkIdentityToClusterMatcher := &LinkIdentityToClusterMatcher{DefaultRequestMatcher{requestID: requestID, saToken: saToken}}
+
+	return func(r *recorder.Recorder) {
+		r.SetMatcher(linkIdentityToClusterMatcher.RequestMatcher(clusterURL))
+	}
+}
+
+// WithLinkIdentityToClusterRequestPayloadMatcher an option to specify the RequestPayload matcher for the recorder
```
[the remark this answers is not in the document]
dipak-pawar: Addressed in ebaf3d2

cluster/service/cluster.go
```diff
@@ -77,6 +83,62 @@ func (s *clusterService) Stop() {
...
+			"cluster_url":     clusterURL,
+			"response_status": res.Status,
+			"response_body":   bodyString,
+		}, "unable to link identity to cluster in cluster management service")
+		return errors.Errorf("failed to link identity to cluster in cluster management service. Response status: %s. Response body: %s", res.Status, bodyString)
+	}
+	return nil
+}
+
+func (s *clusterService) RemoveIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error {
```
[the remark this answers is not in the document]
dipak-pawar: Addressed in ebaf3d2

cluster/service/cluster.go
```diff
@@ -77,6 +83,62 @@ func (s *clusterService) Stop() {
 	}
 }
 
+func (s *clusterService) AddIdentityToClusterLink(ctx context.Context, identityID uuid.UUID, clusterURL string, options ...rest.HTTPClientOption) error {
```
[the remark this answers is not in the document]
dipak-pawar: Addressed in ebaf3d2

application/service/services.go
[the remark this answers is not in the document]
dipak-pawar: Addressed in ebaf3d2

cluster/service/cluster.go
xcoulon: ok

cluster/service/cluster_blackbox_test.go
```diff
@@ -48,37 +49,154 @@ func (s *ClusterServiceTestSuite) TestClustersFail() {
...
+		clusterURL := "https://cluster.unauthorized/"
+		r, err := recorder.New("../../test/data/cluster/unlink_identity_to_cluster", recorder.WithUnLinkIdentityToClusterRequestPayloadMatcher(clusterURL, reqID, s.saToken))
+		require.NoError(t, err)
+		defer func() { require.NoError(t, stopRecorder(r)) }()
+
+		err = s.Application.ClusterService().UnlinkIdentityFromCluster(ctx, identityID, clusterURL, rest.WithRoundTripper(r.Transport))
+		require.EqualError(t, err, "failed to unlink identity to cluster in cluster management service. Response status: 401 Unauthorized. Response body: unauthorized")
+	})
+
+	s.T().Run("404", func(t *testing.T) {
```
xcoulon: in its current state, the `POST /api/clusters/identities` and `DELETE /api/clusters/identities` endpoints cannot return `404 Not Found` responses (https://github.com/fabric8-services/fabric8-cluster/blob/master/design/clusters.go#L125-L149), so this test may be necessary (but I guess there's nothing wrong in supporting this case at the client level anyways...)

dipak-pawar: it returns for DELETE look at https://github.com/fabric8-services/fabric8-cluster/blob/master/design/clusters.go#L146

xcoulon: oh thanks, I missed that :/ But it's not used in the controller (https://github.com/fabric8-services/fabric8-cluster/blob/master/controller/clusters.go#L175-L198) and IMO, we shouldn't return a `404 Not Found`, but a `400 Bad Request` if the parameters in the payload are invalid, IMO

dipak-pawar: Hmm, but it might be possible parameters in the payload are valid and there is no cluster available in DB. So, in that case, we should return 404 right?

xcoulon: you mean, returning a `400 Bad Request` if the identity ID is not a valid UUID or if the cluster URL is not a parseable URL, but returning a `404 Not Found` response if there's no record matching the identityID/clusterURL ? Yeah, I guess we could do that, indeed.

dipak-pawar: yes exactly, but we can do that only for clusterURL, because we try to find cluster using url. We don't try to find identity using identityID as we don't have any identity info in cluster. And I think it would be too much complext to check existence of identity by talking to auth.

xcoulon: but for unlinking, the `404` response would mean that the identity/cluster link record was not found, right? No need to call `auth` to check that there's actually an identify matching the ID?

dipak-pawar: yes right

dipak-pawar: btw we are already covering this. Look at https://github.com/fabric8-services/fabric8-cluster/blob/master/controller/clusters_blackbox_test.go#L354

xcoulon: ok, cool then ;)

cluster/cache.go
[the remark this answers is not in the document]
alexeykazakov: External clients (such as tenant service, oso-proxy, jenkins-idler, etc) are proxied to cluster service. The cache is not used.
Internal clients (different auth services, like account linking) should not ask for the list of all clusters at all. They should ask for the particular cluster with particular URL. And they should call our cluster service every time.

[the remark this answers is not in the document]
alexeykazakov: But as I said let's just create an issue for that and do it in a separate PR.

test/recorder/cluster.go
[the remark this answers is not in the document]
alexeykazakov: Sure but with IdentityID marcher you would also make sure that the identityID is set properly in the out-coming request. Right now if you don't pass the identity ID from the service to cluster the test won't fail.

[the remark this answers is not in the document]
dipak-pawar: I will improve it when introducing ginko

cluster/cache.go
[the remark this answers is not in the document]
dipak-pawar: Created https://github.com/fabric8-services/fabric8-auth/issues/750

PR #746, opened 2018-12-19
Implement Token Management - Registration and Logout

### LINKED ISSUE #726: Token management - Implement Token Registration
Currently the token service only provides support for management of RPT tokens.  We need to extend this to include management of all token types, including access tokens and refresh tokens.  All tokens must have corresponding entries in the database, and user logout should invalidate all current tokens for that user.

sbryzak:
This pull request implements token management functionality.

- Sets `LOGGED_OUT` status for all user tokens when user logs out
- Registers access and refresh tokens when user logs in
- Registers new access and refresh tokens when token is refreshed

Fixes #726


authentication/logout/service/logout_service.go
```diff
@@ -72,7 +73,10 @@ func (s *logoutServiceImpl) Logout(ctx context.Context, redirectURL string) (str
 	parameters.Add("redirect_uri", redirectURL)
 	logoutURL.RawQuery = parameters.Encode()
 
-	// TODO implement logout business logic here, i.e. invalidate all of the user's tokens
+	// If a token string was passed, then logout all tokens for the same identity
+	if tokenString != nil {
+		s.Services().TokenService().SetStatusForAllIdentityTokens(ctx, *tokenString, token.TOKEN_STATUS_LOGGED_OUT)
```
golangcibot: Error return value of `(github.com/fabric8-services/fabric8-auth/application/service.TokenService).SetStatusForAllIdentityTokens` is not checked (from `errcheck`)

authorization/token/service/token_service.go
```diff
@@ -301,20 +300,8 @@ func (s *tokenServiceImpl) Audit(ctx context.Context, identity *accountrepo.Iden
...
-			ExpiryTime: time.Unix(expiresAt, 0),
-		}
-
-		// Persist the token record to the database
-		err = s.Repositories().TokenRepository().Create(ctx, newTokenRecord)
-		if err != nil {
-			return errors.NewInternalError(ctx, err)
-		}
+		// Register the token record
+		newTokenRecord, err := s.RegisterToken(ctx, identity.ID, newTokenID, authtoken.TOKEN_TYPE_RPT, time.Unix(expiresAt, 0))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

authorization/token/service/token_service.go
```diff
@@ -387,14 +362,34 @@ func (s *tokenServiceImpl) ExchangeRefreshToken(ctx context.Context, accessToken
...
+
+	tokenID, err := s.tokenManager.ExtractTokenID(ctx, generatedToken.AccessToken)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{"error": err}, "could not extract token ID from token string")
+		return nil, errors.NewBadParameterErrorFromString("tokenString", generatedToken.AccessToken,
+			"could not extract token ID from token string")
+	}
+
+	// Register the access token
+	s.RegisterToken(ctx, identity.ID, *tokenID, authtoken.TOKEN_TYPE_ACCESS, generatedToken.Expiry)
```
golangcibot: Error return value of `s.RegisterToken` is not checked (from `errcheck`)

authorization/token/service/token_service.go
```diff
@@ -387,14 +362,34 @@ func (s *tokenServiceImpl) ExchangeRefreshToken(ctx context.Context, accessToken
...
+
+	tokenID, err = s.tokenManager.ExtractTokenID(ctx, generatedToken.RefreshToken)
+	if err != nil {
+		log.Error(ctx, map[string]interface{}{"error": err}, "could not extract token ID from token string")
+		return nil, errors.NewBadParameterErrorFromString("tokenString", generatedToken.RefreshToken,
+			"could not extract token ID from token string")
+	}
+
+	// Register the refresh token
+	s.RegisterToken(ctx, identity.ID, *tokenID, authtoken.TOKEN_TYPE_REFRESH, generatedToken.Expiry)
```
golangcibot: Error return value of `s.RegisterToken` is not checked (from `errcheck`)

authentication/logout/service/logout_service.go
```diff
@@ -72,7 +73,14 @@ func (s *logoutServiceImpl) Logout(ctx context.Context, redirectURL string) (str
 	parameters.Add("redirect_uri", redirectURL)
 	logoutURL.RawQuery = parameters.Encode()
 
-	// TODO implement logout business logic here, i.e. invalidate all of the user's tokens
+	// If a token string was passed, then set the status to "logged out" for all tokens with the same identity
+	if tokenString != nil {
```
alexeykazakov: Well, you actually could extract the token from the context here instead of the controller. And if there is a token then "log out" all the tokens. Then you could call this service from both controller endpoints logout and logoutv2. So, you don't have to check if the token is present twice (in controller v2 and in the service).

design/login.go
```diff
@@ -111,4 +111,26 @@ var _ = a.Resource("logout", func() {
...
+		)
+		a.Headers(func() {
+			a.Header("Referer", d.String)
+		})
+		a.Params(func() {
+			a.Param("redirect", d.String, "URL to be redirected to after successful logout. If not set then will redirect to the referrer instead.")
+		})
+		a.Description("Logout user")
+		a.Response(d.BadRequest, JSONAPIErrors)
+		a.Response(d.TemporaryRedirect)
```
alexeykazakov: I don't think we can use TemporaryRedirect here. We have to use window.location in browser to redirect to logout endpoint to be able to handle 307 responses. But we can't add any headers to such redirects (when using window.location).
We had the same problem with out account/token linking API. As a workaround we return 200 OK with a payload with redirectLocation - https://github.com/fabric8-services/fabric8-auth/blob/master/design/token.go#L162
I think we have to do the same here. So, in UI when some user clicks on logout then we will make ajax call to our logout v2 API with the access token in the Auth header and extract the redirect location from its response payload. Then use window.location to navigate to that location.

controller/logout.go
```diff
@@ -33,27 +31,37 @@ func NewLogoutController(service *goa.Service, app application.Application) *Log
 
 // Logout runs the logout action.
 func (c *LogoutController) Logout(ctx *app.LogoutLogoutContext) error {
-	return c.doLogout(ctx, ctx.Redirect, ctx.Referer, ctx.ResponseData, nil)
+	logoutRedirect, err := c.doLogout(ctx, ctx.Redirect, ctx.Referer)
+	if err != nil {
+		log.Error(ctx, nil, "Failed to logout.")
+		return jsonapi.JSONErrorResponse(ctx, err)
```
alexeykazakov: This service is also used by the old logout endpoint, right? So, if the token is nil then just set tokenString to “” otherwise we break the current API. If I’m not missing anything :)
Goa will validate the token in the request for the new endpoint for us.

authorization/token/service/token_service.go
```diff
@@ -381,184 +343,210 @@ func (s *tokenServiceImpl) ExchangeRefreshToken(ctx context.Context, accessToken
...
+	tkn := &tokenrepo.Token{
+		TokenID:    tokenID,
+		IdentityID: identityID,
+		Status:     0,
+		TokenType:  tokenType,
+		ExpiryTime: expiryTime,
+	}
+
+	// Persist the token record to the database
+	s.ExecuteInTransaction(func() error {
```
golangcibot: Error return value of `s.ExecuteInTransaction` is not checked (from `errcheck`)

controller/logout.go
```diff
@@ -16,19 +17,52 @@ type LogoutController struct {
...
 	return &LogoutController{Controller: service.NewController("LogoutController"), app: app}
 }
 
 // Logout runs the logout action.
 func (c *LogoutController) Logout(ctx *app.LogoutLogoutContext) error {
-	redirect := ctx.Redirect
-	referrer := ctx.Referer
+	logoutRedirect, err := c.doLogout(ctx, ctx.Redirect, ctx.Referer)
+	if err != nil {
+		log.Error(ctx, nil, "Failed to logout.")
```
xcoulon: why don't you include the error here?

sbryzak:
@alexeykazakov thanks, I'm just working on a last minute issue that has popped up... when we manually populate the token claims for testing (in [this method](https://github.com/fabric8-services/fabric8-auth/blob/master/test/login.go#L26)) the `.Raw()` function doesn't return the token string, which means when I call it in `logout_service.go` it currently returns an empty string and so the test fails.  I'm currently trying to find a workaround for this.

sbryzak:
I've resolved the issue with the token `Raw()` function, however another issue has presented itself.. how should we handle requests where the token contains a `sub` claim that doesn't represent an identity that doesn't exist in the database?  For testing we are using `testsupport.TestIdentity` which isn't backed up by an actual identity record in the database, which is causing the endpoint to fail.  @alexeykazakov @xcoulon How do you guys think we should handle this?

xcoulon:
> I've resolved the issue with the token `Raw()` function, however another issue has presented itself.. how should we handle requests where the token contains a `sub` claim that doesn't represent an identity that exists in the database? For testing we are using `testsupport.TestIdentity` which isn't backed up by an actual identity record in the database, which is causing the endpoint to fail. @alexeykazakov @xcoulon How do you guys think we should handle this?

@sbryzak you mean, a token was issues (the user existed in the DB), but before the user tries to logout, the `identity`/`user` records were removed from the db and hence you can't look them up during the logout? That would be a very rare edge case, so I suggest we log an error and forward the user to an error page (if that's possible?) Does it make sense?

sbryzak:
@xcoulon I guess I'm more questioning the nature of the tests, take [this line](https://github.com/fabric8-services/fabric8-auth/blob/master/controller/logout_blackbox_test.go#L45) of code from `logout_blackbox_test.go` for example:


`svc, ctrl := s.UnSecuredController()`

What does the "unsecured controller" function actually mean?  It's still calling `testsupport.ServiceAsUser()` which under the covers contains this code:

```
	svc := goa.New(serviceName)
	svc.Context = WithIdentity(svc.Context, u)
	svc.Context = manager.ContextWithTokenManager(svc.Context, testtoken.TokenManager)
	return svc
```

So we're creating a new service instance, with a context that contains an access token - so why is this an "unsecured" controller?  @alexeykazakov perhaps you could shed some light on this.  Also, the existing logout endpoint isn't secured (i.e. there's no jwt security constraint on the endpoint definition) so it's a little confusing as to why the tests for this endpoint are using a context that contains an access token.

sbryzak:
@alexeykazakov I've fixed the test suite so that the unsecured controller is now actually unsecured (i.e. no token).  Everything passes now, so would be great if you and/or @xcoulon could take one last look at the PR to give me the ok to merge.

test/login.go
```diff
@@ -20,6 +20,7 @@ import (
 // Token is filled using input Identity object
 func WithIdentity(ctx context.Context, ident account.Identity) context.Context {
 	token := fillClaimsWithIdentity(ident)
+	testtoken.TokenManager.SignRPTToken(ctx, token)
```
golangcibot: Error return value of `testtoken.TokenManager.SignRPTToken` is not checked (from `errcheck`)

controller/logout.go
```diff
@@ -62,7 +60,7 @@ func (c *LogoutController) doLogout(ctx jsonapi.InternalServerError, redirect *s
 	if redirect == nil {
 		if referrer == nil {
 			log.Error(ctx, nil, "Failed to logout. Referer Header and redirect param are both empty.")
-			return "", jsonapi.JSONErrorResponse(ctx, goa.ErrBadRequest("referer Header and redirect param are both empty (at least one should be specified)"))
+			return "", goa.ErrBadRequest("referer Header and redirect param are both empty (at least one should be specified)")
```
alexeykazakov: Why returning goa error instead of json error response?

controller/logout.go
```diff
@@ -77,7 +75,7 @@ func (c *LogoutController) doLogout(ctx jsonapi.InternalServerError, redirect *s
 			"redirect_url": *redirect,
 			"err":          err,
 		}, "Failed to logout. Unable to parse provided redirect url.")
-		return "", jsonapi.JSONErrorResponse(ctx, goa.ErrBadRequest(err.Error()))
+		return "", goa.ErrBadRequest(err.Error())
```
alexeykazakov: Same here

sbryzak:
@alexeykazakov I've modified the `doLogout()` function to return json errors instead.

alexeykazakov:
Use errors.NewBadParameter() instead of explicit goa.BadRequest as we do in other places.

Besides that looks good.

PR #752, opened 2019-01-02
feat(#747): rollback transaction to create user and identity when failed to link identity to cluster

dipak-pawar:
Fixes: #747

### DIFF
--- a/authentication/account/repository/identity.go
+++ b/authentication/account/repository/identity.go
@@ -127,7 +127,7 @@ type IdentityRepository interface {
 	Delete(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) error
 	DeleteForResource(ctx context.Context, resourceID string) error
 	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
-	List(ctx context.Context, funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
+	List(ctx context.Context) ([]Identity, error)
 	IsValid(context.Context, uuid.UUID) bool
 	Search(ctx context.Context, q string, start int, limit int) ([]Identity, int, error)
 	FindIdentityMemberships(ctx context.Context, identityID uuid.UUID, resourceType *string) ([]authorization.IdentityAssociation, error)
@@ -156,6 +156,7 @@ func (m Membership) TableName() string {
 
 // Load returns a single Identity as a Database Model
 // This is more for use internally, and probably not what you want in  your controllers
+// arguments funcs can be used to add conditions dynamically to current database connection
 func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) (*Identity, error) {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "load"}, time.Now())
 
@@ -254,7 +255,7 @@ func (m *GormIdentityRepository) Save(ctx context.Context, model *Identity) erro
 	return errs.WithStack(err)
 }
 
-// Delete removes a single record.
+// Delete removes a single record. argument funcs can be used to add conditions dynamically to current database connection
 func (m *GormIdentityRepository) Delete(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) error {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "delete"}, time.Now())
 
@@ -368,11 +369,11 @@ func IdentityFilterByProviderType(providerType string) func(db *gorm.DB) *gorm.D
 }
 
 // List return all user identities
-func (m *GormIdentityRepository) List(ctx context.Context, funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error) {
+func (m *GormIdentityRepository) List(ctx context.Context) ([]Identity, error) {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "list"}, time.Now())
 	var rows []Identity
 
-	err := m.db.Scopes(funcs...).Model(&Identity{}).Order("username").Find(&rows).Error
+	err := m.db.Model(&Identity{}).Order("username").Find(&rows).Error
 	if err != nil && err != gorm.ErrRecordNotFound {
 		return nil, errs.WithStack(err)
 	}
--- a/authentication/account/repository/identity.go
+++ b/authentication/account/repository/identity.go
@@ -119,15 +119,15 @@ func NewIdentityRepository(db *gorm.DB) *GormIdentityRepository {
 // IdentityRepository represents the storage interface.
 type IdentityRepository interface {
 	base.Exister
-	Load(ctx context.Context, id uuid.UUID) (*Identity, error)
+	Load(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) (*Identity, error)
 	LoadWithUser(ctx context.Context, id uuid.UUID) (*Identity, error)
 	Create(ctx context.Context, identity *Identity) error
 	Lookup(ctx context.Context, username, profileURL, providerType string) (*Identity, error)
 	Save(ctx context.Context, identity *Identity) error
-	Delete(ctx context.Context, id uuid.UUID) error
+	Delete(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) error
 	DeleteForResource(ctx context.Context, resourceID string) error
 	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
-	List(ctx context.Context) ([]Identity, error)
+	List(ctx context.Context, funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
 	IsValid(context.Context, uuid.UUID) bool
 	Search(ctx context.Context, q string, start int, limit int) ([]Identity, int, error)
 	FindIdentityMemberships(ctx context.Context, identityID uuid.UUID, resourceType *string) ([]authorization.IdentityAssociation, error)
@@ -156,11 +156,11 @@ func (m Membership) TableName() string {
 
 // Load returns a single Identity as a Database Model
 // This is more for use internally, and probably not what you want in  your controllers
-func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID) (*Identity, error) {
+func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) (*Identity, error) {
 	defer goa.MeasureSince([]string{"goa", "db", "identity", "load"}, time.Now())
 
 	var native Identity
-	err := m.db.Table(m.TableName()).Where("id = ?", id).Find(&native).Error
+	err := m.db.Scopes(funcs...).Table(m.TableName()).Where("id = ?", id).Find(&native).Error
 	if err == gorm.ErrRecordNotFound {
 		return nil, errs.WithStack(errors.NewNotFoundError("identity", id.String()))
 	}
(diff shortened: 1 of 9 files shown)


controller/users.go
```diff
@@ -278,6 +287,22 @@ func (c *UsersController) createUserInDB(ctx *app.CreateUsersContext, identityID
 	return identity, user, nil
 }
 
+func (c UsersController) deleteUser(ctx context.Context, userID uuid.UUID, identityID uuid.UUID) error {
```
alexeykazakov: Worth moving it to authentication/account/service/user.go
Something like `HardDeleteUser(ctx context.Context, identity Identity) error`. We will need to start moving the existing business logic from user(s) controllers to services later but for now let's add new stuff to services.

authentication/account/repository/identity.go
```diff
@@ -156,11 +156,11 @@ func (m Membership) TableName() string {
 
 // Load returns a single Identity as a Database Model
 // This is more for use internally, and probably not what you want in  your controllers
-func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID) (*Identity, error) {
+func (m *GormIdentityRepository) Load(ctx context.Context, id uuid.UUID, funcs ...func(*gorm.DB) *gorm.DB) (*Identity, error) {
```
alexeykazakov: Add a param description saying that funcs can be used to add dynamic conditions to all methods where you are adding this param.

authentication/account/repository/user.go
```diff
@@ -174,7 +174,7 @@ func (m *GormUserRepository) Delete(ctx context.Context, id uuid.UUID) error {
 }
 
 // List return all users
-func (m *GormUserRepository) List(ctx context.Context) ([]User, error) {
+func (m *GormUserRepository) List(ctx context.Context, funcs ...func(*gorm.DB) *gorm.DB) ([]User, error) {
```
alexeykazakov: It seems that you don't use funcs in this method. Then why adding it?

authentication/account/repository/identity_blackbox_test.go
```diff
@@ -43,6 +45,55 @@ func (s *IdentityRepositoryTestSuite) TestDelete() {
...
+		// then
+		require.Nil(t, err)
+
+		identity, err := s.Application.Identities().Load(s.Ctx, identity1.ID(), unscoped)
+		require.EqualError(t, err, fmt.Sprintf("identity with id '%s' not found", identity1.ID()))
+		require.Nil(t, identity)
+
+		identities, err := s.Application.Identities().List(s.Ctx)
+		require.NoError(t, err, "Could not list identities")
+		require.True(t, len(identities) >= 1)
```
alexeykazakov: Or you could use `require.NotEmpty(t, identities)`

authentication/account/repository/identity_blackbox_test.go
```diff
@@ -43,6 +45,55 @@ func (s *IdentityRepositoryTestSuite) TestDelete() {
...
+		// when hard delete user
+		err := s.Application.Identities().Delete(s.Ctx, identity1.ID(), unscoped)
+		// then
+		require.Nil(t, err)
+
+		identity, err := s.Application.Identities().Load(s.Ctx, identity1.ID(), unscoped)
+		require.EqualError(t, err, fmt.Sprintf("identity with id '%s' not found", identity1.ID()))
+		require.Nil(t, identity)
+
+		identities, err := s.Application.Identities().List(s.Ctx)
```
alexeykazakov: We don't really need to use List() here, do we? We need to pass scopes for `Delete()` to be able to hard delete identities. We need to pass scopes to `Load()` to test the record has been deleted. But we don't have to add it to List(), do we?
I would suggest not to modify List() at all. In your tests just check that the record is not found for both `Load(s.Ctx, identity1.ID(), unscoped)` and `Load(s.Ctx, identity1.ID())`.
And the same for User.

controller/users_blackbox_test.go
```diff
@@ -1430,6 +1431,55 @@ func (s *UsersControllerTestSuite) TestCreateUserAsServiceAccountWithAllFieldsOK
...
+	secureService, secureController := s.SecuredServiceAccountController(testsupport.TestOnlineRegistrationAppIdentity)
+
+	// when
+	createUserPayload := newCreateUsersPayload(&user.Email, &user.FullName, &user.Bio, &user.ImageURL, &user.URL, &user.Company, &identity.Username, &rhdUserName, user.ID.String(), &user.Cluster, &identity.RegistrationCompleted, &approved, user.ContextInformation)
+
+	// then
+	test.CreateUsersInternalServerError(s.T(), secureService.Context, secureService, secureController, createUserPayload)
+
+	// verify identity and user is deleted
+	err := s.Application.Identities().CheckExists(context.TODO(), identity.ID.String())
```
alexeykazakov: Soft deleted identities and users won't be found here. You need to use Load() with unscoped to make sure the records where hard deleted.

controller/users.go
dipak-pawar: Addressed in 72b25766ca1fbc16a01c54b7e801702037a4b174

authentication/account/repository/user.go
dipak-pawar: Good Catch. Addressed in 72b2576

authentication/account/repository/identity_blackbox_test.go
dipak-pawar: Addressed in 72b2576

authentication/account/repository/identity_blackbox_test.go
dipak-pawar: Addressed in 72b2576

controller/users_blackbox_test.go
dipak-pawar: Addressed in 72b2576

authentication/account/repository/identity.go
dipak-pawar: Addressed in 72b2576

authentication/account/repository/identity_blackbox_test.go
```diff
@@ -36,13 +38,59 @@ func (s *IdentityRepositoryTestSuite) TestDelete() {
...
 
+	s.T().Run("ok hard delete by identity ID", func(t *testing.T) {
+		// given
+		g := s.NewTestGraph(t)
+		identity1 := g.CreateIdentity()
+		// create a second identity
+		g.CreateIdentity()
+
+		unscoped := func(s *gorm.DB) *gorm.DB {
+			return s.Unscoped()
```
nurali-techie: This variable name `s *gorm.DB` mixed up with `s` for suite .. renaming this variable would make things clearer.

controller/users_blackbox_test.go
```diff
@@ -1430,6 +1432,69 @@ func (s *UsersControllerTestSuite) TestCreateUserAsServiceAccountWithAllFieldsOK
 	require.Equal(s.T(), uint64(1), s.clusterServiceMock.LinkIdentityToClusterCounter)
 }
 
+func (s *UsersControllerTestSuite) TestCreateUserAsServiceAccountWhenFailedForLinkingIdentityToCluster() {
```
nurali-techie: This is really big name, can we think of some better little small name :-)  Go is not as much as verbose as Java and we should try to align with Go verbosity.  If we see go src code, we generally find small names, we should take inspiration from same.

dipak-pawar: I am fine with it as long as it's readable instead of using short names and getting confused

authentication/account/repository/identity_blackbox_test.go
dipak-pawar: addressed in df44f17

dipak-pawar:
@alexeykazakov Added tests for service method in  5ca52f3

dipak-pawar:
@xcoulon: This is a really good Point. Addressed in 4f1dcd25c3931d09bc6c5b915165c6a559ea72d9

authentication/account/repository/identity_blackbox_test.go
```diff
@@ -79,6 +80,25 @@ func (s *IdentityRepositoryTestSuite) TestLoad() {
...
+
+		// when soft delete identity
+		err := s.Application.Identities().Delete(s.Ctx, identity1.ID())
+		// then
+		require.Nil(t, err)
+
+		identity, err := s.Application.Identities().Load(s.Ctx, identity1.ID(), func(db *gorm.DB) *gorm.DB {
+			return db.Unscoped()
+		})
+		require.Nil(t, err)
```
xcoulon: could be `require.NoError(...)` here

authentication/account/repository/identity_blackbox_test.go
```diff
@@ -79,6 +80,25 @@ func (s *IdentityRepositoryTestSuite) TestLoad() {
...
+		// given
+		g := s.NewTestGraph(t)
+		identity1 := g.CreateIdentity()
+
+		// when soft delete identity
+		err := s.Application.Identities().Delete(s.Ctx, identity1.ID())
+		// then
+		require.Nil(t, err)
+
+		identity, err := s.Application.Identities().Load(s.Ctx, identity1.ID(), func(db *gorm.DB) *gorm.DB {
```
xcoulon: this func could be extracted to something like `includeSoftDeletes` for more clarity in the code

authentication/account/repository/user_blackbox_test.go
```diff
@@ -31,23 +32,25 @@ func (s *userBlackBoxTest) SetupTest() {
...
-	createAndLoadUser(s, false)
+	s.T().Run("ok by user ID", func(t *testing.T) {
+		// create 2 users, where the first one would be deleted.
+		user := createAndLoadUser(s, false)
+		createAndLoadUser(s, false)
 
-	err := s.repo.Delete(s.Ctx, user.ID)
-	assert.Nil(s.T(), err)
+		err := s.repo.Delete(s.Ctx, user.ID)
+		assert.Nil(t, err)
```
xcoulon: could be `assert.NoError(...)` here

authentication/account/repository/user_blackbox_test.go
```diff
@@ -31,23 +32,25 @@ func (s *userBlackBoxTest) SetupTest() {
...
 
-	// lets see how many are present.
-	users, err := s.repo.List(s.Ctx)
-	require.Nil(s.T(), err, "Could not list users")
-	require.True(s.T(), len(users) > 0)
+		// lets see how many are present.
+		users, err := s.repo.List(s.Ctx)
+		require.Nil(t, err, "Could not list users")
+		require.True(t, len(users) > 0)
```
xcoulon: could be `require.NotEmpty(..)` here

authentication/account/repository/user_blackbox_test.go
```diff
@@ -31,23 +32,25 @@ func (s *userBlackBoxTest) SetupTest() {
...
+		err := s.repo.Delete(s.Ctx, user.ID)
+		assert.Nil(t, err)
 
-	// lets see how many are present.
-	users, err := s.repo.List(s.Ctx)
-	require.Nil(s.T(), err, "Could not list users")
-	require.True(s.T(), len(users) > 0)
+		// lets see how many are present.
+		users, err := s.repo.List(s.Ctx)
+		require.Nil(t, err, "Could not list users")
```
xcoulon: could be `require.NoError(...)` here

authentication/account/repository/user_blackbox_test.go
```diff
@@ -61,6 +64,24 @@ func (s *userBlackBoxTest) TestOKToLoad() {
...
 }
 
+func (s *userBlackBoxTest) TestOKToLoadByDynamicCondition() {
+	// create 2 users, where the first one would be deleted.
+	user := createAndLoadUser(s, false)
+	createAndLoadUser(s, false)
+
+	// soft delete user
+	err := s.repo.Delete(s.Ctx, user.ID)
+	require.Nil(s.T(), err)
```
xcoulon: `require.NoError(...)`

controller/users.go
```diff
@@ -266,6 +253,17 @@ func (c *UsersController) createUserInDB(ctx *app.CreateUsersContext, identityID
 		if err != nil {
 			return err
 		}
+
```
xcoulon: What about refactoring a little more this controller and move the logic of creating the user and identity, as well as calling the remote `WIT` and `Cluster` services in the `UserService`? 
AFAIU, this is the design that we should use: controllers validate the input then call services which are responsible for carrying the transaction and the dealing with the business logic. In this case, the call to the local repositories and remote services would take place within the same transaction, as you already did here.

alexeykazakov: 1. We can't call remote service inside DB transactions.
2. Let's not blow up this PR by starting refactoring not related code. This controller requires refactoring indeed but let's do it as a separate task. We have to move forward with more important tasks right now.

dipak-pawar:
Thank you @xcoulon, @alexeykazakov

Issue #753, opened 2019-01-03
Call notification service once to send email message


dipak-pawar:
Currently, notification service is called two times to send an email message to the recipient. Make sure we are calling it only once.

https://github.com/fabric8-services/fabric8-auth/blob/master/notification/service/notification.go#L69-L74

PR #754, opened 2019-01-03
fix(#753): call notification service only once to send email message

dipak-pawar:
Fixes: #753

*Changes Proposed in this PR*
* Calls notification service only one time
* Update all tests to use gock, so that we can catch this type of issues.


dipak-pawar:
Thanks @xcoulon

PR #756, opened 2019-01-07
feat: add api to populate identity cluster relations in cluster service

dipak-pawar:
*Changes Proposed in this PR*
* ADD `curl -X POST clusters/identities/link -H Authorization: Bearer $SA_TOKEN` to create identity cluster relationship in cluster service for all existing users.

### DIFF
--- a/cluster/service/cluster.go
+++ b/cluster/service/cluster.go
@@ -165,6 +165,61 @@ func Start(ctx context.Context, factory service.ClusterCacheFactory, options ...
 	return clusterCache != nil && started == uint32(1), nil
 }
 
+func (s *clusterService) LinkExistingIdentitiesToCluster(ctx context.Context, options ...rest.HTTPClientOption) error {
+	identitiesWithClusters, err := s.Repositories().Identities().GetIdentitiesWithClusterURL(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "failed to load existing identities with default IDP")
+	}
+
+	if _, err := s.linkIdentitiesToCluster(ctx, identitiesWithClusters, options...); err != nil {
+		return errors.Wrapf(err, "failed while linking existing identities to it's cluster url")
+	}
+	return nil
+}
+
+func (s *clusterService) linkIdentitiesToCluster(ctx context.Context, identitiesWithClusterURL map[uuid.UUID]string, options ...rest.HTTPClientOption) (chan error, error) {
+	signer := newJWTSASigner(ctx, s.config, options...)
+	client, err := signer.createSignedClient()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create JWT signer for cluster service")
+	}
+
+	errs := make(chan error, len(identitiesWithClusterURL))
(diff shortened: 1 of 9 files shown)


controller/clusters.go
```diff
@@ -39,3 +44,20 @@ func (c *ClustersController) Show(ctx *app.ShowClustersContext) error {
 	}
 	return nil
 }
+
+func (c *ClustersController) LinkExistingIdentitiesToCluster(ctx *app.LinkExistingIdentitiesToClusterClustersContext) error {
+	// currently using online registration as we already have it atleast for prod-preview, so we can use it directly.
+	if !token.IsSpecificServiceAccount(ctx, token.OnlineRegistration) {
```
alexeykazakov: Let’s use some dedicated temporal service account. “migration” for example. We don’t have add any credentials for that SA. We will manually generate a token for that account. Just hardcode its name in the controller.

cluster/service/cluster.go
```diff
@@ -165,6 +165,61 @@ func Start(ctx context.Context, factory service.ClusterCacheFactory, options ...
...
+		return errors.Wrapf(err, "failed to link identity %s to cluster having url %s", identityID, clusterURL)
+	}
+	defer rest.CloseResponse(res)
+	bodyString := rest.ReadBody(res.Body) // To prevent FDs leaks
+	if res.StatusCode != http.StatusNoContent {
+		log.Error(ctx, map[string]interface{}{
+			"identity_id":     identityID,
+			"cluster_url":     clusterURL,
+			"response_status": res.Status,
+			"response_body":   bodyString,
```
alexeykazakov: Maybe add some `"clusterlink_migration" : "true", ` or something like that too? For easier searching in the logs.

cluster/service/cluster.go
```diff
@@ -165,6 +165,61 @@ func Start(ctx context.Context, factory service.ClusterCacheFactory, options ...
...
+	}
+
+	errs := make(chan error, len(identitiesWithClusterURL))
+	go func() {
+		defer close(errs)
+		for id, u := range identitiesWithClusterURL {
+			if e := s.link(ctx, client, id, u); e != nil {
+				errs <- e
+			}
+		}
```
alexeykazakov: Log the info message when the linking is started and when it's finished with the total count of the linked identities. Then we will know when the migrations is complete.

alexeykazakov:
Just for clarification. This is a temporal migration endpoint. We will need to use it once and then revert the commits.

cluster/service/cluster.go
```diff
@@ -187,11 +187,16 @@ func (s *clusterService) linkIdentitiesToCluster(ctx context.Context, identities
 	errs := make(chan error, len(identitiesWithClusterURL))
 	go func() {
 		defer close(errs)
+		log.Logger().Infof("linking identity to cluster url has been started. total_identities=%d", len(identitiesWithClusterURL))
```
alexeykazakov: Why not using log.Info() instead?

PR #764, opened 2019-01-31
Add index to TOKEN.IDENTITY_ID database column

### LINKED ISSUE #763: Add index for token table
It seems that user-based token updates may be suffering from a performance issue due to there being no index on the `TOKEN.IDENTITY_ID` database column.  We need to add this index to improve the new logout (`/logout/v2`) endpoint which invalidates all of the user's tokens.

sbryzak:
Fixes #763 

This PR should fix performance issues with the `TOKEN` table.

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -233,6 +233,9 @@ func GetMigrations(configuration MigrationConfiguration) Migrations {
 	// Version 41
 	m = append(m, steps{ExecuteSQLFile("041-identity-role-index.sql")})
 
+	// Version 42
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @sbryzak for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/b3a5a584281a7cbc2a6b4c438b97b50d160747b0/.ike-prow/test-keeper.yaml).

PR #766, opened 2019-02-01
Bulk update token status values

### LINKED ISSUE #765: Bulk update token status values instead of individual updates
Currently when logging out with the new logout endpoint, token status values are updated individually like so:

```
for _, tkn := range tokens {
  tkn.SetStatus(status, true)
  err = s.Repositories().TokenRepository().Save(ctx, &tkn)
```

For users with a large number of tokens, this might have negative performance implications.  Instead, we should perform a bulk update of the token status values for that user:

`UPDATE token SET status = status | (value for TOKEN_STATUS_LOGGED_OUT) WHERE identity_id = :identity_id`

sbryzak:
Fixes #765 

This PR adds a new method to the token repository, `SetStatusFlagsForIdentity()` which allows the bulk update of token status flags for an identity.  The logout service now uses this method instead of updating token status values individually, which should hopefully result in a performance improvement.

### DIFF
--- a/authorization/token/repository/token_blackbox_test.go
+++ b/authorization/token/repository/token_blackbox_test.go
@@ -185,3 +185,30 @@ func (s *tokenBlackBoxTest) TestCreateListPrivileges() {
 	require.Len(s.T(), privs, 1)
 	require.Equal(s.T(), pc.PrivilegeCache().PrivilegeCacheID, privs[0].PrivilegeCacheID)
 }
+
+func (s *tokenBlackBoxTest) TestSetStatusFlagsForIdentity() {
+	user1 := s.Graph.CreateUser()
+	user2 := s.Graph.CreateUser()
+
+	t1 := s.Graph.CreateToken(user1)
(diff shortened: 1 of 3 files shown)


authorization/token/repository/token_blackbox_test.go
```diff
@@ -185,3 +185,30 @@ func (s *tokenBlackBoxTest) TestCreateListPrivileges() {
...
+	t1 := s.Graph.CreateToken(user1)
+	t2 := s.Graph.CreateToken(user1)
+	s.Graph.CreateToken(user1)
+
+	t4 := s.Graph.CreateToken(user2)
+	s.Graph.CreateToken(user2)
+
+	require.True(s.T(), t1.Token().Valid())
+
+	s.repo.SetStatusFlagsForIdentity(s.Ctx, user1.IdentityID(), tokenPkg.TOKEN_STATUS_REVOKED)
```
golangcibot: Error return value of `s.repo.SetStatusFlagsForIdentity` is not checked (from `errcheck`)

PR #772, opened 2019-02-05
Add token cleanup worker

### LINKED ISSUE #767: Implement automatic cleanup for expired tokens
At present, there is no cleanup mechanism for expired tokens meaning that the token table will continually grow in size, potentially causing performance issues.

sbryzak:
Fixes #767 

This PR introduces a token cleanup worker that runs inside a go routine.  It routinely (once per hour) invokes a token cleanup routine that removes expired tokens after a configurable retention period, set by the `AUTH_EXPIRED_TOKEN_RETENTION_HOURS` configuration parameter, which has a default of 24 hours.

### DIFF
--- a/authorization/token/repository/token_blackbox_test.go
+++ b/authorization/token/repository/token_blackbox_test.go
@@ -2,6 +2,7 @@ package repository_test
 
 import (
 	"testing"
+	"time"
 
 	tokenPkg "github.com/fabric8-services/fabric8-auth/authorization/token"
 	tokenRepo "github.com/fabric8-services/fabric8-auth/authorization/token/repository"
@@ -212,4 +213,43 @@ func (s *tokenBlackBoxTest) TestSetStatusFlagsForIdentity() {
 
 	t4Loaded := s.Graph.LoadToken(t4.TokenID())
 	require.True(s.T(), t4Loaded.Token().Valid())
+}
+
+func (s *tokenBlackBoxTest) TestCleanupExpiredTokens() {
+	now := time.Now()
+	yesterday := now.AddDate(0, 0, -1)
+	tomorrow := now.AddDate(0, 0, 1)
+
+	s.Graph.CreateToken(yesterday)
+	s.Graph.CreateToken(yesterday)
+	s.Graph.CreateToken(now)
+	s.Graph.CreateToken(now)
+	s.Graph.CreateToken(now)
+	s.Graph.CreateToken(tomorrow)
+	s.Graph.CreateToken(tomorrow)
+
+	require.Equal(s.T(), s.countTokens(s.T()), 7)
+
+	// Let's start by cleaning up all tokens that expired more than 1 hour ago
+	err := s.repo.CleanupExpiredTokens(s.Ctx, 1)
+	require.NoError(s.T(), err)
+
+	// We should be left with 5 tokens (the "yesterday" tokens should now be gone)
+	require.Equal(s.T(), s.countTokens(s.T()), 5)
+
+	// Now let's clean up all the expired tokens, without any retention
+	err = s.repo.CleanupExpiredTokens(s.Ctx, 0)
+	require.NoError(s.T(), err)
+
+	// We should now be left with just 2 tokens
+	require.Equal(s.T(), s.countTokens(s.T()), 2)
+}
+
+func (s *tokenBlackBoxTest) countTokens(t *testing.T) int {
+	var result *int64
+
+	err := s.DB.Table("token").Count(&result).Error
+	require.NoError(t, err)
+
+	return int(*result)
 }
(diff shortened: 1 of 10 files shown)


authorization/token/worker/token_cleanup_worker_blackbox_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"testing"
+	"time"
+)
+
+type tokenCleanupWorkerBlackBoxTest struct {
+	gormtestsupport.DBTestSuite
+	repo tokenRepo.TokenRepository
```
golangcibot: `repo` is unused (from `structcheck`)

authorization/token/worker/token_cleanup_worker.go
```diff
@@ -0,0 +1,51 @@
...
+	w.stopCh = make(chan bool, 1)
+
+	go w.cleanupLoop()
+}
+
+func (w *tokenCleanupWorker) cleanupLoop() {
+	for {
+		select {
+		case <-w.ticker.C:
+			w.app.TokenService().CleanupExpiredTokens(w.ctx)
```
golangcibot: Error return value of `(github.com/fabric8-services/fabric8-auth/application/service.TokenService).CleanupExpiredTokens` is not checked (from `errcheck`)

authorization/token/repository/token_blackbox_test.go
```diff
@@ -213,3 +214,42 @@ func (s *tokenBlackBoxTest) TestSetStatusFlagsForIdentity() {
...
+
+	// We should be left with 5 tokens (the "yesterday" tokens should now be gone)
+	require.Equal(s.T(), s.countTokens(s.T()), 5)
+
+	// Now let's clean up all the expired tokens, without any retention
+	err = s.repo.CleanupExpiredTokens(s.Ctx, 0)
+	require.NoError(s.T(), err)
+
+	// We should now be left with just 2 tokens
+	require.Equal(s.T(), s.countTokens(s.T()), 2)
```
alexeykazakov: Let's check the specific tokens. Not the number but their IDs to make sure the expected tokens are deleted.

authorization/token/worker/token_cleanup_worker.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewTokenCleanupWorker(ctx context.Context, app application.Application) TokenCleanupWorker {
+	return &tokenCleanupWorker{
+		ctx: ctx,
+		app: app,
+	}
+}
+
+type tokenCleanupWorker struct {
+	ctx    context.Context
```
alexeykazakov: You don't use ctx which makes sense. So, let's remove it.

authorization/token/worker/token_cleanup_worker_blackbox_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Activate token cleanup once every hour
+	worker.Start(time.NewTicker(time.Second))
+	defer worker.Stop()
+
+	// Wait 5 seconds
+	time.Sleep(time.Duration(5) * time.Second)
+
+	// We should now have 1 token remaining
+	require.Equal(s.T(), s.countTokens(s.T()), 1)
```
alexeykazakov: Same here. Let's check token IDs. Not the overall numbers.

authorization/token/worker/token_cleanup_worker_blackbox_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Start the worker with a 1 second timer
+	worker := worker.NewTokenCleanupWorker(s.Ctx, s.Application)
+
+	// Activate token cleanup once every hour
+	worker.Start(time.NewTicker(time.Second))
+	defer worker.Stop()
+
+	// Wait 5 seconds
+	time.Sleep(time.Duration(5) * time.Second)
```
alexeykazakov: Don't waist 5 seconds on needless waiting in our test run time :) Start a timer with N milliseconds (10 for example) and check the results every 100 ms for up to 30 times or so in a loop. So the test execution time will be from 100ms up to 3s when it will fail by timeout.

authorization/token/worker/token_cleanup_worker.go
sbryzak: It's used further down in the cleanup thread, where it's passed to the token service.

authorization/token/repository/token_blackbox_test.go
sbryzak: Done

authorization/token/worker/token_cleanup_worker_blackbox_test.go
sbryzak: Done

authorization/token/worker/token_cleanup_worker_blackbox_test.go
sbryzak: Good point, I've modified this to use a loop as suggested.  Worker timer I set to 50ms though, 10ms seems a little too aggressive ;)

Issue #778, opened 2019-02-08
[user deactivation] soft-delete user and identity records


xcoulon:
Part of the [User Deactivation Epic](https://github.com/fabric8-services/fabric8-auth/issues/774)

`soft-delete` the user account in `auth` (soft delete for now, so we don't have to deal with FK). This can be revised in the future. 

Having the user information obfuscated (see #775) will not block for further registration of inactive users (i.e., they come back to the platform). But all previous data will have been erased, so it's a new/fresh start.

Issue #774, opened 2019-02-08
[Epic] User deactivation


xcoulon:
Provide a new `/api/namedusers/:username/deactivate` endpoint to:
- [x] https://github.com/fabric8-services/fabric8-auth/issues/775 obfuscate and deactivate the account in `auth` service
- [x] https://github.com/fabric8-services/fabric8-auth/issues/776 revoke all tokens
- [x] https://github.com/fabric8-services/fabric8-auth/issues/777 call other services to deactivate the account
- [x] https://github.com/fabric8-services/fabric8-auth/issues/778 soft-delete the account
- [x] https://github.com/fabric8-services/fabric8-auth/issues/783 unlink the external accounts (GitHub, etc.)
- [x] https://github.com/fabric8-services/fabric8-auth/issues/780 expose endpoint (once everything else above is done, since we don't want to expose an incomplete feature)
- [x] https://github.com/fabric8-services/fabric8-auth/issues/796 track user activity based on token validation and renewal
- [x] ~~https://github.com/fabric8-services/fabric8-auth/issues/791 expose endpoint to list the users to deactivate~~

alexeykazakov:
> * unlink the external accounts? (GitHub, etc.)

Good point. Yes!

xcoulon:
> > * unlink the external accounts? (GitHub, etc.)
> 
> Good point. Yes!

added in https://github.com/fabric8-services/fabric8-auth/issues/783

alexeykazakov:
We also need to make sure that we log all deprovisioned users.

alexeykazakov:
There is one subtask is missing. Before we expose the endpoint we have to change the registration endpoint. It should handle the situation when we get a request to register a new user who was previously deactivated/soft-deleted.

alexeykazakov:
Never mind. With data obfuscation we don't have to do extra in the registration flow :)

alexeykazakov:
This Epic (provide `/api/namedusers/:username/deactivate` endpoint) is done. Closing it.

xcoulon:
reopening as I realized that https://github.com/fabric8-services/fabric8-auth/issues/777 was incomplete (missing calls to Che, Jenkins and Env services).

PR #781, opened 2019-02-08
feat(775): obfuscate and deactive user at service layer

### LINKED ISSUE #775: [user deactivation] obfuscate and deactivate
Part of the [User Deactivation Epic](https://github.com/fabric8-services/fabric8-auth/issues/774)

this is the subtask to obfuscate all user personal information (full name, email, etc.) and mark the account as `active=false` in the database when a deactivation request is received.

Keeping the user data in an obfuscated fashion will avoid any conflict if the user comes back again: the registration app will still be allowed to register with the same username/email address as previously

xcoulon:
Added a new `Active` field on the User type (with an
associated column in the DB with default value 'true')

Obfuscate all personal data in the user and identity records,
and mark the user as `active=false`.

Fixes #775

### DIFF
--- a/authentication/account/service/user.go
+++ b/authentication/account/service/user.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/fabric8-services/fabric8-auth/authentication/account"
+
 	"github.com/fabric8-services/fabric8-auth/application/service"
 	"github.com/fabric8-services/fabric8-auth/application/service/base"
 	servicecontext "github.com/fabric8-services/fabric8-auth/application/service/context"
@@ -13,7 +15,7 @@ import (
 	"github.com/fabric8-services/fabric8-auth/log"
 
 	"github.com/jinzhu/gorm"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 )
 
 // NewUserService creates a new service to manage users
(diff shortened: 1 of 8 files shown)


authentication/account/service/user.go
```diff
@@ -98,6 +100,50 @@ func (s *userServiceImpl) DeprovisionUser(ctx context.Context, username string)
...
+		identity.User.Bio = obfuscatated
+		identity.User.Cluster = obfuscatated
+		identity.User.FeatureLevel = obfuscatated
+		// empty data
+		identity.User.ContextInformation = account.ContextInformation{}
+		err = s.Repositories().Identities().Save(ctx, identity)
+		if err != nil {
+			return err
+		}
+		return s.Repositories().Users().Save(ctx, &identity.User)
```
alexeykazakov: You do obfuscate the data but I don't see you soft delete User & Identity.

Also load after obfuscating and deleting User and Identity from the default IdP we would want to load all other Identities (not in use currently, but still theoretically can be used in the future) and obfuscate&soft-delete them too. But it's low priority and can be handled in a separate issue/PR in the future.

test/account.go
```diff
@@ -312,3 +312,34 @@ func AssertIdentityEqual(t require.TestingT, expected, actual *account.Identity)
...
+		actual.User.Bio,
+		actual.User.Company,
+		actual.User.Email,
+		actual.User.FullName,
+		actual.User.ImageURL,
+		actual.User.URL,
+		actual.User.Cluster,
+		actual.User.FeatureLevel,
+	} {
+		_, err := uuid.FromString(field)
```
alexeykazakov: Nice way of checking the obfuscated data! :)

xcoulon: thanks ;)

authentication/account/service/user.go
xcoulon: soft-delete will be taken care of in https://github.com/fabric8-services/fabric8-auth/issues/778 which is also part of https://github.com/fabric8-services/fabric8-auth/issues/774, but I thought it would be simpler to review the work if it was splitted in small subtasks

alexeykazakov: OK. That looks fine to me.

PR #793, opened 2019-02-22
soft-delete account during user deactivation

### LINKED ISSUE #778: [user deactivation] soft-delete user and identity records

xcoulon:
also include changes in the `testgraph` to be able to retrieve soft-deleted records

fixes #778

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,83 +2,62 @@
 
 
 [[projects]]
-  digest = "1:2aff5edb9bccd2974090fddb17ca7ab05a3f5c983db567c30c7f0b53404f5783"
   name = "github.com/ajg/form"
   packages = ["."]
-  pruneopts = "UT"
   revision = "cc2954064ec9ea8d93917f0f87456e11d7b881ad"
   version = "v1.5"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


Gopkg.lock
```diff
@@ -614,99 +479,33 @@
...
-    "golang.org/x/crypto/bcrypt",
-    "golang.org/x/net/context",
-    "golang.org/x/net/websocket",
-    "golang.org/x/oauth2",
-    "golang.org/x/oauth2/github",
-    "gopkg.in/h2non/gock.v1",
-    "gopkg.in/square/go-jose.v2",
-    "gopkg.in/yaml.v2",
-  ]
+  inputs-digest = "91c2d17f41b78bab590d1ae93d7a2e41171135e7783b7cc6f8ccf914167a17ec"
```
alexeykazakov: Not sure why your dep keep changing this file.. Looks like you randomly use different dep versions ;) But it's not that critical.

xcoulon:
thanks for the review, @sbryzak !

PR #795, opened 2019-02-25
Fix oauth.provider.client.id secret mounting

alexeykazakov:
Currently because of a bug in our template we do not set client ID env var correctly, so we always use the default value which is "fabric8-platform-online". But we have to fix this altogether with updating the corresponding secret when switching to RHD SSO.

**Should be merged only when we update CM/Secrets to switch to RHD!** If we merge it without updating the secrets then incorrect ID will be used.

### DIFF
--- a/openshift/auth.app.yaml
+++ b/openshift/auth.app.yaml
@@ -49,7 +49,7 @@ objects:
               secretKeyRef:
                 name: auth
                 key: db.password
-          - name: AUTH_AUTH_PROVIDER_CLIENT_ID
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/df60aca442c339b32eb0dca8f3b7e61a27b6f030/.ike-prow/test-keeper.yaml).

alexeykazakov:
[test]

PR #798, opened 2019-02-27
Add local dev configuration option for toolchain-operator

dipak-pawar:
This PR adds SA configuration to support local development setup for toolchain-operator


alexeykazakov:
The secrets are in both preview and prod already. We can merge it.

PR #803, opened 2019-03-12
Set default value on all empty `last_activity` records

xcoulon:
Only update null values
Also make sure that new values do not have an empty value from now on.

Fixes #ODC184


xcoulon:
Note: we can't make the column `last_active` as non nullable for now, or at least until the application has been deployed on all pods at least once, without a need to rollback.

PR #806, opened 2019-03-19
list identities to deactivate

xcoulon:
fixes #ODC216

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,83 +2,62 @@
 
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


Gopkg.lock
```diff
@@ -2,83 +2,62 @@
 
 
 [[projects]]
-  digest = "1:2aff5edb9bccd2974090fddb17ca7ab05a3f5c983db567c30c7f0b53404f5783"
```
alexeykazakov: Here we go again :)
Seriously, what's wrong with dep? Is it some issue with different dep versions or a bug in dep?

xcoulon: no idea... I have not changed `dep` on my machine for a while... 🤷‍♂️

PR #807, opened 2019-03-20
reset deactivation notification timestamp on user activity

xcoulon:
done when "touching" the user's "last active" field

fixes #ODC217


xcoulon:
thanks @alexeykazakov 🙌

PR #810, opened 2019-03-29
merge subscription and subscription/service pkg

xcoulon:
also, remove the dummy subscription service impl
and use Gock to mock calls to the Registration App
in tests, instead.

Fixes #ODC386


xcoulon:
thanks for your reviews, @sbryzak and @alexeykazakov 🙌

PR #811, opened 2019-04-01
make sure users were notified 'n' days before deactivation

xcoulon:
this will avoid the case where the user receives a notification
but gets deactivated the next day (if the notification was sent too late)

Fixes #ODC394


authentication/account/service/user_blackbox_test.go
```diff
@@ -291,6 +304,9 @@ func (s *userServiceBlackboxTestSuite) TestListUsersToDeactivate() {
 		config.GetUserDeactivationInactivityPeriodFunc = func() time.Duration {
 			return 30 * 24 * time.Hour // 30 days
 		}
+		config.GetUserDeactivationInactivityNotificationPeriodFunc = func() time.Duration {
+			return 20 * 24 * time.Hour // 20 days
+		}
```
alexeykazakov: Do we have a test for a following flow:
- User A is not active for N days.
- Check if User A is returned to be notified.
- Check if User A is not returned to be deactivated yet.
- User A does some activity.
- Check if User A is not returned to be notified.
- Check if User A is not returned to be deactivated.
- After notification period the user A still not among users to notify or deactivate.

xcoulon: ok, let me add a flow to cover that case, although I believe that my sub tests already cover those cases, but independently

authentication/account/repository/identity.go
```diff
@@ -415,12 +415,13 @@ func (m *GormIdentityRepository) ListIdentitiesToNotifyForDeactivation(ctx conte
...
 // if limit is a negative value (eg: '-1'), it is ignored
-func (m *GormIdentityRepository) ListIdentitiesToDeactivate(ctx context.Context, lastActivity time.Time, limit int) ([]Identity, error) {
+func (m *GormIdentityRepository) ListIdentitiesToDeactivate(ctx context.Context, lastActivity, notification time.Time, limit int) ([]Identity, error) {
 	defer goa.MeasureSince([]string{"goa", "db", "user", "listIdentitiesToDeactivate"}, time.Now())
 	var identities []Identity
 	// sort identities by most inactive and then by date of creation to make sure we always get the same sublist of identities between
 	// queries to notify before deactivation and queries to deactivate for real.
-	err := m.db.Model(&Identity{}).Where("last_active < ? and deactivation_notification is NOT NULL", lastActivity).Order("last_active, created_at").Limit(limit).Find(&identities).Error
+	// Also, make sure that the users to deactivate were notified at least `n` days ealier.
+	err := m.db.Model(&Identity{}).Where("last_active < ? and deactivation_notification < ?", lastActivity, notification).Order("last_active, created_at").Limit(limit).Find(&identities).Error
```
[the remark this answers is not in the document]
xcoulon: yes, that user's case covers it: https://github.com/xcoulon/fabric8-auth/blob/IssueODC394_deactivate_n_days_after_notification/authentication/account/service/user_blackbox_test.go#L233 -> the user has been inactive for a long-enough period, but since she did not get the notification, she is not listed in the users to deactivate.

authentication/account/service/user.go
```diff
@@ -181,9 +181,10 @@ func (s *userServiceImpl) NotifyIdentitiesBeforeDeactivation(ctx context.Context
 
 // ListIdentitiesToDeactivate lists the identities to deactivate
 func (s *userServiceImpl) ListIdentitiesToDeactivate(ctx context.Context) ([]repository.Identity, error) {
-	since := time.Now().Add(-s.config.GetUserDeactivationInactivityPeriod()) // remove 'n' days from now (default: 31)
+	since := time.Now().Add(-s.config.GetUserDeactivationInactivityPeriod())                                                                    // remove 'n' days from now (default: 31)
```
sbryzak: Would it be possible to include the unit type in the configuration parameter name, in order to reduce confusion about what kind of value to configure here?  For example, `GetUserDeactivationInactivityPeriodDays`

authentication/account/service/user_blackbox_test.go
sbryzak: I had a related question.. is the `deactivation_notification` value reset to null when there is user activity?

authentication/account/service/user.go
xcoulon: sure, that's a good idea, let me do that.

xcoulon:
@sbryzak to answer your question above (not sure why I can't reply directly after your comment, but 🤷‍♂️): yes, the `deactivation_notification` column is set back to `NULL` when the user is active again: https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/repository/identity.go#L888

authentication/account/service/user.go
xcoulon: done in a8be694

authentication/account/service/user_blackbox_test.go
xcoulon: @alexeykazakov test added in 9c3c1aa

PR #812, opened 2019-04-04
worker routine to notify inactive users

xcoulon:
using the `pglock` library to obtain a lock in the
database, in order to avoid concurrent processing (1 per pod)
lock is acquired at the worker init and released at the call
to its `Stop()` method which MUST be within a `deferred` call
in the `main.go`, so that the pod that owns the lock can release
it when it is shutdown (ie, during a deployment rollout)

Fixes #ODC146


worker/repository/worker_lock.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// AcquireLockToNotifyUsersToDeactivate acquires a lock on the postgres table for the requesting
+// user deactivation notification worker
+// Returns an error if the lock could not be obtained
+func (r *lockRepositoryImpl) AcquireLockToNotifyUsersToDeactivate(ctx context.Context) (*pglock.Lock, error) {
+	return r.acquireLock(ctx, "user_deactivation_notification_worker_lock")
+}
+
+// AcquireLockToNotifyUsersToDeactivate acquires a lock on the postgres table for the requesting
```
alexeykazakov:
```suggestion
// acquireLock acquires a lock on the postgres table for the requesting
```

worker/repository/worker_lock.go
```diff
@@ -0,0 +1,96 @@
...
+// user deactivation notification worker
```
alexeykazakov:
```suggestion
// user deactivation or deactivation notification worker
```

authentication/account/worker/user_deactivation_notification_worker.go
```diff
@@ -0,0 +1,116 @@
...
+	log.Info(w.ctx, map[string]interface{}{
+		"owner": w.owner,
+	}, "starting user deactivation notification worker")
+
+	l, err := w.app.WorkerLockRepository().AcquireLockToNotifyUsersToDeactivate(w.ctx)
+	if err != nil {
+		log.Warn(w.ctx, map[string]interface{}{
+			"error": err,
+			"owner": w.owner,
+		}, "unable to acquire user deactivation notification worker lock")
```
alexeykazakov:
```suggestion
		}, "unable to acquire user deactivation notification worker lock which is OK if another pod has already started a worker")
```

authentication/account/worker/user_deactivation_notification_worker.go
```diff
@@ -0,0 +1,116 @@
...
+	}, "stopping user deactivation notification worker")
+	w.ticker.Stop()
+	if w.lock != nil {
+		err := w.lock.Close()
+		if err != nil {
+			log.Error(w.ctx, map[string]interface{}{
+				"err": err,
+			}, "error while releasing user deactivation notification worker lock")
+		}
+		log.Info(w.ctx, map[string]interface{}{
```
alexeykazakov: Should be in `} else {...log.Info(...)...}`

authentication/account/worker/user_deactivation_notification_worker.go
```diff
@@ -0,0 +1,116 @@
...
+	ctx    context.Context
+	app    application.Application
+	lock   *pglock.Lock
+	ticker *time.Ticker
+	stopCh chan bool
+	owner  string
+}
+
+// Start starts the worker with the given timer
+func (w *userDeactivationNotificationWorker) Start(freq time.Duration) {
```
alexeykazakov: We can't guarantee that Start() will be always called. So, we have to come up with some health check. For example we could write a timestamp of the last worker cycle and check it when starting a worker. But we should also take into account rolling updates.
Take a look at our tenant update mechanism which deals with the same problem. @MatousJobanek can help you with that.

authentication/account/worker/user_deactivation_notification_worker.go
```diff
@@ -0,0 +1,116 @@
...
+func (w *userDeactivationNotificationWorker) Stop() {
+	if w.stopCh != nil {
+		log.Info(w.ctx, map[string]interface{}{
+			"owner": w.owner,
+		}, "time to stop the user deactivation notification worker")
+		w.stopCh <- true
+	}
+}
+
+func (w *userDeactivationNotificationWorker) stop() {
```
sbryzak: Having two identically named functions is confusing... could we call this one `cleanup()` or something along those lines instead?

authentication/account/worker/user_deactivation_notification_worker.go
sbryzak: Where is the worker started from, and what is the default value for `freq` ?

authentication/account/worker/user_deactivation_notification_worker.go
xcoulon: sure, I could rename the `stop` func if it clarifies.

authentication/account/worker/user_deactivation_notification_worker.go
xcoulon: the `Start(time.Duration)` func must be called in `main.go` (not implemented yet, actually)

xcoulon: @alexeykazakov the `Start(time.Duration)` func will be called in the `main.go` during the bootstrap, so I'm not sure why you say that we can't guarantee it will always be called. Could you elaborate, please?
More importantly, we must ensure that the lock will be released during shutdown, ie, we must have a "graceful shutdown" function to handle that..

authentication/account/worker/user_deactivation_notification_worker.go
xcoulon: done in 7774b95

authentication/account/worker/user_deactivation_notification_worker.go
xcoulon: Good point. Fixed in 7774b95

authentication/account/worker/user_deactivation_notification_worker.go
alexeykazakov: Oh, did I say Start()? :)
I meant Stop() of course! :)
We can’t guarantee it’s always called. Even defer is not 100% guaranty.

xcoulon: well, I hope that a graceful shutdown hook can at least release the lock in a proper manner, which would work for the rolling updates. But that would not work for the case when the process is "hard killed", indeed.

authentication/account/worker/user_deactivation_notification_worker_test.go
```diff
@@ -0,0 +1,127 @@
...
+		time.Sleep(time.Millisecond * 100)
+		// then load the user and check her deactivation notification status
+		result, err := s.Application.Identities().Load(context.Background(), identity1.ID)
+		require.NoError(s.T(), err)
+		assert.NotNil(s.T(), result.DeactivationNotification)
+		// notification only sent once to the user
+		assert.Equal(s.T(), uint64(1), notificationServiceMock.SendMessageAsyncCounter)
+	})
+
+	s.Run("multiple workers but only one working", func() {
```
MatousJobanek: I think that you need to put it into a separated test (not subtest) to make the suite passing - the first subtest deactivates the user, so there is nothing left for the second one.

authentication/account/worker/user_deactivation_notification_worker_test.go
```diff
@@ -0,0 +1,127 @@
...
+		w2.Start(freq)
+		w3.Start(freq)
+		w4.Start(freq)
+		// wait a few cycles before checking the results
+		time.Sleep(freq * 5)
+		w1.Stop()
+		w2.Stop()
+		w3.Stop()
+		w4.Stop()
+		time.Sleep(freq * 10) // give workers some time to stop for good
```
MatousJobanek: for this type of test, I would use approach using a latch - something like this:
```go
                var starter sync.WaitGroup
		starter.Add(1)
		var workers []worker.UserDeactivationNotificationWorker
		freq := time.Millisecond * 50
		for i := 0; i < 100; i++ {
			w := s.newUserDeactivationNotificationWorker(fmt.Sprintf("pod-%d", i), app)
			workers = append(workers, w)

			go func(w worker.UserDeactivationNotificationWorker) {
				starter.Wait()
				w.Start(freq)
			}(w)
		}
		starter.Done()
		// wait a few cycles before checking the results
		time.Sleep(freq * 5)
		for _, w := range workers {
			w.Stop()
		}
		time.Sleep(freq * 10) // give workers some time to stop for good
```

authentication/account/worker/user_deactivation_notification_worker.go
MatousJobanek: I guess that this should be handled by `pglock` library - the [heartbeat frequency](https://github.com/cirello-io/pglock#send-automatic-heartbeats) should be exactly for this, right? However, the question is if it is handled correctly. 

In addition, could it happen that when the pod that is running the worker is killed, and a new one is started quicker than the lock is released? This scenario could be actually solved by a `time.Sleep(lease_duration)` at the beginning of the goroutine that starts the worker. Or, more gracefully,  sleep for that time when the acquiring of the lock fails and then try it one more time. If the second attempt fails as well, then stop trying it.

nurali-techie:
@xcoulon I have two related things.

**First**, on freq [link](https://github.com/fabric8-services/fabric8-auth/pull/812/files#diff-7ddfb3e035b42cd70649cc33393fe32cR303)
Does this mean that every 24 hrs from pod_start the worker will do the work ??

For ex, assuming pod started and got lock at 01-Jan 3.00 PM
Run1 - 01-Jan 3.00 PM
Run2 - 02-Jan 3.00 PM

If above is the case then I would rather prefer to run at some particular time during day.  For ex, every day mid-night.

For same ex, assuming pod started and got lock at 01-Jan 3.00 PM
Run1 - 02-Jan 00.00 AM
Run2 - 03-Jan 00.00 AM

This will give benefit during debugging and trouble shooting.  Like whenever we have issue with these workers, we always need to check logs of around mid night time.

**Second**, I understood that when pod start it tried to get lock, assume it's successful.  Now when that pod shutdown the pod MUST release the lock.  Here, graceful_shutdown become very important. [link](https://github.com/fabric8-services/fabric8-auth/pull/812/files#diff-7ddfb3e035b42cd70649cc33393fe32cR306).  I also see that @MatousJobanek has mentioned about one edge case where he is giving suggestion to use `lease_duration`.

Here, I would prefer that each time when pod about to start work, all running pod should compete for lock and one of the running pod will get lock.  That pod will run the work and after completing the work it will immediately release the lock.  With this, we no need to bother about graceful_shutdown of pod and related scenario.

**Summary**, here First and Second go hand in hand to make sure that all running pod complete for lock at same time (for ex at mid-night daily) and one of the pod get the lock to run the work for that particular instance.

authentication/account/worker/user_deactivation_notification_worker.go
alexeykazakov: This is a good point @MatousJobanek.
Heartbeat thing looks exactly what we need here. Just let make sure we use it correctly with all timeouts set correctly, etc.

> Or, more gracefully, sleep for that time when the acquiring of the lock fails and then try it one more time. If the second attempt fails as well, then stop trying it.

+1. Just need to make sure that the sleep is longer than the Heartbeat frequency.

alexeykazakov:
@nurali-techie @xcoulon not sure what are you referring to when talking about midnight or other particular time of the day. Midnight for whom? What time zone? :)

main.go
```diff
@@ -291,11 +293,19 @@ func main() {
...
-	// Stop the worker when the app shuts down
-	defer worker.Stop()
+	// token cleanup, running once every hour
+	tokenCleanupWorker := tokenworker.NewTokenCleanupWorker(context.Background(), appDB)
+	tokenCleanupWorker.Start(time.Hour)
+	// user deactivation and notification workers, running once per day
+	ctx := manager.ContextWithTokenManager(context.Background(), tokenManager)
+	ctx = context.WithValue(worker.LockOwner, config.GetPodName())
+	userDeactivationWorker := userworker.NewUserDeactivationWorker(ctx, appDB)
+	userDeactivationWorker.Start(time.Hour * 24)
```
alexeykazakov: The `defaultUserDeactivationFetchLimit` is set to 10. It means that we will deactivate 10 users every 24 hours. It's not even nearly enough :)

We don't have any delays between deactivation calls for particular users within the cycle, do we?
If so then we don't really want to raise the defaultUserDeactivationFetchLimit to some very big number since it may overload the other services.
So, let's increase the cycle frequency instead then. Let's say 1 minute. So, we will be deactivating up to 10 users per minute or 14400 per 24 hours. It should be good enough.

Also let's put this to configuration. We don't need any config maps for that yet. It's just matter of having all these numbers in one place (default config) so it's easy to track/change them. With some good comment, so, it's clear how these numbers affects the cycle.

main.go
```diff
@@ -291,11 +293,19 @@ func main() {
...
-	worker := worker.NewTokenCleanupWorker(context.Background(), appDB)
-	// Activate token cleanup once every hour
-	worker.Start(time.NewTicker(time.Hour))
-	// Stop the worker when the app shuts down
-	defer worker.Stop()
+	// token cleanup, running once every hour
+	tokenCleanupWorker := tokenworker.NewTokenCleanupWorker(context.Background(), appDB)
+	tokenCleanupWorker.Start(time.Hour)
+	// user deactivation and notification workers, running once per day
+	ctx := manager.ContextWithTokenManager(context.Background(), tokenManager)
```
golangcibot: ineffectual assignment to `ctx` (from `ineffassign`)

nurali-techie:
> @nurali-techie @xcoulon not sure what are you referring to when talking about midnight or other particular time of the day. Midnight for whom? What time zone? :)

@alexeykazakov the solution has nothing to do with time_zone.  The solution is more about running work on **same_time for all run irrelevant of time_zone**.

In case of schedule with 24-hrs, we mentioned to run at mid_night and for completeness we can say mid_night of pod/cluster time_zone but it can be any preferred time (with time_zone) we decide.

Take another ex, we also have schedule with every_hour for cleanup_worker [link](https://github.com/fabric8-services/fabric8-auth/pull/812/files#diff-7ddfb3e035b42cd70649cc33393fe32cR298), for this we should try to run work at 1PM, 2PM, 3PM likewise rather than 1:10:30 PM, 2:10:30 PM, 3:10:30 PM with assumption that pod started and acquire lock at 1:10:30 PM.

nurali-techie:
> I'll see if there's something closer to cron scheduling instead of a basic timer.

@xcoulon we don't need cron scheduling mechanism.  We just need simple logic for one or two schedule.  I just try this to get next schedule for 24-hrs scheduling.

Code: https://play.golang.org/p/MycTNawM6wo
Output on my system:
```
nowTime=2019-04-08 10:31:52.23250857 +0530 IST m=+0.000130172
todayRoundTime=2019-04-08 05:30:00 +0530 IST
tomorrowRoundTime=2019-04-09 05:30:00 +0530 IST
```
This is for 24-hrs schedule.
For every_hour schedule, change d to `d := time.Hour` and it will give next_hour worker run time.

alexeykazakov:
@nurali-techie see https://github.com/fabric8-services/fabric8-auth/pull/812#discussion_r272703493

We must be ready to deactivate 10000+ users per day. It's a full time job. It doesn't really matter when we start. So, let's not bother with any scheduling. Start the cycle as soon as it's ready.

xcoulon:
@nurali-techie I changed the logic wrt to lock acquisition and release for every cycle in e7b23af

main.go
xcoulon: @alexeykazakov those are default settings, to avoid running on the whole set of users unless explicitly configured. We'll increase the limit and frequency once we're confident with the whole workflow, of course ;)
We already have a "wait period" between each deactivation: https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L154-L155

But I'm fine with increasing the frequency to 1min by default

xcoulon: @alexeykazakov timer set to 1min for both workers in 80d413e

main.go
xcoulon: fixed in 1befb6d

authentication/account/worker/user_deactivation_notification_worker.go
xcoulon: ok, so with https://github.com/fabric8-services/fabric8-auth/pull/812/commits/e7b23af3ca548b4a2b2c7a936c2951950d2d425a, I assume that things are a little simpler: if a worker cannot acquire the lock, then it will try again, at the next cycle. The question remains for the case where the pod is killed while the worker acquired the lock and is still working. I'll see how the heartbeat can help in that case (with some tests, if possible)

authentication/account/worker/user_deactivation_notification_worker_test.go
xcoulon: @MatousJobanek done in 84bd48f

xcoulon: and thanks for the suggestion 🙌

xcoulon:
@alexeykazakov as you suggested, I disabled the workers in `main.go` in https://github.com/fabric8-services/fabric8-auth/pull/812/commits/d37f51bddd75a082990a43091cd4fec935b10c47

main.go
alexeykazakov: > We already have a "wait period" between each deactivation: https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L154-L155

This is for user notifications. Not deactivation. But that's fine.

BTW there is some inconsistency between deactivation and notification workers and services. For notifications we iterate users in the user service https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L122 And in the worker just calls that service.
But for deactivation we iterate users in the worker (not in the service). And in the service we have two functions only: Deactivate() a user and GetList().

It would be more consistent if ether:
Move NotifyIdentitiesBeforeDeactivation() to a worker in keep Notify() and GetListToNotify() in the service.
OR
Create DeactivateUsers() in the user service and use it from the worker.

But it's not critical :) I'm OK to keep it as it is.

alexeykazakov:
@xcoulon also @sbryzak has reviewed https://github.com/cirello-io/pglock and approved it for using :) The hearbeat/lease logic looks pretty robust and should guarantee proper locking and unlocking in case of hard crash. So, your PR looks good in term of locking/unlocking.
We might want to relax the lease and hearbeat timeouts a bit though. 3s and 1s looks a bit too aggressive but it's not critical. Maybe 3 and 1 minutes? But it's minor.

configuration/defaults.go
```diff
@@ -122,4 +122,8 @@ vwIDAQAB
 	defaultUserDeactivationInactivityPeriod = defaultUserDeactivationInactivityNotificationPeriod + 7 // 31 days (7 days after defaultUserDeactivationInactivityNotificationPeriod)
 	// defaultPostDeactivationNotificationDelay the default number of milliseconds between to user account deactivation notifications by the same go routine
 	defaultPostDeactivationNotificationDelay = 500
+	// defaultPostDeactivationNotificationDelay the default interval between 2 cycles of the user deactivation worker
+	defaultUserDeactivationWorkerIntervalMinutes = 1
```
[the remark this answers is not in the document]
xcoulon: I'm fine with increasing the interval to 4mins. If more than 1 pod can get the lock in a cycle (assuming their timer ticks at a slightly different time), then it's still fine, indeed. As long as the processing is sequential, we're good ;)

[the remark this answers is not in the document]
xcoulon: done in 312c2bd

nurali-techie:
@xcoulon one feedback (optional idea), in case it relevant and helps.

**Topic: release lock acquired by dead_pods:**

Assume this is sequence of events.
T1 - pod1 acquired lock
T2 - pod1 crashed
T3 - pod2 try to acquire lock

For above scenario, pod1 can be considered as dead_pod.  If we have below mechanism either we can code or from pglock or from combination of both.

At T3, pod2 try to acuqire lock
```
try (acuire_lock by pod2) {
	if not locked {
		give lock to pod2 and return
	}

	if locked {
		live_pods = get list of live pods name
		for each live_pods {
			if live_pod.name == lock.owner_name {
				reject lock by pod2 and return
			}
		}
		// it means lock owner is dead_pod
		release lock
		give lock to pod2 and return
	}
}
```
I am not sure, if we can get list of live_pods (status=running pods for auth_service in openshift cluster).  But if we can achieve above then our solution will become robust in case of lock acquired by dead_pods.  And there are many edge cases where this will be applicable.

main.go
xcoulon: ok, let me refactor this code. The aforementioned `wait period` was supposed to be used for the deactivations, not the notifications.
I suggest I address those changes in another PR to keep this one focused on the workers.

alexeykazakov:
> Topic: release lock acquired by dead_pods:

@nurali-techie this is already handled properly by pglock.
If Pod1 crashed without explicitly releasing the lock the heartbeat routine form that pod will not updating the lease anymore and Pod2 will successfully grab the lock after the previous lock owner (Pod1) lease ended.

xcoulon:
@alexeykazakov I've update the PR in b527a98 to call the OSO Reg App to deactivate the users. I suggest I take care of refactoring the code in the user service to expose a single `DeactivateUsers(context.Context, *time.Time)` method in a subsequent PR, as this is more of a  cosmetic concern (ie, the business logic is covered)

xcoulon:
[test]

PR #813, opened 2019-04-09
Delete user on Che

xcoulon:
Call Che to delete user account, but do not delete tenants if something wrong happened on Che side.

Fixes #ODC140


configuration/configuration.go
```diff
@@ -120,8 +120,9 @@ const (
 	//
 	//------------------------------------------------------------------------------------------------------------------
 
-	varAccessTokenExpiresIn  = "useraccount.token.access.expiresin"  // In seconds
-	varRefreshTokenExpiresIn = "useraccount.token.refresh.expiresin" // In seconds
+	varAccessTokenExpiresIn    = "useraccount.token.access.expiresin"    // In seconds
```
xcoulon: or `...expires.in`?

authorization/token/manager/token_manager.go
```diff
@@ -513,6 +514,26 @@ func (m *tokenManager) GenerateUnsignedUserAccessTokenForIdentity(ctx context.Co
 	return token, nil
 }
 
+// GenerateTransientUserAccessTokenForIdentity generates a transient user access token, an extremely short-lived token
+func (m *tokenManager) GenerateTransientUserAccessTokenForIdentity(ctx context.Context, identity repository.Identity) (*string, error) {
```
xcoulon: what about using the term `action token` here?

che/service/che_service.go
```diff
@@ -1,30 +1,45 @@
...
+	manager.TokenManagerConfiguration
 	GetCheServiceURL() string
 }
+
 // NewCheService creates a new Che service.
-func NewCheService(config Configuration) service.CheService {
+func NewCheService(context servicecontext.ServiceContext, config Configuration) service.CheService {
+	tokenManager, err := manager.NewTokenManager(config)
+	if err != nil {
+		log.Panic(nil, map[string]interface{}{
```
xcoulon: why `Panic` here?

configuration/configuration.go
sbryzak: I haven't changed this variable, it's only showing here because of reformatting

che/service/che_service.go
sbryzak: Good catch, this was copied straight from the token service.. i'll change it to something more sensible

authorization/token/manager/token_manager.go
sbryzak: See my reply in Slack ;)

che/service/che_service.go
sbryzak: I've removed the tokenManager member from che_service, and instead exposed it in TokenService so that we can access the token manager functions from anywhere that has access to services.

che/service/che_service.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return errs.Wrapf(err, "unable to delete user '%s' in Che", identityID.String())
+	}
+
+	identity, err := s.Repositories().Identities().Load(ctx, identityID)
+	if err != nil {
+		return errs.Wrapf(err, "unable to delete user '%s' in Che", identityID.String())
+	}
+
+	token, err := s.Services().TokenService().TokenManager().GenerateTransientUserAccessTokenForIdentity(ctx, *identity)
```
alexeykazakov: Or you can always use the default token manager: https://github.com/fabric8-services/fabric8-auth/blob/master/authorization/token/manager/token_manager.go#L46

sbryzak: Yes, I prefer that much better.. updated!

che/service/che_service.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return errs.Wrapf(err, "unable to delete user '%s' in Che", identity.ID.String())
+	}
+
+	token, err := tokenManager.GenerateTransientUserAccessTokenForIdentity(ctx, identity)
+	if err != nil {
+		return errs.Wrapf(err, "unable to delete user '%s' in Che", identity.ID.String())
+	}
+
+	_, err = s.Services().TokenService().RegisterToken(ctx, identity.ID, *token, token2.TOKEN_TYPE_ACCESS, nil)
```
xcoulon: I thought the point of having a transient access token was because we did not want to register it? Or is it mandatory for the token validation? If so, could we have an extra claim field to flag the token as transient and skip the database validation?

che/service/che_service_blackbox_test.go
```diff
@@ -0,0 +1,97 @@
...
+		return s.Configuration.GetAccessTokenExpiresIn() // reuse value from default config
+	}
+	config.GetTransientTokenExpiresInFunc = func() int64 {
+		return s.Configuration.GetTransientTokenExpiresIn() // reuse value from default config
+	}
+	svc := cheservice.NewCheService(svcCtx, config)
+	s.Run("ok", func() {
+		// given
+		identity := s.Graph.CreateIdentity().Identity()
+		tokenManager, err := manager.DefaultManager(config)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

che/service/che_service.go
alexeykazakov: Yes, we have to handle it but I don't see it in this PR.
We either should skip the registration and validation completely or keep registration/validation but skip touching user activity field during validation.

che/service/che_service_blackbox_test.go
xcoulon: fixed in https://github.com/fabric8-services/fabric8-auth/pull/813/commits/5f017f9d2a59068472bc2c7999cf3e4982725860

che/service/che_service.go
sbryzak: We already have a specified claim field for transient tokens ("transient"=true) however Alexey has made a good point about validation, this is something I overlooked so I'll address it ASAP.

sbryzak: I've added a check to the `ValidateToken()` to skip updating the identity's last active timestamp if the token is transient.  There's still a failure in the user_blackbox_test though, for which I'm not sure how to fix yet.

sbryzak: Strange.. the tests pass in CI even though they fail for me locally.  Not sure what to make of that, @xcoulon can you try running the tests locally?

sbryzak: Ignore my previous comment, after cleaning up the database the tests now pass successfully.  It seems that Alexey is still experiencing a test failure though, so my point still stands @xcoulon that it would be interesting to see what result you get.

xcoulon: @sbryzak see https://github.com/fabric8-services/fabric8-auth/pull/816 as a separate PR to address this problem (as I could reproduce it locally on `master` branch as well)

xcoulon:
[test]

xcoulon:
[test]

PR #815, opened 2019-04-11
fix expiry date field and check notification msg in tests

xcoulon:
Fixes #ODC507 

### DIFF
--- a/authentication/account/service/user.go
+++ b/authentication/account/service/user.go
@@ -129,9 +129,7 @@ func (s *userServiceImpl) NotifyIdentitiesBeforeDeactivation(ctx context.Context
 	// for each identity, send a notification and record the timestamp in a separate transaction.
 	// perform the task for each identity in a separate Tx, and just log the error if something wrong happened,
 	// but don't stop processing on the rest of the accounts.
-	expirationDate := now().
-		Add(s.config.GetUserDeactivationInactivityPeriodDays() - s.config.GetUserDeactivationInactivityNotificationPeriodDays()).
-		Format("Mon Jan 2")
+	expirationDate := GetExpiryDate(s.config, now)
 	// run the notification/record update in a separate routine, with pooling of child routines to avoid
 	// sending too many requests at once to the notification service and to the database
 	defer ants.Release()
@@ -177,6 +175,15 @@ func (s *userServiceImpl) NotifyIdentitiesBeforeDeactivation(ctx context.Context
 	}
 	wg.Wait()
 	return identities, nil
+}
+
+// GetExpiryDate a utility function which returns the expiry date, ie, when the user deactivation will happen
+// The date is based on the given 'now', and takes into account the delay for which the user is given a chance
+// to come back (7 days by default)
+func GetExpiryDate(config UserServiceConfiguration, now func() time.Time) string {
+	return now().
+		Add(config.GetUserDeactivationInactivityPeriodDays() - config.GetUserDeactivationInactivityNotificationPeriodDays()).
+		Format("Mon Jan 2")
 }
 
 // ListIdentitiesToDeactivate lists the identities to deactivate
--- a/authentication/account/service/user.go
+++ b/authentication/account/service/user.go
@@ -188,7 +188,7 @@ func (s *userServiceImpl) ListIdentitiesToDeactivate(ctx context.Context, now fu
 }
 
 func (s *userServiceImpl) notifyIdentityBeforeDeactivation(ctx context.Context, identity repository.Identity, expirationDate string, now func() time.Time) error {
-	msg := notification.NewUserDeactivationEmail(identity.ID.String(), identity.Username, expirationDate)
+	msg := notification.NewUserDeactivationEmail(identity.ID.String(), identity.User.Email, expirationDate)
 	_, err := s.Services().NotificationService().SendMessageAsync(ctx, msg)
 	if err != nil {
 		return errs.Wrap(err, "failed to send notification to user before account deactivation")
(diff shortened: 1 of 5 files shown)


authentication/account/service/user_blackbox_test.go
```diff
@@ -140,6 +143,11 @@ func (s *userServiceBlackboxTestSuite) TestListUsersToNotifyBeforeDeactivation()
 		require.NoError(s.T(), err)
 		require.NotNil(s.T(), identity.DeactivationNotification)
 		assert.True(s.T(), time.Now().Sub(*identity.DeactivationNotification) < time.Second*2)
+		// also verify that the message to send to the user has the correct data
+		assert.Equal(s.T(), identity2.ID.String(), msgToSend.TargetID)
+		expiryDate := time.Now().Add(config.GetUserDeactivationInactivityPeriodDays() - config.GetUserDeactivationInactivityNotificationPeriodDays()).Format("Mon Jan 2")
```
nurali-techie: finding expected value using time.Now() and actual value also use time.Now() will fail if test has started about `11:59:59 PM` and end about `12:00:00 AM`.  In this case there is chances that the test will fail because two different days will be compared.  I am talking about very rare possibility.  The point is, calling time.Now() two times is not right way to compare actual and expected values.

nurali-techie: I would really like if this logic extracted somewhere, not only for test but for actual code.
```
		expiryDate := time.Now().Add(config.GetUserDeactivationInactivityPeriodDays() - config.GetUserDeactivationInactivityNotificationPeriodDays()).Format("Mon Jan 2")
```
One helper function, which takes required input and return expiryDate (mostly without format) and then caller do the needful.  And then this helper func can be used everywhere.

authentication/account/service/user_blackbox_test.go
```diff
@@ -177,8 +185,10 @@ func (s *userServiceBlackboxTestSuite) TestListUsersToNotifyBeforeDeactivation()
 		config.GetUserDeactivationInactivityNotificationPeriodDaysFunc = func() time.Duration {
 			return 30 * 24 * time.Hour // 30 days
 		}
+		var msgToSend []notification.Message
 		notificationServiceMock := servicemock.NewNotificationServiceMock(s.T())
 		notificationServiceMock.SendMessageAsyncFunc = func(ctx context.Context, msg notification.Message, options ...rest.HTTPClientOption) (r chan error, r1 error) {
+			msgToSend = append(msgToSend, msg)
```
nurali-techie: cI see the possibility that the msg come out of order.  As I understood, this is the [code](https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L147) which will call this `SendMessageAsyncFunc` and things run parallel due to this [code](https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L139)

Note: I am not totally sure about above thing.  I am adding this comment so that you can also think if that is possible or not.

nurali-techie:
@xcoulon looks good, I have added few comments just for taking your input on the same :+1:

authentication/account/service/user_blackbox_test.go
xcoulon: done in 8febf38

xcoulon: ok, let me address that

authentication/account/service/user_blackbox_test.go
xcoulon: done in 4893646

authentication/account/service/user_blackbox_test.go
xcoulon: done in 4893646

xcoulon:
@nurali-techie I addressed your comments. Could you review the PR again, please?

authentication/account/service/user_blackbox_test.go
```diff
@@ -177,26 +188,44 @@ func (s *userServiceBlackboxTestSuite) TestListUsersToNotifyBeforeDeactivation()
...
-		for _, id := range []uuid.UUID{identity1.ID, identity2.ID} {
+		expiryDate := userservice.GetExpiryDate(config, nowf)
+		customs := []map[string]interface{}{}
+		for i, id := range []uuid.UUID{identity1.ID, identity2.ID} {
 			identity, err := s.Application.Identities().Load(ctx, id)
 			require.NoError(s.T(), err)
 			require.NotNil(s.T(), identity.DeactivationNotification)
 			assert.True(s.T(), time.Now().Sub(*identity.DeactivationNotification) < time.Second*2)
+			// also verify that the message to send to the user has the correct data
+			assert.Equal(s.T(), identity.ID.String(), msgToSend[i].TargetID)
```
nurali-techie: if we can't guarantee order then we can't make this assertion too.
`assert.Equal(s.T(), identity.ID.String(), msgToSend[i].TargetID)`
Because msg can be added any order to msgToSend.

authentication/account/service/user.go
```diff
@@ -179,6 +177,15 @@ func (s *userServiceImpl) NotifyIdentitiesBeforeDeactivation(ctx context.Context
...
 }
 
+// GetExpiryDate a utility function which returns the expiry date, ie, when the user deactivation will happen
+// The date is based on the given 'now', and takes into account the delay for which the user is given a chance
+// to come back (7 days by default)
+func GetExpiryDate(config UserServiceConfiguration, now func() time.Time) string {
+	return now().
+		Add(config.GetUserDeactivationInactivityPeriodDays() - config.GetUserDeactivationInactivityNotificationPeriodDays()).
+		Format("Mon Jan 2")
+}
```
nurali-techie: @xcoulon thanks for this refactoring :+1:

xcoulon:
thanks @nurali-techie 🙌

PR #816, opened 2019-04-11
Fix TestAuthenticationProviderService/TestInvalidOAuthAuthorizationCode test failure

xcoulon:
"escape" the error message which is returned in the query param of the redirect URL


xcoulon:
thanks @alexeykazakov !

PR #817, opened 2019-04-12
add env for Che in deployment template

xcoulon:
include required env var for the Che service URL in the deployment template.


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/78c7e8e30b87f57b13cf1ed7790890c5a63a311f/.ike-prow/test-keeper.yaml).

openshift/auth.app.yaml
```diff
@@ -169,6 +169,11 @@ objects:
               configMapKeyRef:
                 name: auth
                 key: notification.serviceurl
+          - name: AUTH_CHE_SERVICEURL
+            valueFrom:
+              secretKeyRef:
```
alexeykazakov: Don’t have to be a secret. Use config map instead.

Also we need to add it to https://github.com/fabric8-services/fabric8-auth/blob/master/openshift/auth.config.yaml (just as example for documentation porpoises).

And please add a check for this URL in configuration validation. Similar to https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration.go#L334

The validation will prevent deployments when the ConfigMap is not set.

We also have to create the corresponding CM in preview and prod.

xcoulon: > We also have to create the corresponding CM in preview and prod. 
that's already in progress

> And please add a check for this URL in configuration validation. Similar to https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration.go#L334
hum, so having default values just for testing purpose is not a valid solution, then?

xcoulon: @alexeykazakov I added the `che.serviceurl` CM entry in https://github.com/fabric8-services/fabric8-auth/blob/master/openshift/auth.config.yaml in b0bfeac

alexeykazakov:
[test]

PR #818, opened 2019-04-15
Temporary config change / enable workers for testing

sbryzak:
https://jira.coreos.com/browse/ODC-146

This PR enables step 2 of the testing plan described in the above issue.

Still to-do: limit the user deactivation list to a pre-selected list of e-mail addresses.

### DIFF
--- a/.make/test.mk
+++ b/.make/test.mk
@@ -167,7 +167,7 @@ test-integration-with-coverage: prebuild-check clean-coverage-integration migrat
 test-integration: prebuild-check migrate-database $(SOURCES)
 	$(call log-info,"Running test: $@")
 	$(eval TEST_PACKAGES:=$(shell go list ./... | grep -v $(ALL_PKGS_EXCLUDE_PATTERN)))
-	AUTH_DEVELOPER_MODE_ENABLED=1 AUTH_RESOURCE_DATABASE=1 AUTH_RESOURCE_UNIT_TEST=0 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test $(GO_TEST_INTEGRATION_FLAG) -vet off $(GO_TEST_VERBOSITY_FLAG) $(TEST_PACKAGES)
+	AUTH_DEVELOPER_MODE_ENABLED=1 AUTH_USER_DEACTIVATION_TESTING_MODE=FALSE AUTH_RESOURCE_DATABASE=1 AUTH_RESOURCE_UNIT_TEST=0 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test $(GO_TEST_INTEGRATION_FLAG) -vet off $(GO_TEST_VERBOSITY_FLAG) $(TEST_PACKAGES)
 
 test-integration-benchmark: prebuild-check migrate-database $(SOURCES)
 	$(call log-info,"Running benchmarks: $@")
--- a/configuration/defaults.go
+++ b/configuration/defaults.go
@@ -118,11 +118,9 @@ vwIDAQAB
 	// defaultUserDeactivationFetchLimit the default number of accounts to return when fetching the users to deactivate (and deactivate at once)
 	defaultUserDeactivationFetchLimit = 10
 	// defaultUserDeactivationInactivityPeriod the default number of days of inactivity before a user account can be deactivated
-	//defaultUserDeactivationInactivityNotificationPeriod = 24 // 24 days
-	defaultUserDeactivationInactivityNotificationPeriod = 1 / 48 // 30 minutes
+	defaultUserDeactivationInactivityNotificationPeriod = 24 // 24 days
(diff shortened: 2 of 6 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @sbryzak for this contribution!



Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/b9fe21eb021153a2bbe97c2efa2cf0d2f76646f6/.ike-prow/test-keeper.yaml).

sbryzak:
@nurali-techie could you provide a list of e-mail addresses you'd like to limit the deactivation testing to?

configuration/defaults.go
```diff
@@ -118,9 +118,11 @@ vwIDAQAB
 	// defaultUserDeactivationFetchLimit the default number of accounts to return when fetching the users to deactivate (and deactivate at once)
 	defaultUserDeactivationFetchLimit = 10
 	// defaultUserDeactivationInactivityPeriod the default number of days of inactivity before a user account can be deactivated
-	defaultUserDeactivationInactivityNotificationPeriod = 24 // 24 days
+	//defaultUserDeactivationInactivityNotificationPeriod = 24 // 24 days
+	defaultUserDeactivationInactivityNotificationPeriod = 1 / 48 // 30 minutes
 	// defaultUserDeactivationInactivityPeriod the default number of days of inactivity before a user account can be deactivated
-	defaultUserDeactivationInactivityPeriod = defaultUserDeactivationInactivityNotificationPeriod + 7 // 31 days (7 days after defaultUserDeactivationInactivityNotificationPeriod)
+	//defaultUserDeactivationInactivityPeriod = defaultUserDeactivationInactivityNotificationPeriod + 7 // 31 days (7 days after defaultUserDeactivationInactivityNotificationPeriod)
+	defaultUserDeactivationInactivityPeriod = defaultUserDeactivationInactivityNotificationPeriod + 1/48 // 1 hour
```
nurali-techie: please check GetXXX() method for these variables.
defaultUserDeactivationInactivityNotificationPeriod [link](https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration.go#L1101)
defaultUserDeactivationInactivityNotificationPeriod [link](https://github.com/fabric8-services/fabric8-auth/blob/master/configuration/configuration.go#L1106)

The GetXXX() method multiply value of defaults with `default * 24 * time.Hour`  so I guess above default will not serve the purpose.

nurali-techie:
@sbryzak one more thing, I think that this test will consume lot of time and if something goes wrong it will be difficult to debug in prod-preview. I suggest to add few logger statement so in case things don't work you know well in advance rather waiting for worker timer.

sbryzak:
@nurali-techie You were right about the GetXXX() methods, they used `getInt()` to retrieve the configuration parameter values, meaning the fractional values wouldn't work.  I changed them to `getFloat64()` for the time being, which should now all the config values to work.

.make/test.mk
```diff
@@ -167,7 +167,7 @@ test-integration-with-coverage: prebuild-check clean-coverage-integration migrat
 test-integration: prebuild-check migrate-database $(SOURCES)
 	$(call log-info,"Running test: $@")
 	$(eval TEST_PACKAGES:=$(shell go list ./... | grep -v $(ALL_PKGS_EXCLUDE_PATTERN)))
-	AUTH_DEVELOPER_MODE_ENABLED=1 AUTH_RESOURCE_DATABASE=1 AUTH_RESOURCE_UNIT_TEST=0 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test $(GO_TEST_INTEGRATION_FLAG) -vet off $(GO_TEST_VERBOSITY_FLAG) $(TEST_PACKAGES)
+	AUTH_DEVELOPER_MODE_ENABLED=1 AUTH_USER_DEACTIVATION_TESTING_MODE=FALSE AUTH_RESOURCE_DATABASE=1 AUTH_RESOURCE_UNIT_TEST=0 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test $(GO_TEST_INTEGRATION_FLAG) -vet off $(GO_TEST_VERBOSITY_FLAG) $(TEST_PACKAGES)
```
nurali-techie: minor: this is more about feature_flag rather than testing_flag so `AUTH_USER_DEACTIVATION_FEATURE_ENABLE` or something similar would be better.

sbryzak: It's only temporary, i'll be removing this parameter in the next PR.

nurali-techie:
@sbryzak yes, but apart from getInt() to getFloat64(), please observe the value is getting multiplied with `24 * time.Hour`.  Approving the PR assuming you have run the code and check the values :+1:

sbryzak:
@nurali-techie yes, there turns out to be some constraints when using `time.Duration()` so I've just overridden the configuration functions altogether instead of trying to set the defaults.  Looks like this works now (tested locally).

PR #820, opened 2019-04-17
user deactivation e2e test

nurali-techie:
1. Basic test - notification, still inactive, deactivate
2. Test with Activity - notification, new activity, deactivation doesn't happen, no activity again, another notification, no activity this time, deactivation

ToDo:
- [x] reactivation in basic test


test/integration/user_deactivation_e2e_base.go
```diff
@@ -0,0 +1,221 @@
...
+		identityID := extractIDFromNotificationPayload(content)
+		NotificationDone <- identityID
+
+		rw.WriteHeader(http.StatusAccepted)
+	}
+}
+
+func ServeClusterRequests(rw http.ResponseWriter, r *http.Request) {
+	log.Printf("Cluster service, Got Request, path=%s\n", r.URL.Path)
+	rw.Write([]byte(`{
```
golangcibot: Error return value of `rw.Write` is not checked (from `errcheck`)

test/integration/user_deactivation_e2e_base.go
```diff
@@ -0,0 +1,221 @@
...
+	Attributes *Attributes
+}
+
+type Attributes struct {
+	ID string
+}
+
+func extractIDFromNotificationPayload(content []byte) string {
+	var payload Payload
+	json.Unmarshal([]byte(content), &payload)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,118 @@
...
+func (s *UserDeactivationSuite) Test1() {
+	t := s.T()
+	os.Setenv("AUTH_CONFIG_FILE_PATH", "e2e_test_config.yml")
+
+	// start auth_service
+	cmd, display := s.authCmd()
+	defer display(t)
+	err := cmd.Start()
+	assert.NoError(t, err)
+	defer cmd.Process.Kill()
```
golangcibot: Error return value of `cmd.Process.Kill` is not checked (from `errcheck`)

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,201 @@
...
+	wantNotificatinCalls := 1
+	wantDeactivateCalls := 1
+
+	// start auth_service
+	cmd, display := s.authCmd()
+	defer display(t)
+	err := cmd.Start()
+	assert.NoError(t, err)
+	defer func() {
+		cmd.Process.Kill()
```
golangcibot: Error return value of `cmd.Process.Kill` is not checked (from `errcheck`)

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,201 @@
...
+	wantNotificatinCalls := 2
+	wantDeactivateCalls := 1
+
+	// start auth_service
+	cmd, display := s.authCmd()
+	defer display(t)
+	err := cmd.Start()
+	assert.NoError(t, err)
+	defer func() {
+		cmd.Process.Kill()
```
golangcibot: Error return value of `cmd.Process.Kill` is not checked (from `errcheck`)

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (s *UserDeactivationSuite) SetupTest() {
+	s.NotificationDone = make(chan string)
+	s.DeactivateDone = make(chan string)
+	s.BaseSuite.SetupTest(s.NotificationDone, s.DeactivateDone)
+}
+
+// notification, still inactive, deactivate
+func (s *UserDeactivationSuite) TestWithNoUserActivity() {
+	os.Setenv("AUTH_CONFIG_FILE_PATH", "e2e_test_config.yml")
```
alexeykazakov: Need to reset it in deffer (or tear down)

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+
+	assert.Equal(t, wantNotificatinCalls, notificatinCalls, "Notification calls not equal, want:%d, got:%d", wantNotificatinCalls, notificatinCalls)
+	assert.Equal(t, wantDeactivateCalls, deactivateCalls, "Deactivation calls not equal, want:%d, got:%d", wantDeactivateCalls, deactivateCalls)
+}
+
+// notification, new activity, deactivation doesn't happen, no activity again, another notification, no activity this time, deactivation
+func (s *UserDeactivationSuite) TestWithUserActivity() {
+	os.Setenv("AUTH_CONFIG_FILE_PATH", "e2e_test_config.yml")
```
alexeykazakov: same here

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+	identity, err = s.Application.Identities().Load(ctx, id)
+	require.NoError(t, err)
+	require.NotNil(t, identity)
+	require.NotNil(t, identity.DeactivationNotification,
+		"DeactivationNotification is not set after %d seconds from notification call, identity:%v", attempt, identity)
+}
+
+func (s *UserDeactivationSuite) VarifyDeactivate(userID string) {
```
alexeykazakov:
```suggestion
func (s *UserDeactivationSuite) VerifyDeactivate(userID string) {
```

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+	t := s.T()
+	require.NotNil(t, userID, "User deactivation can't be varified as UserID is nil")
```
alexeykazakov: userID can't be nil. It's `string`. Not `*string`.

test/integration/user_deactivation_e2e_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	id, err := uuid.FromString(userID)
+	require.NoError(t, err)
+	user, err := s.Application.Users().Load(context.Background(), id)
+	assert.Error(t, err) // not found as users.delete_at is set
+	assert.Nil(t, user)
+	log.Printf("Deactivate call varified successfully")
```
alexeykazakov:
```suggestion
	log.Printf("Deactivate call verified successfully")
```

nurali-techie:
> `test/*` packages are ignored, so, the test is not run when `make test-integration` is called.

@alexeykazakov yes, we need to find a way to run this during build.  Any suggestion ?? Cc: @xcoulon

alexeykazakov:
> yes, we need to find a way to run this during build. Any suggestion ??

Move the tests to `authentication/account/worker` package.

.make/test.mk
```diff
@@ -191,6 +191,14 @@ test-remote: prebuild-check $(SOURCES)
 test-migration: prebuild-check
 	AUTH_RESOURCE_DATABASE=1 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test -vet off $(GO_TEST_VERBOSITY_FLAG) github.com/fabric8-services/fabric8-auth/migration
 
+.PHONY: test-e2e
+## Runs the e2e tests
+## Make sure you ran "make integration-test-env-prepare" before you run this target.
+test-e2e: prebuild-check migrate-database $(SOURCES)
+	$(call log-info,"Running test: $@")
+	$(eval TEST_PACKAGES:=$(shell go list ./... | grep integration))
```
alexeykazakov: So, if you do include `test` package here then IMO it's worth keeping the tests into `test/integration` package.

nurali-techie:
> test/* packages are ignored, so, the test is not run when make test-integration is called.

New `make test-e2e` target added to run this e2e test.  This target is part of cico_run_test also.

.make/test.mk
nurali-techie: only `integration` package is considered.  It's `grep integration`.
@alexeykazakov If you want, I can move files back to `/test/integration` and here only include `/test/integration` package ??

alexeykazakov: Yes. I just don't like packages which contains `only` test related stuff and which don't have `test` in the name :)
Also add this package to `ALL_PKGS_EXCLUDE_PATTERN` so it's not run during integration tests.

alexeykazakov: You actually don't have to include it into `ALL_PKGS_EXCLUDE_PATTERN` if you move it to `/test/integration` because `test` is already excluded in `ALL_PKGS_EXCLUDE_PATTERN`.

.make/test.mk
```diff
@@ -191,6 +191,14 @@ test-remote: prebuild-check $(SOURCES)
 test-migration: prebuild-check
 	AUTH_RESOURCE_DATABASE=1 F8_LOG_LEVEL=$(F8_LOG_LEVEL) go test -vet off $(GO_TEST_VERBOSITY_FLAG) github.com/fabric8-services/fabric8-auth/migration
 
+.PHONY: test-e2e
```
alexeykazakov: Let's also add it to `make test-all`

.make/test.mk
nurali-techie: done [79ba6c1](https://github.com/fabric8-services/fabric8-auth/pull/820/commits/79ba6c12125723c61a371c5bbd0f18bb7ce222a1)

controller/token.go
```diff
@@ -231,6 +231,9 @@ func (c *TokenController) Exchange(ctx *app.ExchangeTokenContext) error {
 func (c *TokenController) exchangeWithGrantTypeRefreshToken(ctx *app.ExchangeTokenContext) (*app.OauthToken, error) {
 	// retrieve the access token from the request header, but ignore if it was not found
 	accessToken := goajwt.ContextJWT(ctx)
+	if accessToken == nil {
+		return nil, errors.NewBadParameterError("access_token", nil).Expected("not nil")
```
alexeykazakov: Hm.. According to the comment above (line 232) Access Token is optional but it's actually not optional since it's used in line 251. I would delete this change from this PR and address it in a different PR to make sure we are not breaking any API.
Also the error message is not correct. It's not a parameter we expect here. It's `Authorization` header. Would need to use `errors.NewBadParameterErrorFromString()` instead.

design/token.go
```diff
@@ -100,6 +100,7 @@ var _ = a.Resource("token", func() {
 	})
 
 	a.Action("Exchange", func() {
+		a.Security("jwt")
```
alexeykazakov: It will break API. Not all exchange call require JWT security / expect access token in Authorization header.

PR #830, opened 2019-04-30
refactor deactivation worker locking

xcoulon:
attempt to obtain a lock at each cycle, but
only release during shutdown

fixes #ODC625

### DIFF
--- a/main.go
+++ b/main.go
@@ -321,7 +321,7 @@ func main() {
 		workers = append(workers, userDeactivationWorker)
 	}
 	// graceful shutdown
-	go handleShutdown(db, workers[:]...)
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/6267dcfe83708b7f88d27a4c8e4064716f5fa493/.ike-prow/test-keeper.yaml).

xcoulon:
> LGTM. Just wondering if we could add some tests for that...

yeah, we already have some tests, but it seems like there was no need to change anything in the test despite the change in the lock obtention strategy...

PR #833, opened 2019-05-02
Fix deactivation worker locking mechanism

alexeykazakov:
This PR fixes a few issues with the locking mechanism:

1. Don't use `pglock.FailIfLocked()` option in `c.Acquire(name, pglock.FailIfLocked())` when trying to acquire a lock. If this option is used then pglock won't check if the lease has been expired. So even if hearbeat failed (because the owner is gone) this option will always return an error which will prevent to acquire the lock by another owner.
2. Don't try to acquire a lock in every cycle. Do it only once when the worker is started. The acquiring routine will wait until the lock is available anyway (see the # 1 above).
3. There is an upstream bug in pglock which was introduced in the latest release - https://github.com/cirello-io/pglock/pull/1 which breaks the hearbeat completely. While the bug is being fixed upstream I've switched to our fork with the fix in place.
4. More logs
5. More tests

### DIFF
--- a/worker/repository/worker_lock_repository.go
+++ b/worker/repository/worker_lock_repository.go
@@ -36,7 +36,7 @@ func (r *lockRepositoryImpl) AcquireLock(ctx context.Context, owner, name string
 		"owner": owner,
 	}, "acquiring lock...")
 	// obtain a lock to prevent other pods to perform this task
-	clnOpts := []pglock.ClientOption{}
+	var clnOpts []pglock.ClientOption
 	if opts != nil {
 		// Useful for testing
 		clnOpts = opts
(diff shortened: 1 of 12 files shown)


worker/repository/worker_lock_repository.go
```diff
@@ -27,34 +28,78 @@ func NewLockRepository(db *sql.DB) LockRepository {
...
 	log.Info(ctx, map[string]interface{}{
 		"lock":  name,
 		"owner": owner,
 	}, "acquiring lock...")
 	// obtain a lock to prevent other pods to perform this task
-	opts := []pglock.ClientOption{
-		pglock.WithCustomTable("worker_lock"),
-		pglock.WithLeaseDuration(3 * time.Second),
-		pglock.WithHeartbeatFrequency(1 * time.Second),
+	clnOpts := []pglock.ClientOption{}
```
golangcibot: ineffectual assignment to `clnOpts` (from `ineffassign`)

PR #834, opened 2019-05-02
Fix delete users in WIT

alexeykazakov:
We should use username, not ID when calling WIT.
This error is not critical. It's just logged and ignored but still worth fixing.

Also, I've changed the order of deactivation. Delete from WIT, Che, Tenant first. If Che fails then return an error. If Tenant fails then return an error. If both Che and Tenant succeed then proceed with removing the user from Auth DB.

### DIFF
--- a/authentication/account/service/user.go
+++ b/authentication/account/service/user.go
@@ -221,19 +221,40 @@ func (s *userServiceImpl) notifyIdentityBeforeDeactivation(ctx context.Context, 
 
 // DeactivateUser deactivates a user, i.e., mark her as `active=false`, obfuscate the personal info and soft-delete the account
 func (s *userServiceImpl) DeactivateUser(ctx context.Context, username string) (*repository.Identity, error) {
-	var identity *repository.Identity
+	identities, err := s.Repositories().Identities().Query(
+		repository.IdentityWithUser(),
+		repository.IdentityFilterByUsername(username),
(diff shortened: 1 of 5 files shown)


authentication/account/service/user.go
```diff
@@ -221,19 +221,40 @@ func (s *userServiceImpl) notifyIdentityBeforeDeactivation(ctx context.Context,
...
+		log.Error(ctx, map[string]interface{}{"identity_id": identity.ID, "error": err}, "error occurred during user deactivation on WIT Service")
+	}
+	// call Che
+	// call WIT and Tenant to deactivate the user there as well,
+	// using `auth` SA token here, not the request context's token
+	err = s.Services().CheService().DeleteUser(ctx, *identity)
+	if err != nil {
+		// do not proceed with tenant removal if something wrong happened during Che cleanup
+		return nil, errs.Wrapf(err, "error occurred during deactivation of user '%s' on Che Service", identity.ID)
+	}
```
nurali-techie: Do we get HTTP 404 and shall we ignore it ??
T1 - /deactivate called
T2 - called che_delete
T3 - tenant_failed
T4 - /deactivate called second time
T5 - che_delete called and returns HTTP 404 ??

alexeykazakov: Yeah.. I was thinking about that usecase too. We should probably ignore 404 from che (or even tenant) in case of re-deactivation.

PR #836, opened 2019-05-02
Don't fail deactivation if user not found

alexeykazakov:
Follow up on https://github.com/fabric8-services/fabric8-auth/pull/834#discussion_r280301696

User could be deactivated/deleted partially. For example in Che we succeed but failed in Tenant. In this case reg app will reschedule deactivation. So, we want to ignore not found errors in the flow (just log them) otherwise re-deactivation won't ever succeed.

### DIFF
--- a/authentication/account/service/tenant_whitebox_test.go
+++ b/authentication/account/service/tenant_whitebox_test.go
@@ -108,8 +108,13 @@ func (s *TestTenantServiceSuite) TestDelete() {
 	err := s.ts.Delete(ctx, identityID)
 	require.NoError(s.T(), err)
 
-	// Fail if returned anything but No Contented
+	// OK if returned Not Found
 	s.doer.Client.Response = &http.Response{Body: body, StatusCode: http.StatusNotFound}
+	err = s.ts.Delete(ctx, identityID)
+	require.NoError(s.T(), err)
+
+	// Fail if returned another error
+	s.doer.Client.Response = &http.Response{Body: body, StatusCode: http.StatusInternalServerError}
 	err = s.ts.Delete(ctx, identityID)
 	require.Error(s.T(), err)
 	assert.Equal(s.T(), "unable to delete tenant", err.Error())
--- a/che/service/che_service.go
+++ b/che/service/che_service.go
@@ -5,15 +5,16 @@ import (
 	"fmt"
 	"net/http"
 
-	token2 "github.com/fabric8-services/fabric8-auth/authorization/token"
-
 	"github.com/fabric8-services/fabric8-auth/application/service"
 	"github.com/fabric8-services/fabric8-auth/application/service/base"
 	servicecontext "github.com/fabric8-services/fabric8-auth/application/service/context"
 	"github.com/fabric8-services/fabric8-auth/authentication/account/repository"
+	token2 "github.com/fabric8-services/fabric8-auth/authorization/token"
 	"github.com/fabric8-services/fabric8-auth/authorization/token/manager"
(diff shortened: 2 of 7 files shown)


authentication/account/service/tenant_whitebox_test.go
```diff
@@ -108,9 +108,14 @@ func (s *TestTenantServiceSuite) TestDelete() {
 	err := s.ts.Delete(ctx, identityID)
 	require.NoError(s.T(), err)
 
-	// Fail if returned anything but No Contented
+	// OK if returned Not Found
```
xcoulon: shouldn't you add a test with a non-`204`/non-`404` (eg: `500`) status code to check that an error is returned?

che/service/che_service.go
```diff
@@ -76,7 +77,12 @@ func (s *cheServiceImpl) DeleteUser(ctx context.Context, identity repository.Ide
 			"response_status": res.Status,
 			"response_body":   bodyString,
 		}, "unable to delete user in Che")
-		return errs.Wrapf(err, "unable to delete user '%s' in Che", identity.ID.String())
+		if res.StatusCode == http.StatusNotFound {
```
xcoulon: should we log with a `WARN` level if we get a `404`, and log with an `ERROR` level otherwise?

xcoulon: also, you could probably use a `switch(res.StatusCode) {...}` block here to treat the different kind of responses (204/404/others)

authentication/account/service/tenant_whitebox_test.go
alexeykazakov: Already in place - https://github.com/fabric8-services/fabric8-auth/pull/836/files#diff-b2589ca644bffe667c463b4d566aee6aR116

xcoulon: ah ok, thanks! (frankly, subtests make it easier to distinguish the use-cases :P )

che/service/che_service.go
alexeykazakov: Done in https://github.com/fabric8-services/fabric8-auth/pull/836/commits/611f09ae78bbdf18ce1d58dce357e9f203e1eaee

alexeykazakov:
[test]

controller/namedusers.go
```diff
@@ -94,11 +94,17 @@ func (c *NamedusersController) Deactivate(ctx *app.DeactivateNamedusersContext)
 			"err":      err,
 			"username": ctx.Username,
 		}, "error occurred while deactivating user")
-		return jsonapi.JSONErrorResponse(ctx, err)
+		if notFound, _ := errors.IsNotFoundError(err); !notFound {
+			return jsonapi.JSONErrorResponse(ctx, err)
+		}
+		// Ignore notFound error. If user has not found then we don't want the reg-app keep trying to deactivate it in auth.
+		// We can add some dummy Identity in response here
+		identity = &repository.Identity{}
```
sbryzak: Don't you also need to set the `User` property, since that is what is returned in line 109?

alexeykazakov: User will be empty (not nil). Reg app doesn't use the returned identity/user so we can trick our API by returning an empty identity/user.

alexeykazakov:
Thanks for the review!

nurali-techie:
@alexeykazakov I thought it is service_consumer [code_link](https://github.com/fabric8-services/fabric8-auth/blob/master/authentication/account/service/user.go#L246) which decided to ignore `not found` rather service_proivder [code_link](https://github.com/fabric8-services/fabric8-auth/blob/master/che/service/che_service.go#L79).

PR #839, opened 2019-05-03
Fix usage of user inactivity configuration functions

sbryzak:
Renamed these configuration functions:

`GetUserDeactivationInactivityPeriodDays` to `GetUserDeactivationInactivityPeriod`
`GetUserDeactivationInactivityNotificationPeriodDays` to `GetUserDeactivationInactivityNotificationPeriod`

Updated user service blackbox test to not multiple the duration by 24 hours *again*


nurali-techie:
@xcoulon @alexeykazakov I am not sure if dropping `Days` from method name is right thing here, WDYT ??

xcoulon:
> @xcoulon @alexeykazakov I am not sure if dropping `Days` from method name is right thing here, WDYT ??

We can rename those funcs, but let's keep the env var names as they are today so we know we configure in `days`. But being able to return a duration in seconds during the tests make sense, indeed.
As you suggested @nurali-techie, this would bring inconsistency with other funcs which are prefixed with `Seconds` or `Millis`, so we should actually remove all notion of units and return a `time.Duration`, which implies the unit.

alexeykazakov:
@xcoulon thanks for the review!
Yes, now everything should look clean. All methods that return time.duration doesn't time units in the method names to avoid confusion. But all the configuration variables specify the unit because they do define time in specific units.

xcoulon:
> @xcoulon thanks for the review!
> Yes, now everything should look clean. All methods that return time.duration doesn't time units in the method names to avoid confusion. But all the configuration variables specify the unit because they do define time in specific units.

thanks for taking care of those extra changes, @alexeykazakov ! and yes, having such suffixes did not make sense anymore now that the functions return a `time.Duration`

PR #840, opened 2019-05-03
improve stability of worker tests

xcoulon:
make sure that all workers are stopped before
checking the database.
Worker stop is async, so we need to check each worker
state to make sure it was actually stopped.

Fixes #ODC727

### DIFF
--- a/authentication/account/worker/user_deactivation_worker.go
+++ b/authentication/account/worker/user_deactivation_worker.go
@@ -9,13 +9,6 @@ import (
 	"github.com/fabric8-services/fabric8-auth/worker"
 )
 
-// UserDeactivationWorker the interface for the User Deactivation Worker,
-// which takes care of deactivating accounts of inactive users who were previously notified but did not come back afterwards.
-type UserDeactivationWorker interface {
-	Start(freq time.Duration)
-	Stop()
-}
-
 const (
 	// UserDeactivation the name of the worker that deactivates users.
 	// Also, the name of the lock used by this worker.
@@ -23,9 +16,9 @@ const (
 )
 
 // NewUserDeactivationWorker returns a new UserDeactivationWorker
-func NewUserDeactivationWorker(ctx context.Context, app application.Application) UserDeactivationWorker {
+func NewUserDeactivationWorker(ctx context.Context, app application.Application) worker.Worker {
 	w := &userDeactivationWorker{
-		worker.Worker{
+		worker.BaseWorker{
 			Ctx:   ctx,
 			App:   app,
 			Owner: worker.GetLockOwner(ctx),
@@ -37,10 +30,10 @@ func NewUserDeactivationWorker(ctx context.Context, app application.Application)
 }
 
 type userDeactivationWorker struct {
-	worker.Worker
+	worker.BaseWorker
 }
 
-func (w *userDeactivationWorker) deactivateUsers() {
+func (w userDeactivationWorker) deactivateUsers() {
 	log.Info(w.Ctx, map[string]interface{}{
 		"owner": w.Owner,
 	}, "starting cycle of inactive users deactivation")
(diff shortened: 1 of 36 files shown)


xcoulon:
> LGTM. But don't we have the same issue with other worker tests? https://github.com/fabric8-services/fabric8-auth/tree/master/authentication/account/worker

Good point, indeed. I'll check the other tests

authentication/account/worker/user_deactivation_worker.go
```diff
@@ -37,10 +30,10 @@ func NewUserDeactivationWorker(ctx context.Context, app application.Application)
 }
 
 type userDeactivationWorker struct {
-	worker.Worker
+	worker.BaseWorker
 }
 
-func (w *userDeactivationWorker) deactivateUsers() {
+func (w userDeactivationWorker) deactivateUsers() {
```
alexeykazakov: Any particular reason not to use a pointer?

xcoulon:
@alexeykazakov I've converted the worker tests to "benchmark tests" so we can run multiple iterations and make sure that it's always OK.
The Jenkins logs show the following output:
```
...
pkg: github.com/fabric8-services/fabric8-auth/worker
BenchmarkWorker/BenchmarkMultipleWorkers           	stopping 3 workers...
stopping 3 workers...
stopping 3 workers...
stopping 3 workers...
stopping 3 workers...
      20	  60275154 ns/op	   22428 B/op	     396 allocs/op
BenchmarkWorker/BenchmarkMultipleWorkers-2         	stopping 3 workers...
stopping 3 workers...
       2	 553033246 ns/op	  252272 B/op	    4369 allocs/op
BenchmarkWorker/BenchmarkMultipleWorkers-4         	stopping 3 workers...
stopping 3 workers...
       2	 728599718 ns/op	  365292 B/op	    6129 allocs/op
PASS
ok  	github.com/fabric8-services/fabric8-auth/worker	9.426s
...
```

authentication/account/worker/user_deactivation_worker.go
xcoulon: it's not needed

worker/worker_blackbox_test.go
```diff
@@ -18,23 +22,41 @@ import (
 
 type WorkerTestSuite struct {
 	gormtestsupport.DBTestSuite
+	application application.Application
+	ctx         context.Context
```
alexeykazakov: We don't need this application and ctx anymore since you reverted the benchmark thing.

worker/worker_blackbox_test.go
```diff
@@ -18,23 +22,41 @@ import (
...
 }
 
+// SetupSuite overrides the DBTestSuite's function but calls it before doing anything else
+// The SetupSuite method will run before the tests in the suite are run.
+// It sets up a database connection for all the tests in this suite without polluting global space.
+func (s *WorkerTestSuite) SetupSuite() {
+	s.DBTestSuite.SetupSuite()
+	s.ctx = migration.NewMigrationContext(context.Background())
+	s.application = gormapplication.NewGormDB(s.DB, s.Configuration, nil)
+}
```
alexeykazakov: This override is not needed either.

worker/worker_blackbox_test.go
xcoulon: fixed in c3c911f

worker/worker_blackbox_test.go
xcoulon: fixed in c3c911f

alexeykazakov: I don't need to define that method at all since it just calls s.DBTestSuite.SetupSuite()

xcoulon: fixed in 1ee9935

PR #842, opened 2019-05-06
Sync call to notification service when notifying inactive users

alexeykazakov:
We should call notification service synchronously from deactivation notification worker. Otherwise we will mark the user as successfully notified even if we got an error from the notification service.


nurali-techie:
@alexeykazakov just for future ref: notification_service `/api/notify` returns `HTTP 202 Accepted` on successful notification request and later send notification email asynchronously.  So notification will be always async only.  The consumer of notification_service (here auth_service) should send notification synchronously only which this PR does.
Check here for notification seq_flow: https://github.com/fabric8-services/fabric8-notification#notification-api

PR #847, opened 2019-05-09
Remove _total from counter metric names to avoid conversion to Hz

alexeykazakov:
Counter type metrics (as inferred from names like _total) are automatically rate-converted to Hz by pcp.

If we rename the counter metrics away from *_total then the row counter should be available in pcp.

### DIFF
--- a/metric/metric.go
+++ b/metric/metric.go
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	UserDeactivationNotificationCounterName string = "user_deactivation_notification_total"
-	UserDeactivationCounterName             string = "user_deactivation_total"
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/f2943297e2930e2e78648c2e51e4150302707460/.ike-prow/test-keeper.yaml).

metric/metric.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	UserDeactivationNotificationCounterName string = "user_deactivation_notification_total"
-	UserDeactivationCounterName             string = "user_deactivation_total"
+	UserDeactivationNotificationCounterName string = "user_deactivation_notification" // PCP automatically rate-converts *_total counters to Hz. So, we don't use _total in the name to avoid that conversion
```
sbose78: OK 👍

PR #848, opened 2019-05-13
log when deactivation counter metric is incremented

xcoulon:
Log when the deactivation counter is incremented, just like we do for the deactivation notification counter

### DIFF
--- a/metric/metric.go
+++ b/metric/metric.go
@@ -10,8 +10,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/fdcd749c8ed75ce8a3a4fca474b51a66724290ee/.ike-prow/test-keeper.yaml).

PR #853, opened 2019-05-21
Clean up che when banning users

alexeykazakov:
When we ban users we should also clean up che workspaces.

TODO:
- [x] Add tests

### DIFF
--- a/authentication/account/service/user.go
+++ b/authentication/account/service/user.go
@@ -114,6 +114,10 @@ func (s *userServiceImpl) BanUser(ctx context.Context, username string) (*reposi
 		return s.Repositories().Users().Save(ctx, &identity.User)
 	})
 
+	if err == nil {
+		err = s.deleteUserFromOtherServices(ctx, identity)
+	}
+
 	return identity, err
 }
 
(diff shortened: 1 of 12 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Your plugin configuration is stored in the [file](https://github.com/fabric8-services/fabric8-auth/blob/835ca58a6248552b460d9c09feae361ac5abc468/.ike-prow/test-keeper.yaml).

authentication/account/service/user.go
```diff
@@ -114,6 +114,10 @@ func (s *userServiceImpl) BanUser(ctx context.Context, username string) (*reposi
 		return s.Repositories().Users().Save(ctx, &identity.User)
 	})
 
+	if err == nil {
+		err = s.deleteUserFromOtherServices(ctx, identity)
+	}
+
```
nurali-techie: shall we not call this before DB changes ??  I guess, DB changes is better to be last thing to do.

alexeykazakov: I think we should mark user as banned in Auth DB even if Che or Tenant calls failed.

xcoulon: I get your point, @alexeykazakov , but what happens if the user is marked as banned but Che or Tenant deletion failed? We wouldn't be able to process the user again, would we?

controller/namedusers_blackbox_test.go
```diff
@@ -154,49 +171,55 @@ func (s *NamedUsersControllerTestSuite) TestBan() {
...
 				_, ctrl := s.SecuredServiceAccountController(testsupport.TestOnlineRegistrationAppIdentity)
-				test.BanNamedusersForbidden(s.T(), nil, nil, ctrl, userToBan.Identity().Username)
+				test.BanNamedusersForbidden(t, nil, nil, ctrl, userToBan.Identity().Username)
 			})
 
 			t.Run("regular user", func(t *testing.T) {
 				// Regular user can't deprovision either
 				svc, ctrl := s.SecuredController(*s.Graph.CreateUser().Identity())
-				test.BanNamedusersForbidden(s.T(), svc.Context, svc, ctrl, userToBan.Identity().Username)
+				test.BanNamedusersForbidden(t, svc.Context, svc, ctrl, userToBan.Identity().Username)
```
xcoulon: you don't need to make this change anymore, since a recent release of `testify`, the `s.T()` func works in subtests as well.

authentication/account/service/user.go
alexeykazakov: Yes we should be able to re-ban the user if it failed in Che or Tenant the first time. I added a test for that.
Unlike deactivation we don't change the username or email of the user we are banning. We just set the ban flag. We can do it multiple times.

controller/namedusers_blackbox_test.go
alexeykazakov: Ah, ok. Didn't know that.

PR #857, opened 2019-05-29
user real username when creating audit log

xcoulon:
fix to avoid submitting the request with the
obfuscated username :/


xcoulon:
thanks @MatousJobanek 🙌

PR #859, opened 2019-06-05
add whitelist to exclude user from deactivation

xcoulon:
Use a configuration-based list of usernames to exclude from the user deacvation. 
It's enough if those users are not notified: they won't have a scheduled deactivation either.

Fixes #ODC1007

### DIFF
--- a/authentication/account/repository/identity.go
+++ b/authentication/account/repository/identity.go
@@ -134,8 +134,8 @@ type IdentityRepository interface {
 	DeleteForResource(ctx context.Context, resourceID string) error
 	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
 	List(ctx context.Context) ([]Identity, error)
-	ListIdentitiesToNotifyForDeactivation(ctx context.Context, lastActivity time.Time, limit int) ([]Identity, error)
-	ListIdentitiesToDeactivate(ctx context.Context, lastActivity, notification time.Time, exclusions []string, limit int) ([]Identity, error)
+	ListIdentitiesToNotifyForDeactivation(ctx context.Context, lastActivity time.Time, whitelist []string, limit int) ([]Identity, error)
+	ListIdentitiesToDeactivate(ctx context.Context, lastActivity, notification time.Time, whitelist []string, limit int) ([]Identity, error)
 	IsValid(context.Context, uuid.UUID) bool
 	Search(ctx context.Context, q string, start int, limit int) ([]Identity, int, error)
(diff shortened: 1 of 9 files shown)


authentication/account/repository/identity.go
```diff
@@ -135,7 +135,7 @@ type IdentityRepository interface {
 	Query(funcs ...func(*gorm.DB) *gorm.DB) ([]Identity, error)
 	List(ctx context.Context) ([]Identity, error)
 	ListIdentitiesToNotifyForDeactivation(ctx context.Context, lastActivity time.Time, limit int) ([]Identity, error)
-	ListIdentitiesToDeactivate(ctx context.Context, lastActivity, notification time.Time, limit int) ([]Identity, error)
+	ListIdentitiesToDeactivate(ctx context.Context, lastActivity, notification time.Time, exclusions []string, limit int) ([]Identity, error)
```
alexeykazakov: Let's use either `whitelist` or `exclusions` everywhere :) Just pick one.
BTW, `exclusions` is pretty clear in this context. But it's your call. Both are fine.

openshift/auth.app.yaml
```diff
@@ -218,6 +218,11 @@ objects:
               configMapKeyRef:
                 name: auth
                 key: user.deactivation.interval.seconds
+          - name: AUTH_USER_DEACTIVATION_WHITELIST
```
alexeykazakov: Also add the CM to https://github.com/fabric8-services/fabric8-auth/blob/master/openshift/auth.config.yaml (it's just an example of CM/Secret values)

authentication/account/repository/identity.go
xcoulon: you're right, that's a left-over that I will revert to `whitelist` to be consistent with the rest of the changes

xcoulon:
> Looks good. But shouldn't we also use this list in user deactivation notification thing too? It can be confusing if we keep sending notifications but never deactivate such users.

oh, right, I changed the wrong method! 😤🤦‍♂️

authentication/account/repository/identity.go
xcoulon: done in https://github.com/fabric8-services/fabric8-auth/pull/859/commits/c7ab003a977dc205ea5c62b7a26b319ede8840dd

openshift/auth.app.yaml
xcoulon: good point. Let me do that, too

xcoulon: done in https://github.com/fabric8-services/fabric8-auth/commit/8ca4377d791a004511b8d0a490c0aeea428db5d5

PR #860, opened 2019-06-06
Fix user service controller constructor

sbryzak:
This PR fixes the user service controller constructor.  

Related issue: https://github.com/redhat-developer/rh-che/issues/1409

### DIFF
--- a/controller/user_service_test.go
+++ b/controller/user_service_test.go
@@ -2,11 +2,19 @@ package controller
 
 import (
 	"github.com/fabric8-services/fabric8-auth/app"
+	"github.com/fabric8-services/fabric8-auth/app/test"
+	"github.com/fabric8-services/fabric8-auth/authentication/account/repository"
 	"github.com/fabric8-services/fabric8-auth/authentication/account/tenant"
+	"github.com/fabric8-services/fabric8-auth/gormtestsupport"
+	"github.com/fabric8-services/fabric8-auth/resource"
+	testsupport "github.com/fabric8-services/fabric8-auth/test"
+	testtoken "github.com/fabric8-services/fabric8-auth/test/token"
 	"github.com/fabric8-services/fabric8-common/convert/ptr"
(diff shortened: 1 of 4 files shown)


controller/user_service_test.go
```diff
@@ -16,6 +24,35 @@ var (
...
+	controller := NewUserinfoController(svc, s.Application, testtoken.TokenManager)
+	return svc, controller
+}
+
+func (s *UserServiceControllerTestSuite) TestShowOK() {
+	identity, err := testsupport.CreateTestIdentityAndUserWithDefaultProviderType(s.DB, "userTestShowUserOK"+uuid.NewV4().String())
+	require.Nil(s.T(), err)
+
+	svc, ctrl := s.SecuredController(identity)
+	test.ShowUserServiceInternalServerError(s.T(), svc.Context, svc, ctrl)
```
alexeykazakov: It doesn't seem to test ShowOK :)

sbryzak: Oops.. I've fixed the function name now.

alexeykazakov: OK. But what about ShowOK test? We would have caught the bug if we had that but test in the first place, wouldn't we?

sbryzak: We can't easily get an "OK" result from the controller.. we have extensive tests for the TenantService in `tenant_whitebox_test.go`, however it is testing an internally-created tenant service for which the `rest.HttpDoer` is replaced with a dummy object.  We don't have the mechanism to easily do this from the controller.

alexeykazakov: You can use `factory.WithTenantService(serviceMock)`
See example https://github.com/fabric8-services/fabric8-auth/blob/master/controller/users_blackbox_test.go#L57

sbryzak: I'm updating the test with this.

sbryzak: All looks good now.

PR #861, opened 2019-06-14
remove usage of request in generate token methods

nurali-techie:
- **remove usage of http request** in method `GenerateUnsignedUserAccessTokenForIdentity()`.  Such method ideally no need to have request in given context.  Out of http request only AuthServiceURL is extracted which is now taken from `config.GetAuthServiceURL()`

- introduce new method `ReplaceDomainPrefixInAbsoluteURLStr()` in rest package, which works upon URL string input and does similar to what ReplaceDomainPrefixInAbsoluteURL() doing.

- config.GetAuthServiceURL() return `http://auth.localhost` instead of `http://localhost` for default case when it called from localhost while running test.  We need to add `auth` sub-domain so that  GenerateUnsignedUserAccessTokenForIdentity() test run smoothly.  Also, auth_service_url will be always having sub-domain (in prod and prod-preview) and this change will make things consistent.

### DIFF
--- a/authorization/token/manager/token_manager_test.go
+++ b/authorization/token/manager/token_manager_test.go
@@ -42,7 +42,7 @@ func (s *TestTokenSuite) TestGenerateUserTokenForIdentity() {
 
 func (s *TestTokenSuite) TestGenerateTransientUserAccessTokenForIdentity() {
 	// given
-	ctx := testtoken.ContextWithRequest(nil)
+	ctx := context.Background()
 	user := repository.User{
 		ID:       uuid.NewV4(),
 		Email:    uuid.NewV4().String(),
@@ -98,7 +98,7 @@ func (s *TestTokenSuite) TestRefreshedUserTokenForIdentity() {
 
 func (s *TestTokenSuite) TestGenerateUserTokenAndRefreshFlowForAPIClient() {
 	// given
-	ctx := testtoken.ContextWithRequest(nil)
+	ctx := context.Background()
 	identityID := uuid.NewV4()
 	username := uuid.NewV4().String()
 	email := uuid.NewV4().String()
@@ -458,7 +458,7 @@ func (s *TestTokenSuite) checkConvertToken(offlineToken bool) {
 }
 
 func (s *TestTokenSuite) generateToken(offlineToken bool) (*oauth2.Token, repository.Identity, context.Context) {
-	ctx := testtoken.ContextWithRequest(nil)
+	ctx := context.Background()
 	user := repository.User{
 		ID:       uuid.NewV4(),
 		Email:    uuid.NewV4().String(),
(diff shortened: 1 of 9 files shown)


configuration/configuration.go
```diff
@@ -499,7 +499,7 @@ func (c *ConfigurationData) GetAuthServiceURL() string {
 	case prodPreviewEnvironment:
 		return "https://auth.prod-preview.openshift.io"
 	default:
-		return "http://localhost"
+		return "http://auth.localhost"
```
nurali-techie: added sub-domain `auth` for consistency and this is need for test to process auth_service_url correctly.

authorization/token/manager/token_manager_test.go
```diff
@@ -68,8 +68,8 @@ func (s *TestTokenSuite) TestGenerateTransientUserAccessTokenForIdentity() {
 	expiresIn := time.Duration(s.Config.GetTransientTokenExpiresIn()) * time.Second
 	s.assertExpiresIn(claims["exp"], expiresIn, 10*time.Second) // 60 seconds days +/- 10 seconds
 	s.assertIntClaim(claims, "nbf", 0)
-	s.assertClaim(claims, "iss", "https://auth.openshift.io")
-	s.assertClaim(claims, "aud", "https://openshift.io")
+	s.assertClaim(claims, "iss", "http://auth.localhost")
```
xcoulon: why did these expected values changed?

nurali-techie: @xcoulon before we used to pass request with context created using `testtoken.ContextWithRequest(nil)` this method always create request with host `https://auth.openshift.io`.  So in test the same auth_url is used.

Now, we are not using context with request and directly calling config.GetAuthServiceURL() which now will return `http://auth.localhost` and thus everywhere auth_url is now `http://auth.localhost` and openshift url is `http://localhost`.  Hope this clarifies.

xcoulon: ok, makes sense. Thanks for the clarification @nurali-techie

nurali-techie:
@xcoulon thanks for review :-)

rest/rest_blackbox_test.go
```diff
@@ -108,6 +108,57 @@ func TestReplaceDomainPrefixInAbsoluteURLOK(t *testing.T) {
...
+func TestReplaceDomainPrefixInAbsoluteURLStr(t *testing.T) {
+	tables := []struct {
+		urlStr    string
+		replaceBy string
+		relative  string
+		want      string
+		wantErr   bool
+	}{
+		{
+			urlStr:    "http://auth.localhost",
```
alexeykazakov: We are not really interested in testing dummy test/default URLs. Could we add tests for "http**s**://auth.openshift.io" and "http**s**://auth.prod-preview-openshift.io" URLs?

PR #866, opened 2020-05-05
add worker to cleanup old OAuth State References

xcoulon:
The worker is configured to run every hour (configurable)
and delete all OAuth state references older than 24hrs (hard-coded).

The worker is based on the same implementation as the other workers
(for user deactivation) and thus runs in a go routine and uses
a lock to prevent concurrent executions (although, in this case,
it's not a big deal)

Also, include a migration step to remove all OAuth state refs
older than 24hrs during startup, using a DB migration script.

Fixes https://issues.redhat.com/browse/CRT-539

### DIFF
--- a/authentication/provider/worker/state_cleanup_worker.go
+++ b/authentication/provider/worker/state_cleanup_worker.go
@@ -1,0 +1,47 @@
+package worker
+
+import (
+	"context"
+
+	"github.com/fabric8-services/fabric8-auth/application"
+	"github.com/fabric8-services/fabric8-auth/log"
+	"github.com/fabric8-services/fabric8-auth/worker"
+)
+
+const (
+	// OAuthStateReferenceCleanup the name of the worker that cleans up old OAuth state references.
+	// Also, the name of the lock used by this worker.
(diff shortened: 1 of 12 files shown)


authentication/provider/worker/state_cleanup_worker.go
```diff
@@ -0,0 +1,47 @@
...
+
+type oauthStateReferenceCleanupWorker struct {
+	worker.BaseWorker
+}
+
+func (w oauthStateReferenceCleanupWorker) cleanup() {
+	log.Info(w.Ctx, map[string]interface{}{
+		"owner": w.Owner,
+	}, "starting cycle of inactive users deactivation")
+	// user service has the config settings to limit the number of users to deactivate
```
alexeykazakov, line 39:
```suggestion
	}, "starting cycle of cleaning up old OAuth state references")
```

xcoulon: ah, good catch! I've addressed this change in https://github.com/fabric8-services/fabric8-auth/pull/867 since this PR was already merged.

xcoulon:
> Very nice!

thanks ;)

